PR #4, opened 2023-02-21
Fixed issue for the .fc files compilation with the size more than 91K…

logvik:
…b. The root of problem was in WASM heap.


ex3ndr:
Hello, unfortunately i can't accept random binary blob. Can you make an instruction to build a wasm binaries instead?

logvik:
https://github.com/ton-blockchain/ton/pull/618/files

PR #35, opened 2023-03-13
Fix address() check error message

manylov:
Fixed disappointing error message, appearing when one trying address(0)


anton-trunov:
Hi @manylov, if you'd like to claim this SBT https://society.ton.org/contribute-to-tact-compiler, please provide your TON wallet address

PR #43, opened 2023-04-04
Bounced handlers for structs

mrbonezy:
## Context 

Adds bounced handlers for structs, adding the following syntax:

```
bounced(msg: bounced<First>) {
...
```

This is achieved by adding a "virtual" type when parsing types, with a new kind named `partial_struct` which contains at most the first [224 bits](https://docs.ton.org/develop/smart-contracts/guidelines/tips#payload-size-in-bounced-messages) of the original struct.

The compiler therefore throws on access to any field outside the 224 bits limit.

## Comments
1. Comment-based bounced handlers will be added in the next commit
2. First 32 bits of the bounced messages are now skipped automatically, which is a breaking change.
3. Tests fail here on addresses and expected gas changes - 
image

If that's expected I'll update the snapshots in a further commit prior to merging.
4. Nested structs of the Message type are currently not supported (meaning they will not appear in the partial struct). We can discuss adding support for this after we agree on the initial approach taken here.


src/generator/writeProgram.ts
```diff
@@ -269,10 +269,12 @@ function writeAll(ctx: CompilerContext, wctx: WriterContext, name: string, abiLi
     for (let t of sortedTypes) {
         if (t.kind === 'contract' || t.kind === 'struct') {
             let allocation = getAllocation(ctx, t.name);
+            let allocationBounced = getAllocation(ctx, toBounced(t.name));
```
shaharyakir, line 272: Finalize how allocation should work - 
Currently the type store relies on the partialFields property,
whereas the allocator pushes a separate bounced type to the array, which doesn't feel right.

src/generator/writers/writeSerialization.ts
```diff
@@ -240,6 +240,31 @@ export function writeParser(name: string, forceInline: boolean, allocation: Stor
     });
 }
 
+export function writeBouncedParser(name: string, forceInline: boolean, allocation: StorageAllocation, origin: TypeOrigin, ctx: WriterContext) {
```
shaharyakir, line 243: Note - is this the only part that uses the allocation?

src/storage/resolveAllocation.ts
```diff
@@ -92,8 +92,53 @@ export function resolveAllocations(ctx: CompilerContext) {
                 refs: root.size.refs + reserveRefs
             }
         };
+        
         ctx = store.set(ctx, s.name, allocation);
     }
+    
+    // TODO NOT SURE ABOUT THIS
```
mrbonezy: See https://github.com/tact-lang/tact/pull/43/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.ohm&file-filters%5B%5D=.snap&file-filters%5B%5D=.tact&file-filters%5B%5D=.ts&show-viewed-files=true#r1165202648

anton-trunov:
Hi @shaharyakir, if you'd like to claim this SBT https://society.ton.org/contribute-to-tact-compiler, please provide your TON wallet address

PR #85, opened 2023-11-16
added support for import from node_modules

### LINKED ISSUE #84: ability to import tact files from npm
```solidity
import "@npm/@hitasp/npm_imported";
import "@npm/dynasty/another_imported";
```
Can help for #77 integration. 

I can send PR for this if approved.

hitaspdotnet:
usage
```solidity
import "@npm/@tondynasty/packages/token/jetton/JettonWallet";
```

Close #84


alan890104:
Should we consider the below scenario?

if my local tact file import `package A` with `version 1` directly, and the local tact file also import another `package B` which depends on `package A` with `version 2`, will it causes compilation error?

hitaspdotnet:
@alan890104 yes, if v2 has breaking changes! similar to any other packages you have to use same versions. We work with binaries with relative paths in local.

anton-trunov:
Could you rebase this PR on top of `main` to trigger CI?

anton-trunov:
There is no consensus on this feature so far, please see this discussion: #147. Closing this for now, but we might come back to the issue of imports from some centralized resources later. Right now your best workaround is to vendor your dependencies.

PR #95, opened 2023-11-28
add .del function to maps

### LINKED ISSUE #101: Key deletion syntax for maps
Let's say we a have a map `m: map<Int, Int>`. To remove a key `k` from it, one needs to do `m.set(k, null)`.
This issue proposes to add some syntactic sugar for this map operation.

The exact syntax is to be defined.

Gusarich:
very simple update, but for some reason I cannot proceed with making it to work.

The strangest thing is that by commenting different `Del` receivers in the `maps.tact` file I'm getting different results during compilation:

 - leaving all `DelAddrMap` receivers: successful compilation
 - leaving all `DelAddrMap`'s and one `DelIntMap1`: 
 ```
💼 Compiling project maps...
   > MapTestContract: tact compiler
   > MapTestContract: func compiler
[ 1][t 0][2023-11-28 17:35:35.101000][Fift.cpp:67]      top: <continuation 0x28e3a50>
level 1: <continuation 0x28e3ad0>
level 2: <text interpreter continuation>

main.fif:6134:  s(:-?
 ```
 - leaving all `DelAddrMap`'s, `DelIntMap1` and `DelIntMap2`:
```
💼 Compiling project maps...
   > MapTestContract: tact compiler
   > MapTestContract: func compiler
File not found
```

I've been unsuccessfully trying to solve the problem for few hours already..

### DIFF
--- a/src/abi/map.ts
+++ b/src/abi/map.ts
@@ -334,7 +334,7 @@ export const MapFunctions: Map<string, AbiFunction> = new Map([
                 if (args.length !== 2) {
                     throwError("del expects one argument", ref); // Ignore self argument
                 }
-                let self = args[0];
+                const self = args[0];
                 if (!self || self.kind !== "map") {
                     throwError("del expects a map as self argument", ref); // Should not happen
                 }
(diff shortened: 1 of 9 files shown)


src/abi/map.ts
```diff
@@ -325,6 +325,73 @@ export const MapFunctions: Map<string, AbiFunction> = new Map([
...
+                }
+                if (args[1].name !== self.key) {
+                    throwError(
+                        `del expects a ${self.key} as first argument`,
+                        ref,
+                    );
+                }
+
+                // Returns nothing
+                return { kind: "void" };
```
anton-trunov: I think we should let the user know if the key was actually deleted or it was not found in the map, in other words it should have the following signature: `del : map<K,V> -> K -> Bool`, so we return `true` if deletion was successful

anton-trunov (changes requested):
Awesome stuff! I think we should change `del`'s type signature (see below) and add some negative type checker tests just  to be on the safe side

src/abi/map.ts
Gusarich: Done

anton-trunov (approved):

PR #96, opened 2023-11-28
add ubuntu & windows & macos CI

### LINKED ISSUE #90: Add CI to the repo
CI should support Windows, Linux, macOS

Gusarich:
I've only made it to build and run tests. Do you think we need anything else to run there @anton-trunov ?

also it fails on windows by default... I cannot check it myself right now (I'm on macos), but looks like something's wrong with paths

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [18.x]
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [18.x]
-        os: [ubuntu-latest, windows-2016]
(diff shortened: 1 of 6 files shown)


anton-trunov:
>  Do you think we need anything else to run there @anton-trunov ?

I think running [tact-template](https://github.com/tact-lang/tact-template)'s tests with the most recent version of Tact on each PR would be beneficial. And when we fix the dependencies for Blueprint and Tact to work together, having Blueprint-based tests would be awesome.

anton-trunov:
> also it fails on windows by default

I'll try to look at it, although I don't have access to Windows as well

anton-trunov:
Looks like I've fixed the Windows CI failures. I'd probably merge this as is if you think it's ready

Gusarich:
> Looks like I've fixed the Windows CI failures. I'd probably merge this as is if you think it's ready

It builds successfully now:
 * https://github.com/Gusarich/tact/actions/runs/7033832726

let’s merge it! I’ll add tact-template testing in a separate PR

anton-trunov:
Thanks for your PR <3

>  I’ll add tact-template testing in a separate PR

That would be awesome! I created a separate issue for this: #98

PR #97, opened 2023-11-28
Add conditional expressions

### LINKED ISSUE #94: Conditional (ternary) expression
It would be nice if we have something like ternary operator in `javascript` or any other languages. It would help us with less code and easier implementation of simple null checks and if statements.

Gusarich:
closes #94 

not sure but maybe we should make the syntax like `VAL if COND else OTHER_VAL` instead of `COND ? VAL : OTHER_VAL` for simplicity


anton-trunov:
Could you rebase this PR on top of `main` to trigger CI?

PR #99, opened 2023-11-29
Add bin and oct integer literals & Allow underscores in them

### LINKED ISSUE #102: Add binary and octal integer literals
For instance, `0b10101111` and `0b1010_1111`. See also #103.

Gusarich:
ohm doesn't have `binDigit` so I needed to define it myself in grammar


anton-trunov:
I think we should also resolve issue #103 in this PR

Gusarich:
@anton-trunov Done! I've also added oct literals (`0o123`)

anton-trunov:
@Gusarich Awesome stuff! I've noticed you have positive tests, which is great, let's also add a few negative tests as well, like `0b_00101010`, `_42`, `0b123`, etc. do not get accepted by the parser.

Gusarich:
> @Gusarich Awesome stuff! I've noticed you have positive tests, which is great, let's also add a few negative tests as well, like `0b_00101010`, `_42`, `0b123`, etc. do not get accepted by the parser.

`_42` is considered by parser as a valid `id`, so I'm not sure how can we test this case.

anton-trunov:
Ah, sorry, it was a typo: `0_42` should not be valid.

Gusarich:
> Ah, sorry, it was a typo: `0_42` should not be valid.

Currently parser allows trailing zeroes.. should I also modify this behaviour?

src/grammar/grammar.ohm
```diff
@@ -173,10 +173,16 @@ Tact {
...
-    integerLiteralHex = "0x" hexDigit+
-                      | "0X" hexDigit+
+    integerLiteral = integerLiteralHex | integerLiteralBin | integerLiteralOct | integerLiteralDec // Order is important
+    integerLiteralDec = digit+ ("_" | digit)*
+    integerLiteralHex = "0x" hexDigit+ ("_" | hexDigit)*
+                      | "0X" hexDigit+ ("_" | hexDigit)*
+    integerLiteralBin = "0b" binDigit+ ("_" | binDigit)*
+                      | "0B" binDigit+ ("_" | binDigit)*
+    integerLiteralOct = "0o" octDigit+ ("_" | octDigit)*
+                      | "0O" octDigit+ ("_" | octDigit)*
```
anton-trunov: I'd model literals after JS/TS, because it's kind of a go-to language for TON, so people could copy some literals from Tact code into TS tests or something like that.

There is one caveat though: JS's octal literals starting with `0`. For the reasons of backwards compatibility it might be a bit too late to not allow integer literals with leading zeros. So this might result in some suboptimal UX when copying things from Tact to TS tests. (We could introduce a warning about this, though).

These are JS's restrictions, regarding underscores in numeric literals:
1. JS does not allow `_` at the end of literals.
2. As mentioned earlier, we should also forbid `_` after leading zeros.
3. JS does not allow multiple `_` in a row (so `4_2` is a valid literal, but `4__2` is not).
 
Of course, some other languages, like [OCaml](https://ocaml.org) are very liberal in this respect, e.g. it's ok with `0004____2_` being a valid integer literal. However, I think going with JS's grammar (except for octals) is something we should do.

src/grammar/grammar.ts
```diff
@@ -547,7 +547,7 @@ semantics.addOperation<ASTNode>('resolve_expression', {
 
     // Literals
     integerLiteral(n) {
-        return createNode({ kind: 'number', value: BigInt(n.sourceString), ref: createRef(this) }); // Parses dec-based integer and hex-based integers
+        return createNode({ kind: 'number', value: BigInt(n.sourceString.replaceAll('_', '')), ref: createRef(this) }); // Parses dec, hex, and bin numbers
```
anton-trunov, line 595: Hmm, looks like we translate all integer literals into FunC's decimal literals (because we strip this info here). Let's not change it now, but might be nice to have direct correspondence (modulo binary and octal literals, which FunC to does not support yet)

anton-trunov:
> Currently parser allows trailing zeroes.. should I also modify this behaviour?

You probably meant leading zeros: I tried to address it here: https://github.com/tact-lang/tact/pull/99#discussion_r1463338468

anton-trunov:
Looks like there are some merge conflicts from your previous PR :)

anton-trunov (approved):

PR #105, opened 2023-11-30
Slices and Strings comparisons

### LINKED ISSUE #24: String comparison with == and !=
We can compare slices, so we should be able to compare Strings

Gusarich:
This PR introduces the ability to compare both slices and strings via `==` operation.

For slice comparison, their hashes (calculated using `HASHSU`) can be compared directly.
As for strings, they are essentially slices and can be compared similarly. However, there's a consideration with strings due to their potential for various serializations (e.g., a string can be split in the middle and continued in a subsequent reference, still constituting a valid string). Upon reviewing the implementation of the `String` type in Tact, it was determined to be deterministic: all slices in a 'snake' representation are completely filled, except for the last one. Therefore, the same slice comparison method is applicable to strings as well.


0kenx:
Hash is not the most gas efficient way to do this. There's a native opcode for this:

`int equal_slices(slice a, slice b) asm "SDEQ";`

Skydev0h:
> Hash is not the most gas efficient way to do this. There's a native opcode for this:
> 
> `int equal_slices(slice a, slice b) asm "SDEQ";`

SDEQ is, essentially, an abbreviation for Slice Data Equal. It does not account for refs at all. Use of hash covers both long (multi-cell) and short (single-cell) strings correctly.

src/generator/writers/writeExpression.ts
```diff
@@ -261,6 +261,54 @@ export function writeExpression(f: ASTExpression, ctx: WriterContext): string {
...
+            }
+            if (!lt.optional && rt.optional) {
+                ctx.used(`__tact_slice_${op}_nullable_one`);
+                return `__tact_slice_${op}_nullable_one(${writeExpression(f.right, ctx)}, ${writeExpression(f.left, ctx)})`;
+            }
+            ctx.used(`__tact_slice_${op}`);
+            return `__tact_slice_${op}(${writeExpression(f.right, ctx)}, ${writeExpression(f.left, ctx)})`;
+        }
+
+        // Case for strings equality
```
anton-trunov: Looks like the case of strings is totally identical to the case of slices. If this is correct, I'd suggest to refactor the code avoiding duplication

anton-trunov (changes requested):
Thanks for contributing more features 👍.
Would be great to
- have some runtime tests with `==` and `!=` both for strings and (some other kinds of) slices;
- add an entry to the changelog (to the `Changed` subsection).

**Update**: done

PR #106, opened 2023-12-01
Add native map traversing functionality

### LINKED ISSUE #79: Support loop for map
A map can have different keys (not 1,2,3)

Gusarich:
This PR introduces the `for` loop - currently just the map traversing version of it, but I'm also thinking about implementing general purpose `for` loops in future.

Syntax (may be changed, it's up for discussions):
```
for (k, v in m) {
    // k: key, v: value, m: map
}
```

What have been already done:
 - grammar
 - ast definitions
 - statement resolving

What have to be done:
 - actual implementation of map traversing loop
 - tests


Gusarich:
I've ran into a little problem when was trying to implement it. Here's an example of how to traverse a map in FunC:
```
(int key, slice val, int flag) = d.udict_get_min?(256);
while (flag) {
    ;; do something with pair key->val
    
    (key, val, flag) = d.udict_get_next?(256, key);
}
```
This loop uses 3 variables: key, value and flag that indicates whether the loop should be terminated or not (actually it's set to `false` when there are no more keys in map).
We can easily define key and value variables by using same names in the FunC code (`id(keyName)` etc), but I'm not sure about the best solution regarding the flag variable. It must not be overlapped with any other existing variable names **including other flag variables in case with nested loops**.

Well, we can just ban nested map traversing loops... but I don't think it's a good idea.

@anton-trunov what do you think?

xsr:
I'm sorry if mine response here is unrelated (delete than) but just trying dive into TACT devel and feel compatible to join discussion to ask:

«just generate random name like `for_flag_RANDOM_HASH` ?»

Gusarich:
> «just generate random name like `for_flag_RANDOM_HASH` ?»

@xsr I've thought about this, but it adds some kind of indeterminacy to compiler which isn't good.

nonam3e:
> I've ran into a little problem when was trying to implement it. Here's an example of how to traverse a map in FunC:
> ```
> (int key, slice val, int flag) = d.udict_get_min?(256);
> while (flag) {
>     ;; do something with pair key->val
>     
>     (key, val, flag) = d.udict_get_next?(256, key);
> }
> ```
> This loop uses 3 variables: key, value and flag that indicates whether the loop should be terminated or not (actually it's set to `false` when there are no more keys in map).
> We can easily define key and value variables by using same names in the FunC code (`id(keyName)` etc), but I'm not sure about the best solution regarding the flag variable. It must not be overlapped with any other existing variable names **including other flag variables in case with nested loops**.
> 
> Well, we can just ban nested map traversing loops... but I don't think it's a good idea.
> 
> @anton-trunov what do you think?

Maybe can it be solved by checking key instead of flag (key.is_null?)?

Gusarich:
> Maybe can it be solved by checking key instead of flag (key.is_null?)?

@nonam3e Well, it’ll work, but will consume more gas (18 for each iteration).

novusnota:
Found an interesting read on designing a `for-each` loop syntax: https://harelang.org/blog/2024-04-01-introducing-for-each-loops-in-hare. This won't help much with [this issue](https://github.com/tact-lang/tact/pull/106#issuecomment-1835245319), but it can be somewhat inspirational nonetheless, despite coming from a systems programming language :)

anton-trunov:
> We can easily define key and value variables by using same names in the FunC code (id(keyName) etc), but I'm not sure about the best solution regarding the flag variable. It must not be overlapped with any other existing variable names including other flag variables in case with nested loops.

@Gusarich The usual approach here is to define a fresh identifier generator. Something like this:

```ts
// fresh.ts

let counter = 0;    // a mutable variable

function freshIdentifier(prefix: string): string {
    const fresh = `__tact_freshId${prefix}${counter}`;
    counter += 1;
    return fresh;
}
```

and then you can use it to generate fresh `flag` variables with `freshIdentifier("flag")`. Using a prefix is not necessary, it should make debugging a bit easier.

src/test/features/map-traverse.tact
anton-trunov, line 1: TODO: empty map traverse

tact.config.json
anton-trunov, line 1: looks like something went wrong during merging/rebasing: most of the contents of this file got deleted

src/test/features/map-traverse.tact
anton-trunov: TODO: negative tests for the type checker and variable scope: e.g.
```tact
foreach (k, k in m) { ... }
```
is a variable scope mistake

or 
```tact
foreach (k, m in m) { ... }
```
is a mistake too

anton-trunov (dismissed):
HUGE!

anton-trunov (approved):

Issue #113, opened 2023-12-21
Syntax Error:  nativeReserve


tactfunc:
```
        nativeReserve(ton("0,1"), 0);
````

Still enable to compile. It's a bug.

Should be 
```
        nativeReserve(ton("0.1"), 0);
```

Reference: https://docs.tact-lang.org/language/ref/advanced#nativereserve

Gusarich:
how is this a bug? you just have to follow the number format.

anton-trunov:
Let's keep this one open for now. There are ways to address this. Just not right now

PR #120, opened 2024-01-15
Check if the input of `ton` global function is a valid number

mhbdev:
This checked will solve the problem faced in  #113

### DIFF
--- a/src/abi/global.ts
+++ b/src/abi/global.ts
@@ -23,16 +23,14 @@ export const GlobalFunctions: { [key: string]: AbiFunction } = {
             if (args[0].name !== 'String') {
                 throwError('ton() expects single string argument', ref);
             }
-            if(!isNaN(parseFloat(args[0].name))) {
-                throwError('ton() expects a valid number', ref);
-            }
             return { kind: 'ref', name: 'Int', optional: false };
         },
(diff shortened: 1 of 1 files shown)


Gusarich:
I don't think it's necessary. If the number isn't valid, it throws an error during compilation which tells what exactly is wrong.

mhbdev:
> I don't think it's necessary. If the number isn't valid, it throws an error during compilation which tells what exactly is wrong.

You are right.
I have made another change hope you guys find it useful.

Gusarich:
@mhbdev can you please explain what this change does?

mhbdev:
> @mhbdev can you please explain what this change does?

I decided to make this change to let the input contain separators like comma and underscore. So there would be no error using those separators to make your numbers more readable.

Gusarich:
> I decided to make this change to let the input contain separators like comma and underscore. So there would be no error using those separators to make your numbers more readable.

I'm not sure if it's correct to do that for in string literals. This needs clarification from @anton-trunov

anton-trunov:
Hey @mhbdev, thanks for the PR, however, I'm inclined to close it because it allows a bit too much freedom, like several commas in a row or leading or trailing commas, which would arguably allow for even less readable code. Hope this won't discourage you from future contribution.

PR #157, opened 2024-02-29
ESLint whole Tact codebase

### LINKED ISSUE #141: Fix issues reported by ESLint

anton-trunov:
Fixes #141

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,7 @@
+node_modules/
+dist/
+converage/
+**/output/
+src/func/funcfiftlib.js
+src/grammar/grammar.ohm*.ts
+src/grammar/grammar.ohm*.js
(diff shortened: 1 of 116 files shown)


.eslintignore
```diff
@@ -0,0 +1,11 @@
+node_modules/
+dist/
+converage/
```
novusnota: `converage/` → `coverage/`

anton-trunov: good catch!
```suggestion
coverage/
```

examples/increment.spec.ts
novusnota, line 1: Counters state is now neglected due to `counters` and `counters2` being unused. Is it ok?

anton-trunov: I think it's ok, I don't see how those vars were used. It's fine to remove that provided we are not after some side effect when calling `contract.getCounters` and `contract.getCounters2`. The test can be improved, though. We should compare the counters against some expected values. Could be implemented in a separate PR

novusnota (dismissed):

PR #166, opened 2024-03-04
Implement `log2` and `log` functions in math.tact

### LINKED ISSUE #12: stdlib: logarithm calculation
stdlib need a gas-efficient logarithm calculation

Gusarich:
closes #12

### DIFF
--- a/src/generator/writers/writeStdlib.ts
+++ b/src/generator/writers/writeStdlib.ts
@@ -1240,7 +1240,7 @@ export function writeStdlib(ctx: WriterContext) {
                 }
                 int result = 0;
                 while (num >= base) {
-                    num = num / base;
+                    num /= base;
                     result += 1;
                 }
                 return result;
--- a/src/generator/writers/writeStdlib.ts
+++ b/src/generator/writers/writeStdlib.ts
@@ -1235,7 +1235,15 @@ export function writeStdlib(ctx: WriterContext) {
         ctx.context('stdlib');
         ctx.body(() => {
             ctx.write(`
-                return __tact_log2(num) / __tact_log2(base);
+                if (num < base) {
+                    return 0;
+                }
+                int result = 0;
+                while (num >= base) {
+                    num = num / base;
+                    result += 1;
+                }
+                return result;
             `);
         });
     });
--- a/src/generator/writers/writeStdlib.ts
+++ b/src/generator/writers/writeStdlib.ts
@@ -1223,4 +1223,20 @@ export function writeStdlib(ctx: WriterContext) {
             `);
         });
     });
+
+    ctx.fun(`__tact_log2`, () => {
+        ctx.signature(`int __tact_log2(int num)`);
+        ctx.context('stdlib');
+        ctx.asm(`asm "UBITSIZE DEC"`);
+    });
+
+    ctx.fun(`__tact_log`, () => {
+        ctx.signature(`int __tact_log(int num, int base)`);
+        ctx.context('stdlib');
+        ctx.body(() => {
+            ctx.write(`
+                return __tact_log2(num) / __tact_log2(base);
+            `);
+        });
+    });
 }
--- a/src/test/feature-math.spec.ts
+++ b/src/test/feature-math.spec.ts
@@ -348,10 +348,51 @@ describe('feature-math', () => {
 
         for (let num = 1n; num <= 10n; num++) {
             for (let base = 2n; base <= 10; base++) {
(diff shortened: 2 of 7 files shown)


src/generator/writers/writeStdlib.ts
```diff
@@ -1223,4 +1223,20 @@ export function writeStdlib(ctx: WriterContext) {
...
+        ctx.context('stdlib');
+        ctx.asm(`asm "UBITSIZE DEC"`);
+    });
+
+    ctx.fun(`__tact_log`, () => {
+        ctx.signature(`int __tact_log(int num, int base)`);
+        ctx.context('stdlib');
+        ctx.body(() => {
+            ctx.write(`
+                return __tact_log2(num) / __tact_log2(base);
```
anton-trunov: This formula does not work for integer logarithms with an arbitrary base. We basically need `floor(log(num, base))`, which, for example, for `log3(2^32 - 1)` gives us 20 as the result, not 32, which this formula provides.

src/test/feature-math.spec.ts
```diff
@@ -335,5 +338,21 @@ describe('feature-math', () => {
...
+            expect(await contract.getLog2(num)).toBe(
+                BigInt(Math.floor(Math.log2(Number(num))))
+            );
+        }
+
+        for (let num = 1n; num <= 10n; num++) {
+            for (let base = 2n; base <= 10; base++) {
+                const a = Math.floor(Math.log2(Number(num)));
+                const b = Math.floor(Math.log2(Number(base)));
+                const c = BigInt(Math.floor(a / b));
```
anton-trunov: we want `c` to be something like `floor(log2(num) / log2(base))`, just like in the comment above

src/test/feature-math.spec.ts
```diff
@@ -335,5 +338,21 @@ describe('feature-math', () => {
         expect(await contract.getCompare28(dictA, dictB)).toBe(true);
         expect(await contract.getCompare28(dictB, dictA)).toBe(true);
         expect(await contract.getCompare28(dictA, dictA)).toBe(false);
+
+        // Test advanced math operations
+        for (let num = 1n; num <= 100n; num++) {
```
anton-trunov, line 343: Please add more tests for corner cases:
- `num` is 0
- `num` is near the maximum integer value, i.e. `[maxint - 99; maxint]`
- `num` is in the middle of the range of the positive integers, i.e. `[maxint/2 - 50, maxint/2 + 50]`

src/test/feature-math.spec.ts
```diff
@@ -335,5 +338,21 @@ describe('feature-math', () => {
...
+            expect(await contract.getLog2(num)).toBe(
+                BigInt(Math.floor(Math.log2(Number(num))))
+            );
+        }
+
+        for (let num = 1n; num <= 10n; num++) {
```
anton-trunov, line 349: Please add more tests for corner cases:
- `num` is 0
- `num` is near the maximum integer value, i.e. `[maxint - 99; maxint]`
- `num` is in the middle of the range of the positive integers, i.e. `[maxint/2 - 50, maxint/2 + 50]`

anton-trunov (changes requested):

src/generator/writers/writeStdlib.ts
```diff
@@ -1235,7 +1235,15 @@ export function writeStdlib(ctx: WriterContext) {
         ctx.context('stdlib');
         ctx.body(() => {
             ctx.write(`
-                return __tact_log2(num) / __tact_log2(base);
+                if (num < base) {
+                    return 0;
+                }
+                int result = 0;
+                while (num >= base) {
+                    num = num / base;
```
anton-trunov:
```suggestion
                    num /= base;
```

or would it be less efficient in FunC?

src/test/feature-math.spec.ts
```diff
@@ -348,10 +348,51 @@ describe('feature-math', () => {
...
+        const maxint = 2n ** 256n - 1n;
+
+        function bigIntLogBase(num: bigint, base: bigint) {
+            let result = 0n;
+            while (num >= base) {
+                num = num / base;
+                result += 1n;
+            }
+            return result;
+        }
```
anton-trunov: This should be basically something like `Math.floor(Math.log2(Number(num)) / Math.log2(Number(base)))` as above. Otherwise the testing code just repeats the code being tested.

anton-trunov (changes requested):

anton-trunov (approved):
Cool solution with `toString(base).length - 1` for tests!

PR #167, opened 2024-03-05
implicit init function if not present

### LINKED ISSUE #8: Implicit init function if one does not exist
```
Error: contract.tact:1:1: Contract HelloWorld does not have init method
```

Contract without an init method does not compile. Might be a good idea to create an implicit empty init method if one is not provided by the user, so code can be boilerplate free.

Gusarich:
closes #8


anton-trunov (commented):
Looks great!

Let's add a test that checks what happens when there are some initialized and *uninitialized* storage vars and the `init` method is missing

anton-trunov (changes requested):
Please add a test mentioned here: https://github.com/tact-lang/tact/pull/167#pullrequestreview-1935103123

src/test/feature-implicit-init.spec.ts
```diff
@@ -45,4 +46,12 @@ describe('feature-send', () => {
         expect(await contract.getGetCounter()).toBe(2n);
         expect(tracker.collect()).toMatchSnapshot();
     });
+
+    it('should not compile with uninitialized storage fields', async () => {
+        const result = await run({
+            configPath: __dirname + '/test-tact.config.json',
+            projectNames: ['implicit-init-2'],
+        });
+        expect(result).toBe(false);
```
anton-trunov, line 73: We should test against the error the user will see, not just to check if it does not compile. The point is to ensure the error message is understandable in this case.

anton-trunov (changes requested):

anton-trunov (approved):
Nicely done!

PR #169, opened 2024-03-05
add `RenounceOwnership` message to `OwnableTransferable` trait

### LINKED ISSUE #32: Add renounceOwnership to OwnableTransferable
It's a good idea to standardize what it means to transfer ownership to nobody

Probably transfer to newAddress(0,0) - but we don't want users to think about these implementation details

Gusarich:
closes #32


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `RenounceOwnership` message was added to `OwnableTransferable`
```
```suggestion
- The `RenounceOwnership` message and its corresponding receiver were added to the `OwnableTransferable` trait allowing to transfer ownership to nobody
```

anton-trunov (changes requested):
Looks like the address corresponding to `newAddress(0,0)` is *temporarily* suspended, which makes it unfit as an address that nobody owns. We need to find ways to deal with this concern.

PR #185, opened 2024-03-22
Refactor: Extract RunArguments type for clarity

Ivan-Mikitenko:


anton-trunov:
Hi @Ivan-Mikitenko, thanks for your contribution! Could you please describe what issue your PR is solving? It's not that we are opposed to having more type information in the codebase, it's actually needed, however each and every change should be motivated: Tact is a long-term project and we want documented decisions to make it easier for future contributors to help with Tact's implementation.

(We don't have a contributing guide yet, but it's on our TODO list: #186)

PR #192, opened 2024-03-26
Implement string escaping and fix bug with error messages

### LINKED ISSUE #25: No possibility to escape characters in strings
```
s2: String = "no escaping \n \t";
```

Does not compile:

```
Syntax error: expected "\""
```

### LINKED ISSUE #53: Contract does not work if we have ` in throw message
```
    receive() {
        require(1 == 2, "condition can`t be...");
    }
```
After compile message wrap to same symbol like this
```
`target can`t be more than 10`
```
and code died

Gusarich:
This PR resolves #25 and fixes #53 

Supported escaping sequences:
- `\\`
- `\"`
- `\n`
- `\r`
- `\t`
- `\v`
- `\b`
- `\f`
- `\u{ABC}`
- `\uABCD`
- `\xAB`


src/grammar/grammar.ohm
```diff
@@ -215,8 +215,17 @@ Tact {
     boolLiteral = ("true" | "false") ~idPart
 
     // String literal
-    stringLiteralCharacter = ~("\"" | "\\" | lineTerminator) any
-    stringLiteral = "\"" stringLiteralCharacter* "\""
+    // stringLiteralCharacter = ~("\"" | "\\" | lineTerminator) any
+    // stringLiteral = "\"" stringLiteralCharacter* "\""
```
```suggestion
```

src/grammar/grammar.ohm
```diff
@@ -215,8 +215,17 @@ Tact {
...
+    stringLiteral = "\"" (nonQuoteOrBackslashChar | escapeSequence)* "\""
```
anton-trunov, line 218: Let's support all the things from the lexical grammar of JS: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#string_literals.

It has more escape sequences including Unicode, and also more ASCII things, like `\v`, etc. And it allows to split long strings across multiple lines.

anton-trunov (changes requested):
Great start! Let's support some more things though

src/grammar/grammar.ohm
novusnota: For an example of this in Ohm, check out [`ohm-grammar.ohm`](https://github.com/novusnota/vscode-ohm/blob/main/samples/ohm-grammar.ohm#L93-107) or [`es5.ohm#L149`](https://github.com/ohmjs/ohm/blob/main/examples/ecmascript/src/es5.ohm#L149)

novusnota (dismissed):

anton-trunov (approved):
Good stuff 👍

anton-trunov (changes requested):
We need to update CHANGELOG.md

PR #197, opened 2024-03-27
Expand `reduceBool` and add `reduceSlice`

### LINKED ISSUE #195: Support more boolean operations in constant evaluation

### LINKED ISSUE #195: Change all `map<k, v>` occurences to `map<K, V>` to conform to `grammar.ohm`

Gusarich:
resolves tact-lang/tact-docs#195 

I didn't cover that with new tests yet, going to do it soon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Update the `dump` function to handle addresses: PR [#175](https://github.com/tact-lang/tact/pull/175)
 - The implicit empty `init` function is now present by default in the contract if not declared: PR [#167](https://github.com/tact-lang/tact/pull/167)
-- Improved `Bool` reduction in constant expressions: PR [#195](https://github.com/tact-lang/tact/pull/195)
+- Improved `Bool` reduction in constant expressions: PR [#197](https://github.com/tact-lang/tact/pull/197)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Update the `dump` function to handle addresses: PR [#175](https://github.com/tact-lang/tact/pull/175)
 - The implicit empty `init` function is now present by default in the contract if not declared: PR [#167](https://github.com/tact-lang/tact/pull/167)
+- Improved `Bool` reduction in constant expressions: PR [#195](https://github.com/tact-lang/tact/pull/195)
 
 ### Fixed
 
--- a/src/types/resolveConstantValue.ts
+++ b/src/types/resolveConstantValue.ts
@@ -1,9 +1,10 @@
-import { Address, Cell, toNano } from "@ton/core";
+import { Address, Cell, Slice, toNano } from "@ton/core";
 import { enabledMasterchain } from "../config/features";
 import { CompilerContext } from "../context";
 import { ASTExpression, throwError } from "../grammar/ast";
 import { printTypeRef, TypeRef } from "./types";
 import { sha256_sync } from "@ton/crypto";
+import { getExpType } from "./resolveExpression";
 
 function reduceInt(ast: ASTExpression): bigint {
     if (ast.kind === 'number') {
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Update the `dump` function to handle addresses: PR [#175](https://github.com/tact-lang/tact/pull/175)
 - The implicit empty `init` function is now present by default in the contract if not declared: PR [#167](https://github.com/tact-lang/tact/pull/167)
+- Improved `Bool` reduction in constant expressions: PR [#195](https://github.com/tact-lang/tact/pull/195)
```
byakuren-hijiri:
```suggestion
- Improved `Bool` reduction in constant expressions: PR [#197](https://github.com/tact-lang/tact/pull/197)
```

src/types/resolveConstantValue.ts
```diff
@@ -59,22 +60,51 @@ function reduceInt(ast: ASTExpression): bigint {
...
+                    } else if (leftType.name === 'String' && rightType.name === 'String') {
+                        return reduceString(ast.left) === reduceString(ast.right);
+                    } else if (leftType.name === 'Int' && rightType.name === 'Int') {
+                        return reduceInt(ast.left) === reduceInt(ast.right);
+                    } else if (leftType.name === 'Bool' && rightType.name === 'Bool') {
+                        return reduceBool(ast.left, ctx) === reduceBool(ast.right, ctx);
+                    } else if (leftType.name === 'Slice' && rightType.name === 'Slice') {
+                        return reduceSlice(ast.left).asCell().equals(reduceSlice(ast.right).asCell());
+                    }
+                } else if (leftType.kind === 'null' && rightType.kind === 'null') {
```
byakuren-hijiri, line 115: What is the use case for this? Could you provide a minimal example please?

src/types/resolveConstantValue.ts
```diff
@@ -59,22 +60,51 @@ function reduceInt(ast: ASTExpression): bigint {
...
+            const rightType = getExpType(ctx, ast.right);
+            if (ast.op === '>') {
+                return reduceInt(ast.left) > reduceInt(ast.right);
+            } else if (ast.op === '<') {
+                return reduceInt(ast.left) < reduceInt(ast.right);
+            } else if (ast.op === '>=') {
+                return reduceInt(ast.left) >= reduceInt(ast.right);
+            } else if (ast.op === '<=') {
+                return reduceInt(ast.left) <= reduceInt(ast.right);
+            } else if (ast.op === '==') {
```
byakuren-hijiri, line 100: How about `!=`?

src/types/resolveConstantValue.ts
byakuren-hijiri: Ideally, we should extend the [`constants.tact`](https://github.com/tact-lang/tact/blob/7ef2ae17baef890d6489688d523bcdf77ce7b561/src/test/features/constants.tact) test

src/types/resolveConstantValue.ts
```diff
@@ -126,6 +156,24 @@ function reduceCell(ast: ASTExpression): Cell {
     throwError('Cannot reduce expression to a constant Cell', ast.ref);
 }
 
+function reduceSlice(ast: ASTExpression): Slice {
+    if (ast.kind === 'op_static_call') {
```
byakuren-hijiri: I would suggest defining the `isSlice` function for clarity:

```typescript
function isSlice(ast: ASTExpression): bool {
    return (ast.kind === 'op_static_call') && (ast.name === 'slice') && (ast.args.length === 1);
}
```

src/types/resolveConstantValue.ts
byakuren-hijiri: Ah, I see mention about the tests in the description.

CHANGELOG.md
anton-trunov: Let's document here what exactly is allowed in constant expressions after this PR is merged (not only the new stuff, but everything) -- this is going to go into the docs, see https://github.com/tact-lang/tact-docs/issues/152

Gusarich:
Current implementation doesn't actually work and throws errors on compiling new test cases.
I'll investigate the problem and try to fix it.

anton-trunov:
The issue here is that constant evaluation is performed on a non-typechecked AST. So when you call `getExpType` to dispatch on the type of `==` / `!=` operators, there is no actual type information. The proper fix here would be to refactor the compiler to perform constant evaluation after typechecking (and to actually do this refactoring we need a new type describing typed ASTs).

anton-trunov:
Looks like this PR should be closed now? Because the architecture of the constant evaluator has changed a lot in #432. Of course, we can still improve a lot of things in the constant evaluator, adding support for slices would be especially nice.

Gusarich:
yes, right

PR #198, opened 2024-03-27
Implement local type inference for `let` statements

### LINKED ISSUE #161: Local type inference
For instance, in many cases `let`-expressions don't need type ascriptions.

Gusarich:
resolves #161


src/grammar/grammar.ohm
```diff
@@ -98,7 +98,8 @@ Tact {
               | StatementRepeat
               | StatementUntil
     StatementBlock = "{" Statement* "}"
-    StatementLet = let id ":" Type "=" Expression ";"
+    StatementLet = let id ":" Type "=" Expression ";" --withType
+                 | let id "=" Expression ";" --withoutType
```
anton-trunov: How about we just make the type ascription optional?
```suggestion
    StatementLet = let id (":" Type)? "=" Expression ";"
```

anton-trunov:
Let's fix the merge conflict in CHANGELOG and move the tests to their new location

Gusarich:
@anton-trunov done

src/test/e2e-emulated/contracts/local-type-inference.tact
anton-trunov, line 1: We should cover more primitive types in tests:
- `Slice`,
- `Cell`,
- `Builder`,
- `String`,
- `StringBuilder`,
- `StateInit`

and also the compound types:
- maps,
- structs,
- optionals

src/grammar/clone.ts
```diff
@@ -28,7 +28,6 @@ export function cloneNode<T extends ASTNode>(src: T): T {
     } else if (src.kind === "statement_let") {
         return cloneASTNode({
             ...src,
-            type: cloneASTNode(src.type),
```
anton-trunov, line 31: you are sure we won't run into issues in the future if we just ignore the type here?

for instance, `def_argument` has the following:
```
    } else if (src.kind === "def_argument") {
        return cloneASTNode({
            ...src,
            type: cloneASTNode(src.type),
        });
```

I'd say we should either clone all types or not clone any.

anton-trunov (changes requested):
Look good! Let's add some more tests (see my comment below) and there is one concern with AST cloning

anton-trunov:
There is a case when local type inference should break: `let foo = null`. It's not clear what type to assign to `foo` in this case, it can be any optional type, e.g. `Int?` or any map type, like `map<Int, Int>` or `map<Address, Cell>`, etc. We should add negative tests for things like that.

src/grammar/grammar.ts
anton-trunov, line 1: let's fix the `StatementLet` parameter names too:
```
StatementLet(_letKwd, id, _optColon, optType, _equals, expression, _semicolon) {
```

src/grammar/clone.ts
Gusarich: those aren't cloned in `ASTField`, `ASTConstant`, `ASTFunction` and `ASTNativeFunction` too. Should I add that type cloning to all of these?

src/grammar/grammar.ts
Gusarich: What's wrong with these?

anton-trunov: Those are currently named `_colon` and `type`, but this PR makes them optional, and everywhere else in `grammar.ts` optional items are prefixed with `opt`

src/grammar/clone.ts
anton-trunov: good point, let's clone them all

Gusarich:
@anton-trunov everything's done!

anton-trunov (approved):
Let's go!

PR #208, opened 2024-03-29
Fix checks for key existance in objects

### LINKED ISSUE #181: Tact rejects user-defined function names is they coincide builtin methods like `valueOf`
Static functions cannot be named as [`Object.prototype` methods](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Object_prototypes). This results in a minor issue: the transpiler rejects valid user-defined functions named, for example, `toString`. The same problem should appear for static constants as well.

Consider a minimal example:
```
import "@stdlib/deploy";

fun valueOf(a: Int, b: Int): Int {
    return (a + b) / 2;
}

contract SampleTactContract with Deployable {
    init() {
    }
    get fun result(): Int {
        return valueOf(1, 10);
    }
}
```

Compiling with tact commit `7ed5e05dc0fcbc880f76fe326ea03aae69a3d966` gives the following error:
```
💼 Compiling project sample...
Tact compilation failed
Error: contract.tact:3:1: Static function valueOf already exists
Line 3, col 1:
  2 |
> 3 | fun valueOf(a: Int, b: Int): Int {
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  4 |     return (a + b) / 2;
```

Expected: successful compilation.

## Solution
It was found while working on #133.

These issues should be fixed in [the following part of code](https://github.com/tact-lang/tact/blob/7ed5e05dc0fcbc880f76fe326ea03aae69a3d966/src/types/resolveDescriptors.ts#L1109) and [here](https://github.com/tact-lang/tact/blob/7ed5e05dc0fcbc880f76fe326ea03aae69a3d966/src/types/resolveDescriptors.ts#L1112) excluding methods of `Object.prototype` when addressing to elements of `staticFunctions` and `staticConstants`.

### LINKED ISSUE #182: Incorrect error message when calling a non-existent function
Tact generates incorrect error messages when calling a non-existent function named as one of [`Object.prototype` methods](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Object_prototypes). This issue is similar to #181.

Consider the following example:
```
import "@stdlib/deploy";

contract SampleTactContract with Deployable {
    init() {
        toString(); // non-existent function
    }
}
```

Result:
```
💼 Compiling project sample...
Tact compilation failed
TypeError: f.resolve is not a function
    at resolveStaticCall (/home/bh/Dev/tact/dist/types/resolveExpression.js:222:28)
    at resolveExpression (/home/bh/Dev/tact/dist/types/resolveExpression.js:440:16)
    at processStatements (/home/bh/Dev/tact/dist/types/resolveStatements.js:166:61)
    at processFunctionBody (/home/bh/Dev/tact/dist/types/resolveStatements.js:252:17)
    at resolveStatements (/home/bh/Dev/tact/dist/types/resolveStatements.js:302:19)
    at precompile (/home/bh/Dev/tact/dist/pipeline/precompile.js:24:53)
    at build (/home/bh/Dev/tact/dist/pipeline/build.js:56:43)
    at Object.run (/home/bh/Dev/tact/dist/node.js:53:47)
    at /home/bh/Dev/tact/bin/tact:25:20
    at Object.<anonymous> (/home/bh/Dev/tact/bin/tact:30:3)
```

Expected:
```
💼 Compiling project sample...
Tact compilation failed
Error: /home/bh/Dev/TON-Projects/tact-template/sources/contract.tact:9:9: Static function "toString" does not exist
Line 9, col 9:
   8 |     init() {
>  9 |         toString();
               ^~~~~~~~~~
  10 |     }
```

It was found while working on #133.

Gusarich:
Closes #181 and closes #182 

* [x] I have updated CHANGELOG.md
* [x] I have added unit tests to demonstrate the contribution is correctly implemented
* [x] I have run all the tests locally and no test failure was reported
* [x] I did not introduce unrelated formatting/whitespace changes in the code base
* [x] I did not do unrelated and/or undiscussed refactorings

It was a bug with how existance of object keys was checked.
Accessing key by doing `myObject[key]` also includes built-in keys like `toString` and `valueOf` and results problems afterwards. In such cases `.hasOwnProperty` method should be used.

UPD: `Object.prototype.hasOwnProperty.call` has to be used instead of `.hasOwnProperty`, as explained in https://stackoverflow.com/a/39283005. This also was reported by ESLint before [fix](https://github.com/tact-lang/tact/pull/208/commits/c7c29c6d794cea135f7233be8f2755b3a57398cd).


src/types/stmts/case-10.tact
```diff
@@ -0,0 +1,14 @@
...
+
+trait BaseTrait { }
+
+fun toString(): Int {
+    return 123;
+}
+
+contract SampleTactContract {
+    init() {
+        toString(); // non-existent function
```
anton-trunov:
```suggestion
        toString();
```

src/generator/writers/writeExpression.ts
```diff
@@ -471,7 +471,7 @@ export function writeExpression(f: ASTExpression, ctx: WriterContext): string {
     if (f.kind === 'op_static_call') {
 
         // Check global functions
-        if (GlobalFunctions[f.name]) {
+        if (Object.prototype.hasOwnProperty.call(GlobalFunctions, f.name)) {
```
anton-trunov: Writing `Object.prototype.hasOwnProperty.call` whenever we need to access a key of a simple map does not look to enticing. I'd propose to declare `GlobalFunctions` a `Map` (as well as the other globals):

```ts
export const GlobalFunctions:Map<string, AbiFunction> = new Map([
    ["ton", { ... }],
    ["pow", { ... }],
    ...
]);
```

 and then use the `.get(...)` method for access here and below. In this case `GlobalFunctions.get("toString")` is guaranteed to return `undefined`.

anton-trunov (changes requested):

src/generator/writers/writeExpression.ts
anton-trunov: In this case, it should be `.has` instead of `.get`, of course

Gusarich: This suggestions looks good. There are actually many `Map` objects being used in other parts of the code, so it will fit well. I'll update the PR soon.

src/types/resolveDescriptors.ts
```diff
@@ -160,7 +160,7 @@ export function resolveDescriptors(ctx: CompilerContext) {
     //
 
     for (const a of ast.types) {
-        if (types[a.name]) {
+        if (Object.prototype.hasOwnProperty.call(types, a.name)) {
```
anton-trunov: Let's also use `Map` for `types`

src/types/resolveStatements.ts
```diff
@@ -46,7 +46,7 @@ function removeRequiredVariable(name: string, src: StatementContext): StatementC
 }
 
 function addVariable(name: string, ref: TypeRef, src: StatementContext): StatementContext {
-    if (src.vars[name]) {
+    if (Object.prototype.hasOwnProperty.call(src.vars, name)) {
```
anton-trunov: Let's use `Map` for `vars`

anton-trunov (changes requested):

Gusarich:
@anton-trunov done!

anton-trunov (approved):
Awesome, thanks!

PR #212, opened 2024-03-29
Implement try-catch statements

### LINKED ISSUE #209: Try-catch statements in Tact
Currently there are no try-catch statements in Tact, but it won't be too hard to add them because FunC has them since 0.4.0.

Suggested syntax:
```
try {
    // Code block where exceptions may occur
} catch (errorCode: Int) {
    // Handle the specific error code
}
```

Gusarich:
Closes #209 

* [x] I have updated CHANGELOG.md
* [x] I have added unit tests to demonstrate the contribution is correctly implemented
* [x] I have run all the tests locally and no test failure was reported
* [x] I did not introduce unrelated formatting/whitespace changes in the code base
* [x] I did not do unrelated and/or undiscussed refactorings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reserve mode constants in `@stdlib/reserve`, namely `ReserveExact`, `ReserveAllExcept`, `ReserveAtMost`, `ReserveAddOriginalBalance`, `ReserveInvertSign`, `ReserveBounceIfActionFail`: PR [#173](https://github.com/tact-lang/tact/pull/173)
 - JSON Schema for `tact.config.json`: PR [#194](https://github.com/tact-lang/tact/pull/194)
 - Display an error for integer overflow at compile-time: PR [#200](https://github.com/tact-lang/tact/pull/200)
+- Try-Catch statements: PR [#212](https://github.com/tact-lang/tact/pull/212)
 - Non-modifying `StringBuilder`'s `concat` method for chained string concatenations: PR [#217](https://github.com/tact-lang/tact/pull/217)
 - `toString` extension function for `Address` type: PR [#224](https://github.com/tact-lang/tact/pull/224)
 - Bitwise XOR operation (`^`): PR [#238](https://github.com/tact-lang/tact/pull/238)
--- a/src/generator/writers/writeFunction.ts
+++ b/src/generator/writers/writeFunction.ts
@@ -169,16 +169,16 @@ export function writeStatement(
         });
         ctx.append(`}`);
         return;
-    } else if (f.kind === 'statement_try') {
+    } else if (f.kind === "statement_try") {
         ctx.append(`try {`);
         ctx.inIndent(() => {
             for (const s of f.statements) {
                 writeStatement(s, self, returns, ctx);
             }
         });
-        ctx.append('} catch (_) { }');
+        ctx.append("} catch (_) { }");
         return;
-    } else if (f.kind === 'statement_try_catch') {
+    } else if (f.kind === "statement_try_catch") {
         ctx.append(`try {`);
         ctx.inIndent(() => {
             for (const s of f.statements) {
--- a/src/generator/writers/writeFunction.ts
+++ b/src/generator/writers/writeFunction.ts
@@ -169,6 +169,30 @@ export function writeStatement(
         });
         ctx.append(`}`);
         return;
+    } else if (f.kind === 'statement_try') {
+        ctx.append(`try {`);
+        ctx.inIndent(() => {
+            for (const s of f.statements) {
+                writeStatement(s, self, returns, ctx);
+            }
+        });
+        ctx.append('} catch (_) { }');
+        return;
+    } else if (f.kind === 'statement_try_catch') {
+        ctx.append(`try {`);
+        ctx.inIndent(() => {
+            for (const s of f.statements) {
+                writeStatement(s, self, returns, ctx);
+            }
+        });
+        ctx.append(`} catch (_, ${id(f.catchName)}) {`);
+        ctx.inIndent(() => {
+            for (const s of f.catchStatements) {
+                writeStatement(s, self, returns, ctx);
+            }
+        });
+        ctx.append(`}`);
+        return;
     }
 
     throw Error("Unknown statement kind");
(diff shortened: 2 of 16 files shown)


src/test/features/try-catch.tact
```diff
@@ -0,0 +1,35 @@
+contract TryCatchTester {
+    receive() {}
+
+    get fun testTryCatch1(): Int {
+        try {
+            throw(101);
```
anton-trunov, line 34: checking that control-flow is not local here:
```suggestion
            throw(101);
            return 42;
```

src/test/features/try-catch.tact
```diff
@@ -0,0 +1,35 @@
...
+        }
+        return 7;
+    }
+
+    get fun testTryCatch2(): Int {
+        try {
+            throw(101);
```
anton-trunov, line 42: checking that control-flow is not local here:
```suggestion
            throw(101);
            return 42;
```

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `log2` and `log` math functions in `@stdlib/math`: PR [#166](https://github.com/tact-lang/tact/pull/166)
 - Reserve mode constants in `@stdlib/reserve`, namely `ReserveExact`, `ReserveAllExcept`, `ReserveAtMost`, `ReserveAddOriginalBalance`, `ReserveInvertSign`, `ReserveBounceIfActionFail`: PR [#173](https://github.com/tact-lang/tact/pull/173)
 - JSON Schema for `tact.config.json`: PR [#194](https://github.com/tact-lang/tact/pull/194)
+- Try-Catch statements: PR [#212](https://github.com/tact-lang/tact/pull/212)
```
anton-trunov, line 16:
```suggestion
- `try...catch` statements: PR [#212](https://github.com/tact-lang/tact/pull/212)
```

src/types/resolveStatements.ts
anton-trunov, line 1: We are going to need some negative (`test-fail`) typechecker tests, for instance:
- the `catch`-variable scoping is violated, in other words it is used outside of the `catch`-block (the variable is used inside `try`, the variable is used before or after `try...catch`),
- the type of the `catch`-variable is not `Int`

src/generator/writers/writeFunction.ts
```diff
@@ -136,6 +136,30 @@ export function writeStatement(f: ASTStatement, self: string | null, returns: Ty
...
         ctx.append(`}`);
         return;
+    } else if (f.kind === 'statement_try') {
+        ctx.append(`try {`);
+        ctx.inIndent(() => {
+            for (const s of f.statements) {
+                writeStatement(s, self, returns, ctx);
+            }
+        });
+        ctx.append('} catch (_) { }');
```
anton-trunov: Is it valid FunC? Shouldn't `catch` have two arguments (`catch(_,_)`) in this case?

src/test/feature-try-catch.spec.ts
anton-trunov, line 1: We also need some state-specific test that cannot be tested using just getters. The FunC docs say:

> Unlike many other languages in the FunC try-catch statement, the changes made in the try block, in particular the modification of local and global variables, all registers' changes (i.e. c4 storage register, c5 action/message register, c7 context register and others) are discarded if there is an error in the try block and consequently all contract storage updates and message sending will be reverted.

So we need tests showing the state roll-backs indeed happen: at least one test for each part of the TVM state that should get reverted.

anton-trunov (changes requested):
This is a very nice feature! It needs to be properly tested, though: please see the comments below

src/grammar/grammar.ohm
```diff
@@ -119,6 +120,8 @@ Tact {
     StatementWhile = while "(" Expression ")" "{" Statement* "}"
     StatementRepeat = repeat "(" Expression ")" "{" Statement* "}"
     StatementUntil = do "{" Statement* "}" until "(" Expression ")" ";"
+    StatementTry = try "{" Statement* "}" ~catch --simple
+                 | try "{" Statement* "}" catch "(" id ")" "{" Statement* "}" --withCatch
```
anton-trunov, line 124: why don't we make the catch part optional (`?`) this should help deduplicate the grammar a bit

src/generator/writers/writeFunction.ts
Gusarich: we can do it this way because of FunC's tensor types.
`_` in this case is a pair that consists of both `catch` arguments.

I did it this way because we don't need to access values themselves here.

src/grammar/grammar.ohm
Gusarich: I thought about that at first, but to match the style and logic of grammar definitions of other statements, I did it as I did.

novusnota: Agree with @Gusarich here. If that's needed, we may later enhance the whole grammar file at once

anton-trunov: sounds reasonable!

src/types/resolveStatements.ts
```diff
@@ -46,6 +46,7 @@ function removeRequiredVariable(name: string, src: StatementContext): StatementC
 }
 
 function addVariable(name: string, ref: TypeRef, src: StatementContext): StatementContext {
+    console.log('addVariable', name, ref, src);
```
anton-trunov: looks like some debugging leftovers

src/types/resolveStatements.ts
```diff
@@ -304,6 +305,7 @@ function processStatements(statements: ASTStatement[], sctx: StatementContext, c
             ctx = r.ctx;
 
             // Process catchName variable for exit code
+            console.log(initialCtx.vars, s.catchName)
```
anton-trunov: looks like some debugging leftovers

anton-trunov (changes requested):
Let's add a negative typechecker test with the case when the type of the catch-variable is not `Int`

Gusarich:
> Let's add a negative typechecker test with the case when the type of the catch-variable is not `Int`

how exactly can it be not `Int`?

src/test/feature-try-catch.spec.ts
Gusarich: done in https://github.com/tact-lang/tact/pull/212/commits/c97369ec24093bcf6166f61e929db2ccb197f2aa

anton-trunov:
>  how exactly can it be not Int?

@Gusarich try to use it in a context where it is not Int, for instance as a string

anton-trunov:
There are some merge conflicts that need to be resolved

Gusarich:
@anton-trunov I think that's it!

anton-trunov (approved):
Perfect!

PR #230, opened 2024-04-03
`break` in loop control

### LINKED ISSUE #114: `break` and `continue`
Currently it's not possible to use `break` from within a loop.

vitorpy:
Closes #114

Initial draft to validate the approach.

Basically, from here, I'd like to:

- [ ] expand `resolveStatements.ts` to ensure break is always in a loop scope
- [ ] include a failed test case for break outside of a loop
- [ ] include a sequence of test cases for each loop type

If this approach sounds sensible, I'll build from here, otherwise I can drop this.

Checklist:
----------

- [ ] I have added unit tests to demonstrate the contribution is correctly implemented


src/generator/writers/writeFunction.ts
```diff
@@ -83,6 +83,9 @@ export function writeStatement(
             }
         }
         return;
+    } else if (f.kind === "statement_break") {
+        ctx.append(`break ;`);
```
anton-trunov, line 87: FunC does not support `break` at the moment, unfortunately

PR #237, opened 2024-04-10
Constant evaluation of `newAddress`

### LINKED ISSUE #214: `newAddress()` should be const eval
## Problem

```
contract A {
    myAddress: Address = newAddress(0, 0);
    receive("set") {
        self.myAddress = sender();
    }
}
```

gives error `Cannot reduce expression to a constant Address`. In order to avoid this any `Address` type variable must be initialized in the `init()` section.

## Proposal

`newAddress()` should be const eval. This would allow more consistent code, i.e all data types (int, bool, Address) can be initiated in the declaration section of the contract.

Gusarich:
Closes #214 

also covered constants with test. for some reason there were no checks for their values.


src/types/resolveConstantValue.ts
```diff
@@ -121,6 +121,23 @@ function reduceAddress(ast: ASTExpression, ctx: CompilerContext): Address {
...
                 return address;
             }
+        } else if (ast.name === "newAddress") {
+            if (ast.args.length === 2) {
+                const wc = reduceInt(ast.args[0]);
+                const addr = Buffer.from(
+                    reduceInt(ast.args[1]).toString(16),
+                    "hex",
+                );
+                if (!enabledMasterchain(ctx)) {
```
anton-trunov, line 137: We are not checking the workchain is either `-1` or `0` like in the code above. Why is that?

src/types/resolveConstantValue.ts
```diff
@@ -121,6 +121,23 @@ function reduceAddress(ast: ASTExpression, ctx: CompilerContext): Address {
                 }
                 return address;
             }
+        } else if (ast.name === "newAddress") {
```
anton-trunov, line 124: The first and second arguments are not necessarily literals, those can be constant expressions too, so we need to recursively process them. (Even if we don't do it elsewhere)

anton-trunov (changes requested):

src/types/resolveConstantValue.ts
Gusarich: > The first and second arguments are not necessarily literals, those can be constant expressions too, so we need to recursively process them. (Even if we don't do it elsewhere)

but that actually happens on the next lines in the code. `reduceInt(ast.args[0])` and `reduceInt(ast.args[1])`.

anton-trunov: you are right, I misread the code, sorry

anton-trunov (approved):
LGTM

PR #265, opened 2024-04-18
Fix operations precendence

### LINKED ISSUE #263: Precendence of bitwise operators is incorrect
The `<<`, `>>`, `&` and `|` operators have the same precedence which is incorrect: `<<` and `>>` should have the same precedence, then `&` has a lower precedence, then (not yet implemented) `^` and then `|`. See the operator precedence table here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence.

Also, the bit shift operators precedence should be fixed with respect to comparisons (shifts have greater precedence).

Gusarich:
Closes #263 


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `@stdlib/stoppable` now imports `@stdlib/ownable` so the programmer does not have to do it separately: PR [#193](https://github.com/tact-lang/tact/pull/193)
 - Support escape sequences for strings (`\\`, `\"`, `\n`, `\r`, `\t`, `\v`, `\b`, `\f`, `\u{0}` through `\u{FFFFFF}`, `\u0000` through `\uFFFF`, `\x00` through `\xFF`): PR [#192](https://github.com/tact-lang/tact/pull/192)
 - `newAddress` function now evaluates to a constant value if possible: PR [#237](https://github.com/tact-lang/tact/pull/237)
+- Operation precendence now matches with JavaScript rules: PR [#265](https://github.com/tact-lang/tact/pull/265)
```
anton-trunov: people in general would expect certain operator precedences, so this is definitely fixed, not changed

src/test/features/precendence.tact
```diff
@@ -0,0 +1,23 @@
+import "@stdlib/deploy";
+
+contract PrecendenceTester with Deployable {
+    init() {
+
+    }
+
+    get fun test1(): Int {
+        return 5 & 6 | 1 << 5 + 11 * 3 % 12 >> 11;
```
anton-trunov: we should also test a different order for `&` and `|`, because if we parse it from left to right it's expected we get `(5 & 6) | 1`, let's have also something like `x | y & z`

CHANGELOG.md
```suggestion
- Operation precendence for bitwise operators, equality and comparisons now matches common languages, like JavaScript: PR [#265](https://github.com/tact-lang/tact/pull/265)
```

src/test/feature-precendence.spec.ts
anton-trunov, line 1: let's move these tests to `math`, because operator precedence is not really a feature

anton-trunov (changes requested):
Good stuff! Let's just do a few small adjustments

anton-trunov:
Perfect!

PR #267, opened 2024-04-19
Add `pow` and `pow2` to stdlib

### LINKED ISSUE #264: Support exponentiation `pow` function at runtime
Currently it is only supported at compile time only, which is confusing for the users

Gusarich:
Closes #264 


Gusarich:
I will resolve merge conflicts now

src/test/feature-math.spec.ts
anton-trunov, line 1: What happens when we call `pow2(-1)`? And the same question for `pow` as well. We need to make sure the semantics of TVM instruction `POW2` and manually implemented `pow` are coherent.

src/generator/writers/writeStdlib.ts
```diff
@@ -1426,4 +1426,24 @@ export function writeStdlib(ctx: WriterContext) {
...
+    ctx.fun(`__tact_pow`, () => {
+        ctx.signature(`int __tact_pow(int base, int exp)`);
+        ctx.context("stdlib");
+        ctx.body(() => {
+            ctx.write(`
+                int result = 1;
+                repeat (exp) {
+                    result *= base;
+                }
+                return result;
```
anton-trunov, line 1440: Would [exponentiation by squaring](https://en.wikipedia.org/wiki/Exponentiation_by_squaring) be more gas efficient?

anton-trunov (changes requested):
way to go! left a few comments, though

src/generator/writers/writeStdlib.ts
Gusarich: yes, right. but only for `exp` larger than some X, probably.
maybe I should do some benchmarks on that and add a simple `if-else` depending on `exp` with two different implementations?

anton-trunov: right, benchmarking sounds like a great idea, if for smaller exponents the gas difference is not that big we can use the fast exp algorithm without a condition

src/test/feature-math.spec.ts
Gusarich: `POW2` expects an argument that is `>= 0` and throws an error 5 if you provide a negative number.
I think the correct solution for us will be to check argument in compile-time and throw a compilation error in such cases.

anton-trunov: sounds good

Gusarich: I've decided to [add throwing in the function itself](https://github.com/tact-lang/tact/pull/267/commits/472ae69f51ee09d41ba3d467bfa9295700337e77). the reason is that there would still be cases where we don't know `exp` in compile-time.

src/generator/writers/writeStdlib.ts
Gusarich: I've done some benchmarking and have come to the conclusion that in case with smart contracts and integer numbers, it's more efficient to leave the `pow` calculation as it is currently done in this PR. The reason is that, due to integer number limits and the absence of real values, we will never need to calculate `pow` with an `exp` large enough to outperform a simple loop using such optimizations.

anton-trunov (approved):
LGTM!

PR #271, opened 2024-04-20
Upgrade and fix debugging feature

### LINKED ISSUE #260: `dump` should output its source code location information
I guess it should be the contract file name, the line and column numbers

### LINKED ISSUE #270: Incorrect typechecking for `dump` function
Tact allows `Cell`, `Slice`, `Builder` and `StringBuilder` types to be provided in `dump()` function but tries to format them using `int_to_string` which is completely incorrect.

For example, such line:
```tact
dump(beginCell().storeInt(123, 32).endCell());
```
Produces the following FunC compilation error:
```
Func compilation error: /Users/daniil/Coding/tact/src/test/features/output/debug_Debug.code.fc:39:86: error: cannot apply function __tact_int_to_string : int -> slice to arguments of type cell: cannot unify type cell with int
      __tact_debug_str(__tact_int_to_string(end_cell(store_int(begin_cell(), 123, 32))));
```

Gusarich:
Closes #260 and closes #270


stdlib/std/debug.tact
```diff
@@ -8,7 +8,4 @@ native nativeThrowWhen(code: Int, condition: Bool);
 native throw(code: Int);
 
 @name(throw_unless)
-native nativeThrowUnless(code: Int, condition: Bool);
-
-@name(dump_stack)
-native dumpStack();
```
anton-trunov, line 14: Hmm, we cannot remove functions from stdlib without breaking backwards compatibility. Is there a way to keep `dumpStack` here?

src/abi/global.ts
```diff
@@ -214,32 +215,40 @@ export const GlobalFunctions: Map<string, AbiFunction> = new Map([
                     return `${ctx.used("__tact_nop")}()`;
                 }
                 const arg = args[0];
+
+                const filePath = ref.file
+                    ? path.basename(ref.file!)
+                    : "unknown";
+                const lineCol = ref.interval.getLineAndColumn();
+                const debugPrint = `[DEBUG] File ${filePath}, Line ${lineCol.lineNum}, Column ${lineCol.colNum}`;
```
anton-trunov: let's do it in the standard way: `filename:line:column` -- most text editors can then open the file at this exact line and column when you click on this in your terminal, for instance

anton-trunov (changes requested):
Awesome stuff!

stdlib/std/debug.tact
Gusarich: How is this breaking backwards compatibility? The function is still available the same way as before, it’s just defined natively in language rather than in stdlib.

Gusarich: Oh, looks like throwUnless was removed somehow. Sorry for that, probably I selected that line accidentally.

anton-trunov: No, no. `throwUnless` is fine. I mean before this change the user needed to explicitly import `std/debug.tact` and only then `dumpStack` would be available. To make it backwards compatible we'd need to make `dumpStack` (defined internally) available only after the proper input.

anton-trunov: Ah, all of the modules in the `std/` directory are automatically imported. So, this does not change anything. Sorry, my bad

Gusarich: Having it in stdlib will make line printing impossible. Maybe we can just leave dumpStack as it was before? Anyway it is a low-level thing that should not be used commonly by devs in Tact.

anton-trunov: I mean your initial approach is the right one. Let's add it in the language and remove from stdlib (it actually won't change anything). With Tact 1.2 I cannot re-use the `dumpStack` name:

```
fun dumpStack() {
    dump("foo");
}
```

will result in this error message

```
/main.tact:1:1: Static function dumpStack already exists 
Line 1, col 1: 
> 1 | fun dumpStack() { 
      ^~~~~~~~~~~~~~~~~ 
  2 |     dump("foo"); 
```

I was worried about this

src/test/feature-debug.spec.ts
```diff
@@ -29,13 +29,23 @@ describe("feature-debug", () => {
             res.indexOf("=== DEBUG LOGS ===") + 19,
             res.indexOf("=== VM LOGS ===") - 2,
         );
-        expect(debugLogs).toStrictEqual(`stack(2 values) : 10000000000 () 
+
+        expect(debugLogs)
+            .toStrictEqual(`[DEBUG] File debug.tact, Line 10, Column 9
```
anton-trunov: you'll also probably need to fix the snapshot here

anton-trunov (approved):

src/abi/global.ts
```diff
@@ -174,32 +175,40 @@ export const GlobalFunctions: Map<string, AbiFunction> = new Map([
                     return `${ctx.used("__tact_nop")}()`;
                 }
                 const arg = args[0];
+
+                const filePath = ref.file
+                    ? path.basename(ref.file!)
```
anton-trunov: I think we should print the filename relative to the current working directory. This would make it easier to access this file from the terminal

anton-trunov (changes requested):

src/abi/global.ts
anton-trunov: image

anton-trunov: this is the difference between Tact's `dump` and the output from Jest

Gusarich:
I've removed `dump` call from `bugs.tact` test because it wasn't checked there anyway and was just adding inconsistency in snapshots with this new debug update.

anton-trunov (dismissed):

src/test/feature-debug.spec.ts
```diff
@@ -29,13 +31,24 @@ describe("feature-debug", () => {
             res.indexOf("=== DEBUG LOGS ===") + 19,
             res.indexOf("=== VM LOGS ===") - 2,
         );
-        expect(debugLogs).toStrictEqual(`stack(2 values) : 10000000000 () 
+
+        const filePath = path.normalize("src/test/features/debug.tact");
```
anton-trunov, line 34: Will this work on Windows? Perhaps we need to fix it in the tests

anton-trunov: Oh, sorry, was reviewing it from my phone and it seemed like the fix was applied in the main logic :)

anton-trunov (approved):

PR #272, opened 2024-04-20
Enables struct fields punning, i.e. `Foo {foo, bar}` is syntactic sugar for `Foo { foo: foo, bar: bar }`

### LINKED ISSUE #253: Struct fields punning, i.e. `{foo, bar}` should be syntactic sugar for `{ foo: foo, bar: bar }`
This makes code more concise, so we should support it. This concerns messages too.

vitorpy:
Closes #253

- [X] I have updated CHANGELOG.md
- [X] I have added unit tests to demonstrate the contribution is correctly implemented


src/grammar/grammar.ohm
```diff
@@ -193,7 +193,8 @@ Tact {
     ExpressionField = ExpressionValue "." id ~"("
     ExpressionCall = ExpressionValue "." id "(" ListOf<Expression, ","> ","? ")"
     ExpressionNew = id "{" ListOf<NewParameter, ","> ","? "}"
-    NewParameter = id ":" Expression
+    NewParameter = id ":" Expression --full
+                 | Expression        --punned
```
anton-trunov: it should be an identifier, not an expression

src/grammar/test/case-34.tact
```diff
@@ -0,0 +1,4 @@
+struct A {
```
anton-trunov, line 1: this feature does not apply to struct/message declarations, only to their instance constructions:

```
receive("foo") {
    let to: Address = newAddress(...);
    let value: Int = 0;
    let mode: Int = SendBounceIfActionFail;

    send(SendParameters {
        to,
        value,
        mode,
    }
}
```

vitorpy: Yep, sorry about that - I hadn't finished yet the unit tests, I should have left a note in the draft PR.

anton-trunov:
Looks like you go in the right direction. I'd like to have a couple more tests here that are not just grammar tests:

- one runtime test should go into https://github.com/tact-lang/tact/blob/main/src/test/features/ordering.tact with the corresponding test in https://github.com/tact-lang/tact/blob/main/src/test/feature-ordering.spec.ts (this essentially tests structures, can be renamed to `feature-struct` actually)
- a failing test for the typechecker when there is no identifier with the same name as a struct field: https://github.com/tact-lang/tact/tree/main/src/types/test-failed

src/types/test-failed/case-26.tact
```diff
@@ -0,0 +1,9 @@
+struct A {
+    x: Int;
+    y: Int;
+}
+
+const a: Int = 5;
+const b: Int = 6;
+
+const D: A = A { a, b };
```
anton-trunov: This should not fail: both `a` and `b` are in scope

src/test/features/ordering.tact
```diff
@@ -22,4 +22,21 @@ contract A with Deployable {
...
+}
+
+contract B with Deployable {
+
+    owner: Address;
+
+    init(owner: Address) {
+        self.owner = owner;
+    }
+
```
anton-trunov: Let's remove this: we don't use in the test

anton-trunov (changes requested):

src/types/test-failed/case-26.tact
vitorpy: So, should we rely on ordering to assign the fields? I thought this should fail as the fields (valid identifiers) are `x` and `y`, not `a` and `b`.

ie.

```
struct A {
    x: Int;
    y: Int;
}

const a: Int = 5;
const b: Int = 6;
const x: Int = 7;
const y: Int = 8;

const D: A = A { a, b } /* should fail as there are no fields a and b */
const E: A = A { x, y } /* should resolve to A{ x: 7, y: 8 } */
```

What am I missing?

anton-trunov: ah, sorry, my bad -- too much code review on Sunday :)

The error message is not the right one, though. The error should be at the instance definition point, not the struct definition. You need to add the following to `case-26.tact` (at the beginning of the file):

```
primitive Int;

trait BaseTrait {
    
}
```

anton-trunov: I'd also make it partially incorrect, for instance like so:

```suggestion
const D: A = A { x: a, b };
```

vitorpy: Hah, no worries at all! Thanks for taking the time to review it on a Sunday. I usually try do something different from the day job and playing with Tact and TON has been a favourite lately.

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `@stdlib/stoppable` now imports `@stdlib/ownable` so the programmer does not have to do it separately: PR [#193](https://github.com/tact-lang/tact/pull/193)
 - Support escape sequences for strings (`\\`, `\"`, `\n`, `\r`, `\t`, `\v`, `\b`, `\f`, `\u{0}` through `\u{FFFFFF}`, `\u0000` through `\uFFFF`, `\x00` through `\xFF`): PR [#192](https://github.com/tact-lang/tact/pull/192)
 - `newAddress` function now evaluates to a constant value if possible: PR [#237](https://github.com/tact-lang/tact/pull/237)
+- Enables struct fields punning, i.e. `{foo, bar}` is syntactic sugar for `{ foo: foo, bar: bar }`: PR [#272](https://github.com/tact-lang/tact/pull/272)
```
anton-trunov: I think this should be in the `Added` section above, since this feature wasn't there before

src/types/__snapshots__/resolveDescriptors.spec.ts.snap
```diff
@@ -258,6 +258,16 @@ Line 8, col 21:
 "
 `;
 
+exports[`resolveDescriptors should fail descriptors for case-26 1`] = `
+"<unknown>:13:14: Expected constant value, got A
```
anton-trunov: This is much closer now, however, I'd expect the `^~~` pointer at `b` and the error message saying that the `b` field is not present in the `A` struct

src/types/test-failed/case-26.tact
anton-trunov: @vitorpy Nice to hear that! Looking forward to have more contributions from you :)

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bitwise XOR operation (`^`): PR [#238](https://github.com/tact-lang/tact/pull/238)
 - `isEmpty` method for `Map` type: PR [#266](https://github.com/tact-lang/tact/pull/266)
 - `pow2` function: PR [#267](https://github.com/tact-lang/tact/pull/267)
+- Enables struct fields punning, i.e. `{foo, bar}` is syntactic sugar for `{ foo: foo, bar: bar }`: PR [#272](https://github.com/tact-lang/tact/pull/272)
```
```suggestion
- Struct fields punning, i.e. `{foo, bar}` is syntactic sugar for `{ foo: foo, bar: bar }`: PR [#272](https://github.com/tact-lang/tact/pull/272)
```

vitorpy:
Will fix conflicts.

vitorpy:
Changelog changes don´t look right, looking.

src/types/__snapshots__/resolveStatements.spec.ts.snap
```diff
@@ -360,6 +360,16 @@ Line 8, col 9:
 "
 `;
 
+exports[`resolveStatements should fail statements for case-36 1`] = `
+"<unknown>:14:26: Unknown fields "b" in type A
+Line 14, col 26:
+  13 | fun function() {
+> 14 |     let D: A = A { x: a, b };
```
anton-trunov, line 367: wait, how did it fix itself? is it because you moved the struct into a function? then it seems we have a bug in constant processing involving structs with wrong field names or something like that

vitorpy: Yep, it's because I moved it into a function. [Here's](https://github.com/tact-lang/tact/blob/86465ab4468fc1e55be89112561c9e80e3ae5341/src/types/resolveConstantValue.ts#L177) the relevant code. Sorry I didn't flag it - I thought that was the intended behaviour and I had missed it.

If const structs should be allowed then I can expand `resolveConstantValue` in a separate PR, if that's ok?

anton-trunov: Sure, I opened an issue: #284

anton-trunov (approved):
@vitorpy Thanks a lot for contributing yet another nice feature!

PR #287, opened 2024-04-23
Improved CLI

### LINKED ISSUE #136: CLI
There are still important suggestions in comments on this ticket. They will eventually be converted to issues.

Some docs, talks and guidelines on how to design CLI interfaces:
- Some nice CLI guidelines: https://clig.dev (btw, it suggests to use https://oclif.io as a CLI arguments parsing library).
- A talk by Adir Amsalem @ NodeTLV '23: [Write CLI applications with Node.js like it’s 2023](https://www.youtube.com/watch?v=fyrdp_AiE9U)

### LINKED ISSUE #1516: `--help` / `-h`: output CLI help
#287

### LINKED ISSUE #1517: Add `--check` to CLI
Stop the compilation pipeline right after typechecking Tact code (done in #287)

### LINKED ISSUE #1518: Add `--func` to CLI
Stop the compilation pipeline right after generating FunC code (done in #287)

### LINKED ISSUE #1519: Single contract projects without creating a config
CLI should support compiling single contract projects without creating a `tact.config.json` file (done in #287)

vitorpy:
Closes https://github.com/tact-lang/tact/issues/1516
Closes https://github.com/tact-lang/tact/issues/1517
Closes https://github.com/tact-lang/tact/issues/1518
Closes https://github.com/tact-lang/tact/issues/1519

Uses [meow](https://github.com/sindresorhus/meow) to rewrite the CLI. Things I added in this PR:

- [X] Compile single file
- [X] Implement `--func`
- [X] Implement `--check`
- [X] Test the release process to amke sure it packages properly 
- [X] Remove the old `bin/tact` file.
- [X] Update GitHub workflow

 Does this sounds reasonable?

- [X] I have updated CHANGELOG.md
- [X] I have added unit tests to demonstrate the contribution is correctly implemented


novusnota:
@vitorpy Hey, nice initiative! But adding oclif is a bit too much as it's a full-blown suite for making CLIs and it implicitly brings over its structure layout. This could've been the issue with the tests failing to find `./dist/grammar/` all of the sudden.

If you want to change the current argument parser library from `arg`, I'd suggest using [meow](https://github.com/sindresorhus/meow) as it's targets easy creation of single-command CLIs, like Tact compiler. Similar in spirit to [docopt](http://docopt.org/), you only need to specify your desired help message and then the tool will parse all the arguments and options based on that. [Here's an example of it in the app, which uses ESM](https://github.com/vadimdemedes/create-ink-app/blob/main/templates/ts/source/cli.tsx).

Additionally, when it comes to `--func` option, this was in the works of @anton-trunov, so we shall first align the work with him :)

vitorpy:
> If you want to change the current argument parser library from `arg`, I'd suggest using [meow](https://github.com/sindresorhus/meow) as it's targets easy creation of single-command CLIs, like Tact compiler. 

Cool - I haven't seen `meow` but it looks nice. I'll rewrite it sometime later this week. I agree with `oclif` being a bit bloated for something simple as the `tact` cli.

> Additionally, when it comes to `--func` option, this was in the works of @anton-trunov, so we shall first align the work with him :)

Yeah, that's why I started to look into this. I wanted to look at FunC outputs (and from there to Fift outputs).

anton-trunov:
I think it's ok to work on the `--func` CLI flag because I just did a quick-and-dirty hack to help me with hunting down a nasty bug related to the WASM version of the FunC compiler (I basically added an option like `funcOnly` for each project in  `tact.config.json` and `--func` just overrides all of those options).

anton-trunov:
Hey @vitorpy, do you think we can still manage do this CLI improvement before 1.3.0 release? Btw, I think we could also add `--check` CLI flag to stop the compilation pipeline right after typechecking contracts.

vitorpy:
Yes, let's aim for the 1.3 release. I should have some time this weekend to look into this. I'll add `--check` as well!

vitorpy:
@novusnota Hey, can you take a look? I haven't test this yet, but I want to make sure you're OK with the approach. `meow` is an ES module and with Tact being CommonJS, there's a little bit of import magic in the tact entry point script. It's not too ugly, IMHO, but please take a look.

If you're OK with the approach, I'll give it some testing, work out the kinks and publish this PR.

novusnota:
@vitorpy if you rename `bin/tact` into `bin/tact.mjs` and add the `bin: { "tact": "bin/tact.mjs" }` property to `package.json`, then you'll be able to use ESM imports there without actually modifying the CLI name — it will stay "tact" as is.

And seems like you forgot to add the `--config` flag. Other then that, splendid work!

vitorpy:
@novusnota Oh, thanks! I thought it wouldn't be OK to rename the tact bin. It looks so much cleaner now.

I kinda feel config_path should be an argument (being always required) but since a breaking change in the binary syntax doesn't make sense, I re-introduced the `--config` flag as a required flag.

vitorpy:
Ah,

```
      - name: Compare Tact version from CLI flag `--version` against package.json
        if: runner.os != 'Windows'
        run: |
          if [ "$(./bin/tact.mjs --version)" != "$(jq -r '.version' < package.json)" ];
          then false
          fi
```

This is now a redundant step in the GitHub workflow, as the version from `--version` is extracted from the `package.json`. Should I remove it?

package.json
```diff
@@ -29,17 +29,17 @@
   ],
   "main": "./dist/main.js",
   "bin": {
-    "tact": "./bin/tact"
+    "tact": "./bin/tact.mjs"
```
anton-trunov: Is there really no way to keep the `tact` name? It feels like having to call it `tact.msj` from command line makes UX a bit worse

bin/tact.mjs
```diff
@@ -0,0 +1,50 @@
...
+import * as main from "../dist/node.js";
+import meow from "meow";
+
+const cli = meow(
+  `
+Usage
+  $ tact --config <config_path>
+
+Options
+  --project <project_name>, -r <project_name>
```
anton-trunov:
```suggestion
  --project <project_name>, -p <project_name>  
```

We should also have some commentary here

bin/tact.mjs
```diff
@@ -0,0 +1,50 @@
...
+import meow from "meow";
+
+const cli = meow(
+  `
+Usage
+  $ tact --config <config_path>
+
+Options
+  --project <project_name>, -r <project_name>  
+  --version                                    Prints the current tact version
```
anton-trunov:
```suggestion
  --version                                    Prints the current Tact version
```

bin/tact.mjs
```diff
@@ -0,0 +1,50 @@
...
+  --func                                       Outputs the intermediate FunC code.
```
anton-trunov:
```suggestion
  --func                                       Outputs the intermediate FunC code and stops
```

bin/tact.mjs
```diff
@@ -0,0 +1,50 @@
...
+  --check                                      Performs type checking and stops.
```
anton-trunov:
```suggestion
  --check                                      Performs type checking and stops
```

bin/tact.mjs
```diff
@@ -0,0 +1,50 @@
...
+  --help                                       Displays this text
+
+Examples
+  $ tact --version
+  1.2.0
```
anton-trunov: can we read `package.json`'s version here?

bin/tact.mjs
```diff
@@ -0,0 +1,50 @@
...
+`,
+  {
+    importMeta: import.meta,
+    flags: {
+      config: { shortFlags: "c", type: "string", isRequired: true },
+      project: { shortFlag: "p", type: "string", isMultiple: true },
+      version: { shortflag: "v", type: "boolean" },
+      check: { shortflag: "c", type: "boolean" },
+      func: { shortFlag: "f", type: "boolean" },
```
anton-trunov: Conflicting short versions for `--config` and `--check`. I think `--check` and `--func` should not have short versions as I wouldn't expect those to be used very often by users

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `isEmpty` extension function for the `Map` type: PR [#266](https://github.com/tact-lang/tact/pull/266)
 - The `pow2` power function with base 2: PR [#267](https://github.com/tact-lang/tact/pull/267)
 - The `try` and `try-catch` statements: PR [#212](https://github.com/tact-lang/tact/pull/212)
+- The `--func` and `--check` flags: PR [#287](https://github.com/tact-lang/tact/pull/287)
```
```suggestion
- The `--func` and `--check` command-line flags: PR [#287](https://github.com/tact-lang/tact/pull/287)
```

src/pipeline/build.ts
```diff
@@ -26,6 +26,8 @@ export async function build(args: {
     project: VirtualFileSystem;
     stdlib: string | VirtualFileSystem;
     logger?: TactLogger | null | undefined;
+    checkOnly?: boolean;
+    func?: boolean;
```
anton-trunov: I'd prefer we modify `ConfigProject` so `checkOnly` and `func` become `config`'s fields. Otherwise the `build`'s function parameter list will be to large with time

.github/workflows/tact.yml
```diff
@@ -57,14 +57,14 @@ jobs:
       - name: Compare Tact version from CLI flag `--version` against package.json
         if: runner.os != 'Windows'
         run: |
-          if [ "$(./bin/tact --version)" != "$(jq -r '.version' < package.json)" ];
+          if [ "$(./bin/tact.mjs --version)" != "$(jq -r '.version' < package.json)" ];
```
anton-trunov: Would be nice to remove this check and reuse `getCompilerVersion` function in the CLI module

anton-trunov (changes requested):
Really nice!

anton-trunov:
We should also add some CLI tests to CI. At least one test per each CLI flag to demonstrate it it's there. You can put the data you need in this folder: https://github.com/tact-lang/tact/tree/main/test/tact-cli

package.json
novusnota: The end user won't call it `tact.mjs` — the name will stay `tact`, as is.

See [this](https://github.com/novusnota/tact-extracted-ls/blob/e273d453e2642aa2b900c19fc82bd53ddbc3cfdf/package.json#L28) for example:
image

Despite that the actual bin script there is called `bundle.js`, the user, who downloads this package from npm, receives only `tact-extracted-ls` :)

More details: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#bin

anton-trunov: Got it, thanks @novusnota. We then should fix it in CI

.github/workflows/tact.yml
```diff
@@ -57,14 +57,14 @@ jobs:
...
           then false
           fi
 
       - name: Tact CLI non-zero exit code
         if: runner.os != 'Windows'
         run: |
-          ! ./bin/tact --config test/tact-cli/tact.config.json
+          ! ./bin/tact.mjs --config test/tact-cli/tact.config.json
```
anton-trunov: This should still be called as `tact`, not `tact.mjs`

novusnota: Does it? npm performs the aliasing to `tact` only when we install the package from the registry. But the local use only has the present files, so `tact.mjs` it is

anton-trunov:
@vitorpy Lots of improvements in this PR! `--check` and `--func` are really a huge help when debugging Tact issues. There is one more thing that I'm personally missing: CLI should support compiling single contract projects without creating a `tact.config.json` file. Basically, I'd like to be able to write a `test.tact` file and compile it like so: `tact test.tact` (the default output folder could be `.` in this case -- this makes inspecting compilation results easier). Of course, `tact test.tact --check` should also work, etc.

.github/workflows/tact.yml
anton-trunov: Why not then install it locally from the repo and perform all the CLI tests using the installed version?

novusnota: Good point!

package.json
vitorpy: So, long story short, we can keep calling it `tact` but the code will be a tad more convulted (basically the cli code would live under `import('meow').then(() => {/* cli code here*/});` since `meow` is an ES module and `tact` is CommonJS. 

If that trade off is OK, I can rename it back to `tact`.

src/pipeline/build.ts
vitorpy: Will do.

package.json
novusnota: @vitorpy it's probably best to keep things simple for the end user, see: https://github.com/tact-lang/tact/pull/287#discussion_r1582045728

anton-trunov: @vitorpy Right, the script's name in the repo does not really matter as long as there is a way to call it `tact` from the command line. And in this case it will be accessible after a local package installation as @novusnota mentioned

anton-trunov:
Btw, I in addition to positive test mentioned here https://github.com/tact-lang/tact/pull/287#issuecomment-2081362002, it would be nice to also have some negative CLI tests showing what happens when something goes wrong, e.g.
- an unknown CLI flag is provided,
- there is the wrong file name,
- incorrect combination of CLI flags,
- ...

vitorpy:
Work in progress:

```
❯ tact hello.tact
(node:59433) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
💼 Compiling project main...
   > HelloWorld: tact compiler
   > HelloWorld: func compiler
FunC compiler crashed
RuntimeError: Aborted(LinkError: WebAssembly.instantiate(): Import #91 module="a" function="La": function import requires a callable). Build with -sASSERTIONS for more info.
    at abort (/home/vitorpy/code/tact/dist/func/funcfiftlib.js:9:8020)
    at /home/vitorpy/code/tact/dist/func/funcfiftlib.js:9:9953
```

Single file compilation works up to a certain point (`--check` and `--func` works). I'll probably take a couple days to debug this and add tests.

src/node.ts
```diff
@@ -5,30 +5,59 @@ import { createNodeFileSystem } from "./vfs/createNodeFileSystem";
...
+async function configForSingleFile(fileName: string): Promise<Config> {
+    return {
+        projects: [{
+            name: "main",
+            path: fileName,
+            output: process.cwd(),
+        }],
+        rootPath: process.cwd(),
+    };    
+}
```
novusnota: Neat! It would be nice to enable debugging by default for convenience — `options: { debug: true }`

anton-trunov:
> FunC compiler crashed

@vitorpy Do you still have the Tact source code that caused the FunC compiler to crash? I think this should be reported to the FunC team

anton-trunov (dismissed):
LGTM! Let's also wait for the feedback from @novusnota

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `isEmpty` extension function for the `Map` type: PR [#266](https://github.com/tact-lang/tact/pull/266)
 - The `pow2` power function with base 2: PR [#267](https://github.com/tact-lang/tact/pull/267)
 - The `try` and `try-catch` statements: PR [#212](https://github.com/tact-lang/tact/pull/212)
+- The `--func` and `--check` command-line flags: PR [#287](https://github.com/tact-lang/tact/pull/287)
```
anton-trunov: Oh, I missed that we have a richer CLI now

novusnota:
@vitorpy Nice work overall! If you don't mind, I'll refine it myself (just a little) in this PR in one hour or so.

Stay tuned for my commit until then :)

vitorpy:
> Stay tuned for my commit until then :)

Waiting for it!

I'm somewhat unhappy with `bin/tact`. Making it a ESM would be nicer and neater. But the CI uses `yarn link` and I had some weird errors. I got around most of them by shipping a little shell script to trigger `tact.mjs` but the CI also runs on Windows and I have no idea how to support that (nor access to a Windows machine to try). So I just went with naming it `bin/tact`.

novusnota:
> I had some weird errors ... on Windows

The error was related to you introducing a bash script, which is unaccessible on Windows (without additional installations of stuff, that is). And the previous one was caused by tact-template calling `./bin/tact`, while there was only `./bin/tact.mjs`. So, I decided to just leave the `bin/tact` as a CommonJS module for the time being. Later, we can try to update to ESM modules, but for now it's enough to just make the CLI robust and convenient for the end users.

Give me a bit more time, I'm testing various inputs to see if everything's working fine and simplifying things where I can

novusnota:
A brief list of changes:
1. Now help message shows if there are no inputs
2. Updated JSON Schema with new options and descriptions for them
3. Enhanced descriptions and help message
4. Fixed typo: `shortFlags` -> `shortFlag`
5. Removed redundant `rootPath` option from configuration & Zod schemas
6. Added links to docs and socials
7. Disallowed unknown flags
8. Meticulously checked various combinations of flags to ensure correctness (Later, it would be nice for our tests to do the same thing. Perhaps, using a fuzzer)
9. Updated CHANGELOG

src/node.ts
```diff
@@ -5,30 +5,73 @@ import { createNodeFileSystem } from "./vfs/createNodeFileSystem";
...
-    projectNames?: string[];
-}) {
+export class CliOptions {
+    public checkOnly: boolean = false;
+    public funcOnly: boolean = false;
+}
+
+type ConfigWithRootPath = Config & {
+    rootPath: string;
+};
```
vitorpy: Oh, that's so neat.

novusnota (dismissed):

CHANGELOG.md
```diff
@@ -49,6 +52,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Use the most recent version of the FunC standard library [`stdlib.fc`](https://github.com/ton-blockchain/ton/blob/4cfe1d1a96acf956e28e2bbc696a143489e23631/crypto/smartcont/stdlib.fc): PR [#283](https://github.com/tact-lang/tact/pull/283)
 - The WASM version of the FunC compiler has been updated to 0.4.4 and patched to work on larger contracts: PR [#297](https://github.com/tact-lang/tact/pull/297)
 - The `return`-statement reachability analysis: PR [#302](https://github.com/tact-lang/tact/pull/302)
+- The `-v`/`--version` flags now automatically source the version from `package.json`: PR [#287](https://github.com/tact-lang/tact/pull/287)
```
anton-trunov: This is not a user-faced fix, so it shouldn't be mentioned in the changelog
```suggestion
```

grammar/configSchema.json
```diff
@@ -52,6 +52,14 @@
...
               }
             }
+          },
+          "checkOnly": {
+            "type": "boolean",
+            "description": "False by default. If set to true, only performs syntax and type checking, preventing further compilation. Mutually exclusive with `funcOnly`."
+          },
+          "funcOnly": {
+            "type": "boolean",
+            "description": "False by default. If set to true, only outputs intermediate FunC code, preventing further compilation. Mutually exclusive with `checkOnly`."
```
anton-trunov: Since `checkOnly` an d`funcOnly` are mutually exclusive, these should be combined into an enum-like datatype: `"funcOnly" | "checkOnly" | "fullCompilation"`

anton-trunov (changes requested):
Great stuff!

anton-trunov:
@novusnota Something went wrong with single-contract compilation, which works as expected.

Here is a shell session with `tact.config.json`:
```shell
❯ ls
tact.config.json  test.tact

tact/contract new-cli*​ ≡
❯ cat tact.config.json 
{
  "$schema": "http://raw.githubusercontent.com/tact-lang/tact/main/grammar/configSchema.json",
  "projects": [
    {
      "name": "test",
      "path": "./test.tact",
      "output": "./",
      "mode": "funcOnly"
    }
  ]
}

tact/contract new-cli*​ ≡
❯ cat test.tact       
contract Foo {
    get fun foo(): Int {
        return 42;
    }
}

tact/contract new-cli*​ ≡
❯ ../bin/tact --config tact.config.json
💼 Compiling project test...
   > Foo: tact compiler
✔️ FunC code generation succeeded.

tact/contract new-cli*​ ≡
❯ ls
tact.config.json  test_Foo.abi      test_Foo.headers.fc  test_Foo.storage.fc
test.tact         test_Foo.code.fc  test_Foo.stdlib.fc
```

And here is a shell session with single-contract compilation with the same contract and and `bin/tact`:
```shell
❯ rm -f *.abi *.boc *.fc *.fif *.pkg *.md *.ts

tact/contract new-cli*​ ≡
❯ ls
tact.config.json  test.tact

tact/contract new-cli*​ ≡
❯ ../bin/tact --func test.tact                
💼 Compiling project main...
   > 👀 Enabling debug
   > Foo: tact compiler
   > Foo: func compiler
   > Foo: fift decompiler
   > Packaging
   > Foo
   > Bindings
   > Foo
   > Reports
   > Foo

tact/contract new-cli*​ ≡
❯ ls
main_Foo.abi       main_Foo.code.fif      main_Foo.md         main_Foo.storage.fc  test.tact
main_Foo.code.boc  main_Foo.code.rev.fif  main_Foo.pkg        main_Foo.ts          
main_Foo.code.fc   main_Foo.headers.fc    main_Foo.stdlib.fc  tact.config.json
```

It does not stop after generating FunC files.

.github/workflows/tact.yml
```diff
@@ -54,17 +57,32 @@ jobs:
...
+
       - name: Compare Tact version from CLI flag `--version` against package.json
         if: runner.os != 'Windows'
         run: |
-          if [ "$(./bin/tact --version)" != "$(jq -r '.version' < package.json)" ];
+          if [ "$(tact --version)" != "$(jq -r '.version' < package.json)" ];
           then false
           fi
 
+      - name: Check command line options
```
anton-trunov: Would be nice to have [cram](https://pypi.org/project/cram/)-like tests. And include there the regression described here: https://github.com/tact-lang/tact/pull/287#issuecomment-2088399034

novusnota:
@anton-trunov hmm, can't reproduce that error, my `--func` emits only .fc + .abi code. Have you pulled the latest commit?

UPD: I may have found a bigger source of potential bugs — bin/tact depends on src/node.ts, not on dist/node.js. Fixing this now.
UPD: Nevermind, didn't read the sources right.

anton-trunov:
@novusnota Yes, I'm on commit 0b18c8cbe0f3fd7a68bb3652a8ee15f8cd612e65. I can reproduce it after `yarn install && yarn clean && yarn gen && yarn build`

anton-trunov:
> UPD: I may have found a bigger source of potential bugs — bin/tact depends on src/node.ts, not on dist/node.js

To me this sounds backwards. Why do you need to reverse the dependency?

novusnota:
No changes there, just a "read issue" :)
Reproduced the bug, now locating why it happens, hold on a bit

UPD: Found out the reason (enum over disjoint values messed up the CLI schema a bit), now working on solution
UPD: Fixed the bug

.github/workflows/tact.yml
novusnota: Should I try to add such tests in this PR or it can be moved to a separate issue+PR?

P.S.: There's a maintained fork of cram, called prysk: https://github.com/prysk/prysk

vitorpy: We could do something similar to the [oclif testing](https://oclif.io/docs/testing), import an async method from `bin/tact` that incapsulates the command and add test coverage. 

I wouldn't mind doing it, but I'd rather do it in a separate PR.

anton-trunov: I'd prefer to stay within the JS/TS/Node.js ecosystem. On the other hand, oclif testing looks like too much JS. It should be based on snapshot, i.e. tests should look a lot like shell interaction histories. Btw, the cram-like framework we are going to pick should definitely support updating those snapshots.

anton-trunov: Yes, let's do that testing as a separate PR

anton-trunov:
```suggestion
      - name: CLI Test: Check single-contract compilation
```

.github/workflows/tact.yml
```diff
@@ -54,17 +57,32 @@ jobs:
         run: |
           type examples\output\echo_Echo.pkg
 
+      - name: Install the package locally
+        run: |
+          npm install -g ./
+
       - name: Compare Tact version from CLI flag `--version` against package.json
```
anton-trunov:
```suggestion
      - name: CLI Test: Compare Tact version from CLI flag `--version` against package.json
```

.github/workflows/tact.yml
```diff
@@ -54,17 +57,32 @@ jobs:
...
           then false
           fi
 
+      - name: Check command line options
+        run: |
+          tact --check test/tact-cli/success.tact
+          tact --func test/tact-cli/success.tact
+          tact test/tact-cli/success.tact
+
+      - name: Check command line arguments parsing
```
anton-trunov:
```suggestion
      - name: CLI Test: Check parsing of a non-existing CLI flag
```

.github/workflows/tact.yml
```diff
@@ -54,17 +57,32 @@ jobs:
...
+        if: runner.os != 'Windows'
+        run: |
+          ! tact --nonexistentoption test/tact-cli/success.config.json
+
       - name: Tact CLI non-zero exit code
```
anton-trunov:
```suggestion
      - name: CLI Test: tact executable return non-zero exit code if compilation fails
```

anton-trunov (changes requested):
Just one thing that is left: it's a de facto standard that more local flags should have priority over less local ones. Here how it fails for `bin/tact`:
```shell
❯ cat tact.config.json     
{
  "$schema": "http://raw.githubusercontent.com/tact-lang/tact/main/grammar/configSchema.json",
  "projects": [
    {
      "name": "test",
      "path": "./test.tact",
      "output": "./",
      "mode": "full"
    }
  ]
}

tact/contract new-cli*​ ≡
❯ ../bin/tact --config tact.config.json --check
💼 Compiling project test...
   > Foo: tact compiler
   > Foo: func compiler
   > Foo: fift decompiler
   > Packaging
   > Foo
   > Bindings
   > Foo
   > Reports
   > Foo
```

It is expected that the CLI flag `--check` has priority over `"mode": "full"` in `tact.config.json`.

novusnota:
> It is expected that the CLI flag `--check` has priority over `"mode": "full"` in `tact.config.json`.

Yup, will correct priorities

anton-trunov (approved):
Great effort @vitorpy, @novusnota! Tact is much more interactive now

anton-trunov:
btw, added a few more things to work on in terms of enhancing Tact's CLI into the tracking issue #136

would be pretty cool to have the `--expr` option to experiment with Tact interactively

novusnota:
> btw, added a few more things to work on in terms of enhancing Tact's CLI into the tracking issue #136

Nice! AST ones would drive their API counterparts, which is a big aid for tooling

> would be pretty cool to have the `--expr` option to experiment with Tact interactively

Not sure about this, as there's already Nujan and (upcoming once I sort out my doc-related and Sublime/Emacs-related debts) interactive `tact play` toolkit. Tact is a DSL, after all, so proper experiments would require some interaction with blockchain, at least on the Sandbox level

anton-trunov:
> proper experiments would require some interaction with blockchain, at least on the Sandbox level

this is why it's for expressions only :)

Tact has some unusual semantics for expressions: for instance, the integer division operation rounds towards negative infinity. That flag makes it easier to experiment when you do your stuff using your terminal. And with something like [fzf](https://github.com/junegunn/fzf) you have easily searchable interaction history available, even if you do not have a nice REPL ready yet.

novusnota:
Ok, I see the point now.

> even if you do not have a nice REPL ready yet.

Well, on that note, we may just make a basic REPL for Tact expressions a part of the CLI — by using the readline module of Node: https://nodejs.org/api/readline.html

As an idea, the actual expression for the `--expr/-e` may be omitted, in which case the expression REPL will be spawned.

anton-trunov:
A dedicated repl in addition to a cli option would be indeed nice! The option allows for better scriptability

PR #302, opened 2024-04-26
fix: `return`-statement reachability analysis

### LINKED ISSUE #268: Missing return statement not caught by Tact compiler
The Tact compiler fails to identify missing return statements during compilation. While the Tact compiler overlooks it, the FunC Compiler catches it but provides an unclear error message. This is not get caught by Tact VS Code extension.

**Expected Behavior:**
The Tact compiler should detect missing return statements and provide a clear and understandable error message during the compilation phase.

For quick testing, you can use [ide.nujan.io](https://ide.nujan.io/)

### Steps to Reproduce:
1. Use the following code snippet in a Tact compiler environment:
```solidity
import "@stdlib/deploy";

contract Counter with Deployable {
    val: Int as uint32;
    init(){
        self.val = 0;
    }

    receive("increment"){
        self.val = (self.val + 1);
    }

    get fun value(): Int {

    }
}
```
2. Compile the code.
3. Notice that the Tact compiler does not flag the missing return statement in the value function, while the FunC Compiler does but provides a vague error message which is hard to understand who are not aware of FunC.

anton-trunov:
Closes #268

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Tact's CLI returns a non-zero exit code if compilation fails: PR [#278](https://github.com/tact-lang/tact/pull/278)
 - Use the most recent version of the FunC standard library [`stdlib.fc`](https://github.com/ton-blockchain/ton/blob/4cfe1d1a96acf956e28e2bbc696a143489e23631/crypto/smartcont/stdlib.fc): PR [#283](https://github.com/tact-lang/tact/pull/283)
 - The WASM version of the FunC compiler has been updated to 0.4.4 and patched to work on larger contracts: PR [#297](https://github.com/tact-lang/tact/pull/297)
+- The `return`-statement reachability analysis: PR [#302](https://github.com/tact-lang/tact/pull/302)
 
 ## [1.2.0] - 2024-02-29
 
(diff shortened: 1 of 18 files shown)


novusnota (approved):
How could we fix the [`stmts-failed/case-41.tact`](https://github.com/tact-lang/tact/pull/302/files#diff-7f06012372007bbd1d4747982f446047cba6bc598f02a5f53cd89d84dfd471eb) in the future? Recursive descent of sorts?

Other than that, splendid work!

anton-trunov:
Thanks @novusnota!

> How could we fix the [stmts-failed/case-41.tact](https://github.com/tact-lang/tact/pull/302/files#diff-7f06012372007bbd1d4747982f446047cba6bc598f02a5f53cd89d84dfd471eb) in the future? Recursive descent of sorts?

It can be done for the simple case of Booleans, but not in general.

Consider the following example:
```
if belongs(x, aSetOfIntegers) {
    return 1;
}
if belongs(x, anotherSetOfIntegers) {
    return 2;
}
```

if `aSetOfIntegers \union anotherSetOfIntegers == Int` then we have covered all the paths, but it's too expensive to try and solve it (or even undecidable). Btw, Rust solves this issue in the same way essentially, but it has `if`-expressions so it's a typechecking error.

You can try using SMT solvers (or your own custom solver) to do this, but it's not worth it in practice.

byakuren-hijiri (approved):

PR #323, opened 2024-05-03
feat: prevent variable declaration from shadowing functions

### LINKED ISSUE #307: Function identifiers are allowed to be shadowed in their bodies
Both of the snippets below compile just fine, but should not, since variable shadowing is not allowed and functions in Tact are recursive by default:

```tact
fun rec(): Int {
    let rec: Int = 42;  // shadowing error
    return rec;
}
```

```tact
fun rec(): Int {
    let rec: Int = 42;  // shadowing error
    return rec();
}
```

vitorpy:
Closes #307

Just checking if this is a sensible approach to handle avoiding variable declarations from shadowing functions. 

I'm guessing I should also avoid shadowing other identifiers, such as contract names? Also non-static functions I guess.


src/types/resolveStatements.ts
```diff
@@ -180,6 +180,11 @@ function processStatements(
                 );
             }
 
+            const staticFunction = getAllStaticFunctions(ctx);
+            if (Object.values(staticFunction).some((f) => f.name === s.name)) {
+                throwError(`Variable shadows existing function: ${s.name}`, s.ref);
+            
+            }
```
anton-trunov: This looks a bit like an ad hoc approach to me. For instance, would it work for contract functions? This should be solved with a unified variable context if you ask me. So, a bit of refactoring is needed here

vitorpy:
@anton-trunov Sorry, can you provide some guidance here on which cases this is expected to cover? ie. which test cases are missing?

src/types/__snapshots__/resolveStatements.spec.ts.snap
```diff
@@ -510,6 +510,16 @@ Line 8, col 12:
 "
 `;
 
+exports[`resolveStatements should fail statements for case-51 1`] = `
+"<unknown>:4:5: Variable already defined : rec
```
anton-trunov: The error message is a bit confusing here: it should say "function 'rec' blah blah"

vitorpy:
@anton-trunov @novusnota Hey guys, should this be allowed?

image

novusnota:
> Hey guys, should this be allowed?

@vitorpy Tact doesn't have first-class functions, so we can allow such declarations, I think. Let's wait for input from @anton-trunov though.

anton-trunov:
> Tact doesn't have first-class functions, so we can allow such declarations

This is an instance of type-directed disambiguation and it's an opportunity to mess up. I say we go for simplicity and ban this, no shadowing should be allowed at this moment. I mean it's easier to ban some behavior and relax the restrictions in later language versions. And it's backwards compatible!

anton-trunov:
@vitorpy Let's disallow `let sender: Address = sender()`

PR #331, opened 2024-05-09
feat: new TVM instructions in stdlib

### LINKED ISSUE #324: Extend standard library with new TVM instructions
TVM Upgrades [2023.07](https://docs.ton.org/learn/tvm-instructions/tvm-upgrade-2023-07) and [2024.04](https://docs.ton.org/learn/tvm-instructions/fee-calculation-instructions) introduced many useful TVM instructions, including those for optimized work with fees.
We have to extend our standard library too.

Gusarich:
Closes #324 


stdlib/std/send.tact
```diff
@@ -18,7 +21,7 @@ struct SendParameters {
     data: Cell? = null;
 }
 
-fun send(params: SendParameters) {
+fun send(params: SendParameters): Int {
```
anton-trunov: Did the semantics of this function change?

stdlib/stdlib.fc
```diff
@@ -259,11 +259,9 @@ cont bless(slice s) impure asm "BLESS";
 () commit() impure asm "COMMIT";
 
 ;;; Not implemented
-;;() buy_gas(int gram) impure asm "BUYGAS";
-
 ;;; Computes the amount of gas that can be bought for `amount` nanoTONs,
 ;;; and sets `gl` accordingly in the same way as [set_gas_limit].
-() buy_gas(int amount) impure asm "BUYGAS";
+;;() buy_gas(int amount) impure asm "BUYGAS";
```
anton-trunov, line 264: Why is this commented out?

stdlib/stdlib.fc
```diff
@@ -441,15 +440,15 @@ builder store_ref(builder b, cell c) asm(c b) "STREF";
...
+;;; Stores (serializes) an integer [x] in the range `0..2^128 − 1` into `builder` [b].
 ;;; The serialization of [x] consists of a 4-bit unsigned big-endian integer `l`,
 ;;; which is the smallest integer `l ≥ 0`, such that `x < 2^8l`,
 ;;; followed by an `8l`-bit unsigned big-endian representation of [x].
 ;;; If [x] does not belong to the supported range, a range check exception is thrown.
 ;;;
 ;;; Store amounts of TonCoins to the builder as VarUInteger 16
 builder store_grams(builder b, int x) asm "STGRAMS";
-builder store_coins(builder b, int x) asm "STGRAMS";
+builder store_coins(builder b, int x) asm "STVARUINT16";
```
anton-trunov, line 451: Did they update this in the upstream FunC stdlib?

stdlib/stdlib.fc
```diff
@@ -626,14 +625,42 @@ int rand(int range) impure asm "RAND";
 ;;; Returns the current random seed as an unsigned 256-bit Integer.
 int get_seed() impure asm "RANDSEED";
 ;;; Sets the random seed to unsigned 256-bit seed.
-() set_seed(int) impure asm "SETRAND";
+() set_seed(int x) impure asm "SETRAND";
```
anton-trunov, line 628: @novusnota `set_seed(int)` is valid FunC? Does our FunC parser handle it well?

Gusarich:
Seems like https://github.com/tact-lang/tact-emulator doesn't support new opcodes yet: I'm getting `Invalid opcode` error in tests.

anton-trunov:
> Seems like https://github.com/tact-lang/tact-emulator doesn't support new opcodes yet: I'm getting Invalid opcode error in tests.

Looks like you are getting @tact-lang/emulator@4.2.3 instead of 4.3.0 in CI

> warning " > @tact-lang/emulator@4.2.3" has unmet peer dependency "ton-abi@^0.0.1".

You might need to update the `yarn.lock` file (and `package.json` too, although it uses `^` for the emulator package)

Gusarich:
Depends on https://github.com/tact-lang/tact-emulator/issues/13

Gusarich:
I guess since we won't use tact emulator anymore after the merge of #651, we can ignore that issue and finish this stdlib update.

anton-trunov:
That would be awesome!

anton-trunov (commented):
Some old comments of mine + a new one

stdlib/stdlib.fc
Gusarich: yes, it's valid

stdlib/stdlib.fc
Gusarich: Grams is a legacy naming for the Toncoin currency which is usually stored as VarUint16

stdlib/stdlib.fc
Gusarich: https://docs.ton.org/develop/func/stdlib#buy_gas

it's not currently implemented as stated in TON documentation

stdlib/std/send.tact
Gusarich: I thought it would be nice to use the new `SENDMSG` instruction for our `send()` function here so that it returns the fwd fee for this message

anton-trunov: This indeed looks nice!

anton-trunov: Btw, let's document the return value in a comment

stdlib/stdlib.fc
novusnota: @anton-trunov Yes, it's valid and our parser handles that — `stdlib.fc` is in the testsuite for a long time now :)

anton-trunov:
Looks like you also need to commit the changes to the generated `stdlib.ts`

stdlib/stdlib.fc
```diff
@@ -441,15 +440,15 @@ builder store_ref(builder b, cell c) asm(c b) "STREF";
 ;;; Stores `slice` [s] into `builder` [b]
 builder store_slice(builder b, slice s) asm "STSLICER";
 
-;;; Stores (serializes) an integer [x] in the range `0..2^120 − 1` into `builder` [b].
+;;; Stores (serializes) an integer [x] in the range `0..2^128 − 1` into `builder` [b].
```
anton-trunov: wait, `VarUInt16`'s max value is represented as `15` for the bits part and `(8*15)`-bit unsigned int for the second part

Gusarich: Yes, you're right. I took the stdlib from [here](https://github.com/ton-community/mintless-jetton/blob/main/contracts/stdlib.fc). Seems like a mistake from their side.

anton-trunov:
@Gusarich Looks like after #768 this can be implemented directly in Tact

stdlib/std/send.tact
```diff
@@ -18,7 +21,7 @@ struct SendParameters {
     data: Cell? = null;
 }
 
-fun send(params: SendParameters) {
+fun send(params: SendParameters): Int { // returns forward fee
```
anton-trunov, line 23: We should probably introduce some more constants for modes. Here is the `SENDMSG` instruction description:

> Creates an output action and returns a fee for creating a message. Mode has the same effect as in the case of SENDRAWMSG. Additionally +1024 means - do not create an action, only estimate fee. Other modes affect the fee calculation as follows: +64 substitutes the entire balance of the incoming message as an outcoming value (slightly inaccurate, gas expenses that cannot be estimated before the computation is completed are not taken into account), +128 substitutes the value of the entire balance of the contract before the start of the computation phase (slightly inaccurate, since gas expenses that cannot be estimated before the completion of the computation phase are not taken into account).

anton-trunov:
@Gusarich Any progress on this?

stdlib/std/send.tact
Gusarich: +64 and +128 here are the same as for regular `SENDRAWMSG`

Gusarich:
> @Gusarich Looks like after #768 this can be implemented directly in Tact

yes, but I think we can keep this PR as it's now and do that for future stdlib updates

anton-trunov:
> I think we can keep this PR as it's now and do that for future stdlib updates

Let's do a bit of doogfooding here: the more we use our own language features for the stdlib the less bugs those are going to have. And this is important because we are going to do a new release pretty soon

Gusarich:
> > I think we can keep this PR as it's now and do that for future stdlib updates
> 
> Let's do a bit of doogfooding here: the more we use our own language features for the stdlib the less bugs those are going to have. And this is important because we are going to do a new release pretty soon

then what about all other functions from stdlib? Many of them are used in `writeStdlib.ts` so we can't just rewrite them natively in Tact.

anton-trunov:
> then what about all other functions from stdlib? Many of them are used in writeStdlib.ts so we can't just rewrite them natively in Tact.

It's fine. If it cannot be done easily (at least now), let's not do it. When I ported some stdlib functions to use `asm`-blocks instead of a FunC binding, I wasn't trying to go overboard with that.

Gusarich:
@anton-trunov so should I move the new functions from `stdlib.fc` to Tact stdlib for this PR?

anton-trunov:
> @anton-trunov so should I move the new functions from stdlib.fc to Tact stdlib for this PR?

Yes, let's go for it

stdlib/std/contract.tact
```diff
@@ -19,3 +19,21 @@ native newAddress(chain: Int, hash: Int): Address;
...
+
+asm(cells bits workchain) fun getForwardFee(workchain: Int, bits: Int, cells: Int): Int { GETFORWARDFEE }
+
+asm(gas_used workchain) fun getSimpleComputeFee(workchain: Int, gas_used: Int): Int { GETGASFEESIMPLE }
+
+asm(cells bits workchain) fun getSimpleForwardFee(workchain: Int, bits: Int, cells: Int): Int { GETFORWARDFEESIMPLE }
+
+asm(fwd_fee workchain) fun getOriginalFwdFee(workchain: Int, fwd_fee: Int): Int { GETORIGINALFWDFEE }
+
+asm fun myStorageDue(): Int { DUEPAYMENT }
```
anton-trunov: 1. Let's add comments explaining what all these functions do.
2. Do we want to do arguments shuffling? This increases gas usage. And, for instance, HipoFinance don't do this shuffling, so porting their contracts to Tact would be slightly harder and error-prone.

stdlib/stdlib.fc
```diff
@@ -325,9 +323,9 @@ cell preload_ref(slice s) asm "PLDREF";
 ;;; Preloads the first `0 ≤ len ≤ 1023` bits from slice [s] into a separate `slice s''`.
 ;; slice preload_bits(slice s, int len) asm "PLDSLICEX";
 
-;;; Loads serialized amount of TonCoins (any unsigned integer up to `2^120 - 1`).
-(slice, int) load_grams(slice s) asm( -> 1 0) "LDGRAMS";
-(slice, int) load_coins(slice s) asm( -> 1 0) "LDGRAMS";
+;;; Loads serialized amount of TonCoins (any unsigned integer up to `2^128 - 1`).
```
anton-trunov:
```suggestion
;;; Loads serialized amount of TonCoins (any unsigned integer up to `2^120 - 1`).
```

anton-trunov (changes requested):

anton-trunov:
LGTM. Just the specllchecker complaints. Btw, we do have Knip now with `pre-push` lints running automatically and those include spellchecking

Gusarich:
> LGTM. Just the specllchecker complaints. Btw, we do have Knip now with `pre-push` lints running automatically and those include spellchecking

fixed

anton-trunov (approved):

Issue #298, opened 2024-04-25
Simple destructuring of fields of structs and messages


novusnota:
Instead of:

```tact
message SomeMsg {
    iField: Int;
    sField: String;
}

receive(msg: SomeMsg) {
    let first: Int = msg.iField;
    let second: Int = msg.sField;
}
```

we could have the following:

```tact
message SomeMsg {
    iField: Int;
    sField: String;
}

receive(msg: SomeMsg) {
    let { first, second } = msg; // when we'll have a local type inference from #161
    // and
    let { onlyFirst, _ } = msg; // discarding the unwanted fields
    
    // Trailing commas should be allowed as well:
    let { firstAgain, secondAgain, } = msg;
    
    // The inverse of such destructuring operation is already present in the language:
    SomeMsg{ iField: first, sField: second };
}
```

Note, that this won't require any advanced pattern matching capabilities and should be rather simple to implement in the compiler. Marked this as a discussion because this is just a suggestion. This may or may not drive further improvements of working with Structs and Messages to make it much more convenient.

Depends on #161.

PR #338, opened 2024-05-15
Allow underscores as unused variable identifiers

### LINKED ISSUE #319: Allow underscores as unused variable identifiers
Code like this should compile:
```tact
let size: Int = 0;
foreach (_, _ in map) {
    size += 1;
}
```

Gusarich:
Closes #319 


anton-trunov (changes requested):
Let's fix the merge conflicts here

novusnota:
As of now, it only works for `foreach` and `catch`, right?

Gusarich:
> As of now, it only works for `foreach` and `catch`, right?

Are there any other cases where we need it?

novusnota:
> Are there any other cases where we need it?

I don't think so, at least now. When we'll have #298 in the future, identifiers in `let` declarations would need them though.

anton-trunov:
> When we'll have https://github.com/tact-lang/tact/issues/298 in the future, identifiers in let declarations would need them though.

Actually, `let`-statements should have underscores even now. This can be used as a clear indication that some function gets called not for its return value but for its side-effects:

```
let _ : Foo = effectfulFun(x, y, z);
```

When we have a bit of type inference, the snippet above will become

```
let _ = effectfulFun(x, y, z);
```

In this case a static analyzer won't complain about unused values, whereas it should issue a warning for 

```
effectfulFun(x, y, z);
```

if `effectfulFun` is a non-void function.

Gusarich:
Added support for let-statements and also resolved all merge conflicts

Gusarich:
resolved merge conflicts

Gusarich:
@anton-trunov resolved merge conflicts yet again after merge of #198

src/types/stmts/var-underscore-name-in-let.tact
```diff
@@ -0,0 +1,10 @@
+primitive Int;
+
+fun someImpureFunction(): Int {
+    return 123;
+}
+
+fun test(): Int {
+    let _: Int = someImpureFunction();
```
anton-trunov, line 8: Let's also add a test with `let` and no type annotation:
```
    let _ = someImpureFunction();
```

src/types/__snapshots__/resolveStatements.spec.ts.snap
```diff
@@ -580,6 +580,36 @@ Line 7, col 9:
...
+Line 6, col 16:
+  5 |     foreach (_, _ in m) {
+> 6 |         return _;
+                     ^
+  7 |     }
+"
+`;
+
+exports[`resolveStatements should fail statements for var-underscore-name-access2 1`] = `
+"<unknown>:7:14: Unable to resolve id _
```
anton-trunov: I would expect this error message to be

> Variable name cannot be '_'

or, if we converge on the proposed change below:

> Wildcard variable name '_' cannot be used

src/types/resolveStatements.ts
```diff
@@ -64,6 +64,9 @@ function addVariable(
     if (src.vars.has(name)) {
         throw Error("Variable already exists: " + name); // Should happen earlier
     }
+    if (name == "_") {
+        throw Error("Variable name cannot be '_'");
```
anton-trunov:
```suggestion
        throw Error("Wildcard variable name '_' cannot be used");
```

src/test/e2e-emulated/contracts/underscore-variable.tact
```diff
@@ -0,0 +1,55 @@
...
+        let x: Int = 0;
+        foreach (k, _ in m) {
+            x += k;
+        }
+        return x;
+    }
+
+    get fun test4(): Int {
+        let _: Int = self.increaseSomething();
+        let _: Int = self.increaseSomething();
```
anton-trunov, line 52:
```suggestion
        let _ = self.increaseSomething();
        let _ = self.increaseSomething();
```

src/types/resolveStatements.ts
```diff
@@ -185,12 +188,16 @@ function processStatements(
                         s.ref,
                     );
                 }
-                sctx = addVariable(s.name, variableType, sctx);
+                if (s.name !== "_") {
+                    sctx = addVariable(s.name, variableType, sctx);
```
anton-trunov: we should probably move the check `s.name !== "_"` inside `addVariable` itself -- it would just ignore it

src/types/resolveStatements.ts
```diff
@@ -388,15 +395,22 @@ function processStatements(
...
             // Process catchName variable for exit code
-            if (initialCtx.vars.has(s.catchName)) {
-                throwError(`Variable already exists: "${s.catchName}"`, s.ref);
+            if (s.catchName != "_") {
+                if (initialCtx.vars.has(s.catchName)) {
+                    throwError(
+                        `Variable already exists: "${s.catchName}"`,
+                        s.ref,
+                    );
+                }
```
anton-trunov: Let's refactor this code into a helper function and reuse it in this file in numerous places.

anton-trunov (changes requested):
Almost there!

Gusarich:
@anton-trunov have a look

anton-trunov (approved):

PR #343, opened 2024-05-17
Add native support of exotic cells

### LINKED ISSUE #154: Exotic cells
Exotic cells are very useful in certain circumstances. For example, they allow efficient interactions with merkle trees and masterchain libraries usage.
But at the moment, there's no easy way to utilize exotic cells in Tact (actually it's not a trivial task in FunC too). It would be great to have a simple interface for working with them, which will handle low-level things (like parsing them automatically and checkin some fields).

It's not obvious how exactly should that interface look like, so it needs a discussion.

I imagine it to look somehow like this (example with merkle proofs. structures from https://github.com/Gusarich/airdrop):
```tact
struct AirdropEntry {
    address: Address;
    amount: Int as coins;
}

struct ProvableData {
    entries: map<Int, AirdropEntry>;
}

message ProcessClaim {
    queryId: Int as uint64;
    proof: MerkleProof<ProvableData>;
    index: Int as uint256;
}

// somewhere in the code later

receive(msg: ProcessClaim) {
    require(msg.proof.root == self.root, "wrong merkle root");
    entry: AirdropEntry = msg.proof.data.get(msg.index);
    // That's it!

    // do something else with data
}
```

However, there is an issue with the way structs are handled in Tact at the moment that prevents us from implementing this feature correctly. Currently, structs are being parsed completely right after their loading. It's fine for regular structs, but in the case of merkle proofs and merkle updates, some fields of the struct might be pruned and therefore impossible to parse. It'll work fine for small structs (like the one from the example above), but if we want to add some more fields to it so that it splits into multiple cells, and then we provide that struct with some of these cells pruned to the contract, it'll throw a runtime error because it won't be able to load pruned cells (even though we won't use them there).

We probably need to introduce a lazy struct parsing mechanism into Tact so that all the fields are parsed the moment they're accessed for the first time. Another approach is to analyze struct field accesses in each receiver to only parse the fields that are used and skip everything else.  
But I'm not sure exactly how it should be implemented and what we should do with `Struct.fromCell()`. Maybe we should only use this "lazy loading" for merkle proof and merkle update cells?

Gusarich:
Closes #154 

[WORK IN PROGRESS]


anton-trunov:
I saw in TON dev chats interest from Tact users regarding the exotic cells support -- they would like to implement cNFTs in Tact.

anton-trunov:
@Gusarich Any progress on this?

Gusarich:
> @Gusarich Any progress on this?

I'm rewriting this thing because at the current state it's easier than trying to resolve all the conflicts with main branch

stdlib/stdlib.fc
```diff
@@ -637,3 +637,6 @@ int equal_slice_bits (slice a, slice b) asm "SDEQ";
 
 ;;; Concatenates two builders
 builder store_builder(builder to, builder from) asm "STBR";
+
+(slice, int) begin_parse_exotic(cell c) asm "XCTOS";
+cell end_cell_exotic(slice s) asm "ENDXC";
```
anton-trunov: We should probably also(?) add these as Tact stdlib asm-functions

src/test/features/merkle-trees.tact
anton-trunov, line 1: This looks just awesome!

Gusarich:
rewrote the whole thing, it *kind of* works (compiles at least), but it's incomplete and doesn't work properly because I didn't finish yet. I think we should move this to 1.6.0 due to a deadline

anton-trunov:
Let's close this for now: we have an issue to track this feature, the code is out of date with the code base and it's hard to manage PRs when there is a lot of stale ones.

PR #200, opened 2024-03-27
Display an error for integer overflow at compile-time

byakuren-hijiri:
I'm not sure how and if we should add a test for fallible compilation.

Closes #189

PR #352, opened 2024-05-23
Fix const eval

### LINKED ISSUE #327: Division by zero in constant evaluation is not detected in Tact and leaks to FunC
This contract
```
contract DivisionByZero {

    get fun foo(): Int {
        let x: Int = 1 / 0;
        return 42;
    }
}
```

fails with the following error:

```shell
Func compilation error: cannot generate code for function `$DivisionByZero$_fun_foo`:
/dist/tact_DivisionByZero.code.fc:20:17: error: division by zero
int $x = (1 / 0); 
``

### LINKED ISSUE #339: Minimal integer `-(2**256)` makes FunC compilation fail
The following snippet 
```tact
contract Test {

    get fun minInt(): Int {
        // -(2**256)
        return -115792089237316195423570985008687907853269984665640564039457584007913129639936;
    }
}
```
produces the following FunC compilation error:

```
Func compilation error: /dist/tact_Test.code.fc:20:23: error: invalid integer constant `115792089237316195423570985008687907853269984665640564039457584007913129639936`
return ($self, (- 115792089237316195423570985008687907853269984665640564039457584007913129639936))
```

Reported by @novusnota.

### LINKED ISSUE #345: Constant evaluation is not used widely during compilation
Turns out that the reason behind both #327 and #339 is that constant evaluation in compiler works not with all values actually. It's only being called in certain cases, such as static calls of globally defined functions.
We should change this behavior.

Gusarich:
Closes #345 and as result resolves #339 and resolves #327 


Gusarich:
I'm not sure if the `.include(...)` thing is acceptable for such error messages checks, but it works.

src/test/features/const-eval.tact
anton-trunov, line 1: This test should be moved elsewhere. The feature tests are supposed to be dynamic tests and the contracts here are supposed to compile. For failing const-eval tests we can create a separate category, similar to the failing typechecker tests

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `log2` and `log` math functions were adjusted for consistency in error throwing: PR [#342](https://github.com/tact-lang/tact/pull/342)
 - Shadowing built-in static functions is now forbidden: PR [#351](https://github.com/tact-lang/tact/pull/351)
+- All integer and boolean expressions are now being attempted to be evaluated as constants. Additionally, compile-time errors are thrown for errors encountered during the evaluation of actual constants: PR [#352](https://github.com/tact-lang/tact/pull/352)
```
anton-trunov: Do we const-evaluate things like `const x: Int = 42 ? 1 : 0`? If this is not the case, we should say something like `More ...` instead of `All ...`

src/types/resolveConstantValue.ts
```diff
@@ -71,13 +71,16 @@ function reduceInt(ast: ASTExpression): bigint {
...
         if (error instanceof RangeError) {
-            throwError(
-                "Cannot evaluate constant expression due to integer overflow",
-                ast.ref,
-            );
-        } else {
-            throw error;
+            if (error.message === "Division by zero") {
+                throwError("Cannot divide by zero", ast.ref);
+            } else if (error.message === "Maximum BigInt size exceeded") {
```
anton-trunov, line 76: What about 257-bit integers? Or integers with storage modifiers like `Int as uint32`?

anton-trunov (changes requested):

src/types/resolveConstantValue.ts
Gusarich: 257-bit integers should work fine, also I didn't add the `Maximum BigInt size exceeded`, it was added before in #200, so why would it stop working correctly?

src/test/features/const-eval.tact
Gusarich: same applies to "implicit-init" negative compilation test too?

anton-trunov: sure

src/types/resolveConstantValue.ts
anton-trunov: No, I meant, BigInts in TypeScript are not 257-bit integers, this is why "Maximum BigInt size exceeded" looks a bit weird here

Gusarich: But it's the error that TypeScript throws, and when we catch it in Tact, an "Cannot evaluate constant expression due to integer overflow" compilation error is thrown instead.

CHANGELOG.md
Gusarich: such expressions are being attempted to be const-evaluated, but due to incompleteness of the const-evaluation itself it doesn't succeed.

anton-trunov:
Looks like there is a merge conflict

Gusarich:
Looks like I've messed up the PR a little bit because of the automatically suggested merging on Github ☹️

anton-trunov:
@Gusarich Any progress here?

Gusarich:
@anton-trunov reverted incorrect merge commit and rebased this PR correctly.
Sorry for the delay!

anton-trunov (approved):

PR #365, opened 2024-05-27
refactor(grammar): simplify grammatical categories

### LINKED ISSUE #332: Incorrect error message for non-initialized contract constants
```
contract SampleTactContract with Deployable {
    const u4: Int;
}
```

Error message:
```
Tact compilation failed
contract.tact:8:5: Abstract function doesn't have abstract modifier
Line 8, col 5:
  7 | contract SampleTactContract with Deployable {
> 8 |     const u4: Int;
          ^~~~~~~~~~~~~~
  9 |   }
```

Expected: Something [like this](https://github.com/tact-lang/tact/blob/79e87d5b2170a19aa44a34325c183f9ffe244155/src/types/resolveExpression.ts#L381), e.g. `Contract constant is not initialized`.

---

When adding such a constant in the `ASTContract` entry directly, compilation fails with an exception [here](https://github.com/tact-lang/tact/blob/79e87d5b2170a19aa44a34325c183f9ffe244155/src/generator/writers/writeExpression.ts#L117), because `s` is `undefined`.

### LINKED ISSUE #341: Simplify grammar.ohm and grammar.ts
This change is supposed to clean up and deduplicate the grammar and AST building process, but it is not supposed to make any changes to the produced ASTs.

anton-trunov:
and introduce some renamings to make it sound more standard

Closes #341

Incidentally, it should fix #332, because we forbid constant and function declarations everywhere except for traits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Refactor AST types to simplify access to third-party tools: PR [#325](https://github.com/tact-lang/tact/pull/325)
 - Refactor the compiler API used to access AST store: PR [#326](https://github.com/tact-lang/tact/pull/326)
 - Update JSON Schema to inform about usage in Blueprint: PR [#330](https://github.com/tact-lang/tact/pull/330)
+- The Tact grammar has been refactored for better readability: PR [#365](https://github.com/tact-lang/tact/pull/365)
 
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


novusnota (approved):
Nice changes in structure and naming. Especially the `Module = Import* ModuleItem*`, which makes imports allowed only at the top-level right in the grammar. Cool!

I'll port some names and structure over to [tree-sitter-tact](https://github.com/tact-lang/tree-sitter-tact) later this week :)

And while we're at it, you may consider adding [rule descriptions](https://ohmjs.org/docs/syntax-reference#rule-descriptions) ([in .ohm](https://github.com/ohmjs/ohm/blob/main/packages/ohm-js/src/ohm-grammar.ohm#L64-L65)) to some of the rules, because they directly affect the syntax error messages produced by Ohm. Although adding them may complicate things a little, see: https://github.com/ohmjs/ohm/issues/302

anton-trunov:
> Especially the Module = Import* ModuleItem*, which makes imports allowed only at the top-level right in the grammar. 

The change just made this restriction explicit, because before this refactoring it was later done in the `parseImports` function, which can now be simplified. In a subsequent refactoring we'll most likely make imports a separate field in the module AST (this is out-of-scope for this PR)

> And while we're at it, you may consider adding [rule descriptions](https://ohmjs.org/docs/syntax-reference#rule-descriptions) ([in .ohm](https://github.com/ohmjs/ohm/blob/main/packages/ohm-js/src/ohm-grammar.ohm#L64-L65)) to some of the rules, because they directly affect the syntax error messages produced by Ohm. Although adding them may complicate things a little, see: https://github.com/ohmjs/ohm/issues/302

Let's try to avoid "while-at-it" changes in the development process and create an issue that explains why this needs to be done and resolve it via a follow-up PR. This is especially important for refactoring PRs that need to be merged asap to not mess with other PRs that add new Tact features

Issue #314, opened 2024-05-01
Tact as library


byakuren-hijiri:
Third-party TypeScript API

novusnota:
> Save types to AST or improve an access API to type information.

This would be insanely useful in tools like language server and such!

In general, I agree with all those points, but implementation of some may just require us to make our own lexer→parser→semantic analysis pipeline, suitable for compiler AND external tooling depending on various steps in it. I'd say this depends on #286, but I may be wrong and we could pull off such feat without compromising the Ohm's toolkit.

byakuren-hijiri:
I would also suggest creating named union types for AST entries used in fields. For example:
```typescript
export type ASTContract = {
    kind: 'def_contract';
    origin: TypeOrigin;
    id: number;
    name: string;
    traits: ASTString[];
    attributes: ASTContractAttribute[];
    declarations: (ASTField | ASTFunction | ASTInitFunction | ASTReceive | ASTConstant)[];
    ref: ASTRef;
};
```

Could be rewritten as:
```typescript
export type ASTContractDeclaration = (ASTField | ASTFunction | ASTInitFunction | ASTReceive | ASTConstant);
export type ASTContract = {
    kind: 'def_contract';
    origin: TypeOrigin;
    id: number;
    name: string;
    traits: ASTString[];
    attributes: ASTContractAttribute[];
    declarations: ASTContractDeclaration[];
    ref: ASTRef;
};
```

This will simplify the life of tooling developers by enabling them to reuse these type definitions from the compiler. Otherwise, I find myself copy-pasting these entries in my projects. Here is an example of a function with such a signature implemented in the static analyzer internals:
```typescript
function getMethodInfo(
    decl: ASTField | ASTFunction | ASTInitFunction | ASTReceive | ASTConstant,
  ): [string | undefined, FunctionKind | undefined] {
```

byakuren-hijiri:
Added three more points:

> Add more context to every internal Error to be thrown in compiler internals. This is crucial for debugging third-party tools.
> Add AST iterators that perform functional map and fold over nested nodes of the AST. This will enable API users to inspect the AST in a more convenient way, for example, sorting nodes of a certain type.
> Add an API that provides equivalence checks between AST nodes of the same type. This is needed, for example, in https://github.com/tact-lang/tact/pull/335 to implement unit tests.

byakuren-hijiri:
Added while working on tests for https://github.com/tact-lang/tact/pull/559:

[...]

PR #368, opened 2024-05-30
feat(api): Introduce `fold`/`map` functions for AST traversal

byakuren-hijiri:
This improves compiler's API providing an interface to traverse the AST. It should be a part of the compiler, since it is be useful for devtools using the API and might be useful internally.

See: #314

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Tests for recursive functions: PR [#359](https://github.com/tact-lang/tact/pull/359)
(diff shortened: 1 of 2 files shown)


byakuren-hijiri:
Here are some examples of usage of these functions:
* https://github.com/nowarp/misti/blob/66f4254b561d1d4f3058974c06f7f4b5a4215f09/src/detectors/builtin/zeroAddress.ts#L38
* https://github.com/nowarp/misti/blob/66f4254b561d1d4f3058974c06f7f4b5a4215f09/src/detectors/builtin/readOnlyVariables.ts#L117

They could be used to traverse the AST by calling callback functions or accumulating information about the AST, so they are quite useful for all the tools working with Tact AST. Implementing this API in the compiler will make it easier to maintain the API upon AST changes.

anton-trunov (approved):
Awesome!

PR #380, opened 2024-06-06
Fix constructor generation for empty messages

### LINKED ISSUE #74: Sending empty message throws at FunC compilation
Hey,

if I try to send an empty message the FunC compiler throws a cryptic error message:

```
Func compilation error:...:184:14: error: previous function return type tuple cannot be unified with return statement expression type (): cannot unify type () with tuple
```

Example:
```
message MyMessage {
}

contract MyContract {
    receive("send") {
        send(SendParameters {
            to: context().sender, 
            value: 0, 
            bounce: false,
            mode: SendIgnoreErrors,
            body: MyMessage {}.toCell()
        });
    }
}
```

After investigating the Tact compiler output it was clear that the error was caused by the empty message.

Gusarich:
Closes #74 


Gusarich:
Another solution would be to forbid defining messages without fields, like it's done with structs. But I don't think this is a good idea for messages.

anton-trunov (approved):

PR #384, opened 2024-06-07
Fix chaining for mutable functions

### LINKED ISSUE #118: Tact produces FunC code with errors
Some errors, such as some type errors, are not caught by the Tact-FunC transpiler, but the FunC compiler would report errors on the FunC code generated by Tact.

This is not acceptable, and Tact should always produce error-free FunC code.

### LINKED ISSUE #211: Tact produces invalid FunC code for modifying load function
This line:
```tact
let x: Int = beginCell().storeUint(0, 1).endCell().beginParse().loadUint(1);
```

Produces this FunC code:
```func
int $x = begin_parse(end_cell(store_uint(begin_cell(), 0, 1)))~load_uint(1);
```

Which results error:
```
error: lvalue expected before `~load_uint`
```

### LINKED ISSUE #348: Application of mutation functions on function returns leaks to FunC and errors there
The following snippet:

```tact
contract Test {
    get fun test() {
        emptySlice().loadRef();  // ← this is the cause
    }
}
```

Produces this FunC compilation error:

```shell
Func compilation error: /test/tact_Test.code.fc:23:25: error: lvalue expected before `~load_ref`
     $global_emptySlice()~loadRef();
```

The error is not related to empty Slice having no references, but rather is about not being able to modify the value from the function return on the fly. The latter is justified and perfectly fine, but we should probably catch similar cases in Tact before they leak to FunC and cause errors there.

Gusarich:
Closes #211, Closes #118, Closes #348


anton-trunov (approved):
Oh, this is clever! Wrapping modifying functions to provide a temporary binding for the argument being modified.

Issue #229, opened 2024-04-03
Support constants initialization dependency


byakuren-hijiri:
Consider the following example:
```
const _0: Int = 1;
const _1: Int = _0 + 1;
```

Compilation output:
```
Tact compilation failed
Error: contract.tact:4:16: Cannot reduce expression to a constant integer
Line 4, col 16:
  3 | const _0: Int = 1;
> 4 | const _1: Int = _0 + 1;
```

Expected: successful compilation as the value of `_0` is known in compile time.

Related to #195.

The fix should support the following case:

```
const x: Int? = 42;
const f: Bool = x == null; // evaluates to false
```

anton-trunov:
Resolved in #664

Issue #400, opened 2024-06-11
Storage vars cannot be initialized by struct instances


byakuren-hijiri:
The following snippet cannot be compiled:

```
struct A {
    a: Bool;
}
contract Test {
    s: A = A {a: false};
}
```

```
"<unknown>:13:14: Expected constant value, got A
```

Note that if all the above initializing expressions move to `init()` and use assignment, everything works.

Related: https://github.com/tact-lang/tact/issues/284

Issue #284, opened 2024-04-22
Constant structs are not supported


anton-trunov:
The following snippet

```tact
struct A {
    x: Int;
    y: Int;
}

const a: Int = 5;
const b: Int = 6;

const D: A = A { x: a, b };
```

produces an error message like this

```
"<unknown>:13:14: Expected constant value, got A
Line 13, col 14:
  12 | 
> 13 | const D: A = A { x: a, b };
                    ^~~~~~~~~~~~~
```

while it is expected the error would point to the `b` in the struct definition (provided the field punning feature is available).

vitorpy:
What would be the correct approach here to support const structs, something along the lines of a `reduceStruct` implementation that would reduce the struct to a const BoC?

byakuren-hijiri:
I would also mention that fields cannot be initialized by struct instances. For example, the following snippet returns the same compiler error:

```
struct LVc {
    UxS4: String;
    KCNI_GZzVYYQ_: Bool;
}
contract __Vb {
    _vINF_u_NA_: LVc = LVc{UxS4: "", KCNI_GZzVYYQ_: false};
}
```

anton-trunov:
@byakuren-hijiri That one looks cool! Let's create a separate issue for that and tag it "found-by-fuzzer"

PR #401, opened 2024-06-11
Implement constant string receivers

### LINKED ISSUE #51: Use of const String in receive parameter
The following code should work:
```
const ConstMsg: String = "ABC";

receive(ConstMsg) {}
```
Currently it report error `Syntax error: expected ":" `

Gusarich:
Closes #51 


Gusarich:
just found out it doesn't work correctly for contract-scoped constants. I'll fix that

Gusarich:
Fixed.
However, there's a temporary limitation on lvalue paths. I had to add it because #284 and #400 aren't resolved yet.
We'll be able to support strings from constant structs after these are resolved.

Gusarich:
Rebased, added more tests and implemented handling constant values from structs.
Also found #439 during testing.

src/types/test-failed/contract-duplicate-const-string-receiver3.tact
anton-trunov, line 1: This is a case of "receiver not found", right? Not a duplicate one, because string3 is missing

src/types/resolveDescriptors.ts
```diff
@@ -1006,6 +1027,95 @@ export function resolveDescriptors(ctx: CompilerContext) {
...
                             ast: d,
                         });
+                    } else if (
+                        d.selector.kind === "internal-const-comment" ||
+                        d.selector.kind === "external-const-comment"
+                    ) {
+                        const internal =
+                            d.selector.kind === "internal-const-comment";
+
+                        let path = d.selector.comment;
```
anton-trunov, line 1037: It feels like the following constant look-up code should be factored out into a helper function. This also should be implemented for expressions as well, can't we reuse that code here?

anton-trunov (changes requested):
Great progress!

Let's add some more tests:
- a positive test: a constant definition in a trait that is used as a receiver name;
- a negative test with a trait constant _declaration_, so it does not have a body.

src/types/resolveDescriptors.ts
anton-trunov: Btw, the constant lookup you have implemented here, should go to the constant evaluator: https://github.com/tact-lang/tact/blob/9ed9193e81dc3ed240a1f6d2530659d2705b7ba2/src/constEval.ts#L521. So, after typechecking the receiver constant string, you just run const-eval on it and you should get your string. This way you will also extend the supported language fragment for const-eval and resolve issue #229.

anton-trunov: We could even allow arbitrary expressions in receivers as long as those are string-valued constant expressions

src/grammar/grammar.ohm
```diff
@@ -85,9 +85,11 @@ Tact {
 
     Receiver = receive "(" Parameter? ")" "{" Statement* "}" --regular
              | receive "(" stringLiteral ")" "{" Statement* "}" --comment
+             | receive "(" LValue ")" "{" Statement* "}" --constComment
```
anton-trunov, line 88: Let's actually change `stringLiteral` to `Expression` here (and for externals) to support arbitrary expressions: it should work as long as it's a string-valued constant expression. In the future we might have some compile-time string builtins or even allow integers, so instead of doing this:
```
message(1) OpCodeOneNoData {}

contract Contract {
  // would be nice to support underscores here too, since OpCodeOneNoData does not carry any further info
  receive(f: OpCodeOneNoData) {
      self.reply("OpCodeOneNoData".asComment())
  }
}
```

one could do
```
contract Contract {
  receive(1) {
      self.reply("OpCodeOneNoData".asComment())
  }
}
```

novusnota: Hmm, interestingly enough it would somewhat change receivers towards #433. I'd propose having arbitrary string-valued expressions specified in constant declarations somewhere on top, or generally just prior to receivers and not embedded in. Motivation here reflects what you've said there: https://github.com/tact-lang/tact/issues/433#issuecomment-2177804185

That said, @Gusarich would have the last word here :)

anton-trunov: yeah, it's a bit too late to have a consistent design for Tact v1.x :) given that we already have messages and strings

src/types/resolveDescriptors.ts
Gusarich: This way we'll try to evaluate a constant that might depend on the resolved contract during resolving that same contract. With static constants it'll work fine, but with cases like `receive(self.something)` it won't work...

anton-trunov: Why not? We can first process all constants and then process receivers

Gusarich: Because these are stored in CompilerContext with the contract itself in the very end of the resolving:
https://github.com/tact-lang/tact/blob/9ed9193e81dc3ed240a1f6d2530659d2705b7ba2/src/types/resolveDescriptors.ts#L1622-L1624

So I can't get contract-scoped constants by calling `getAllTypes` or `getType` in constant evaluator.

anton-trunov: if we cannot move it before receiver processing stage, let's just push this PR to v1.5 and solve this properly

anton-trunov:
Let's come back to this feature a bit later

PR #351, opened 2024-05-23
Forbid shadowing static functions from `global.ts`

Gusarich:
Closes #306 

- [x] I have updated CHANGELOG.md
- [x] I have added tests to demonstrate the contribution is correctly implemented: this usually includes both positive and negative tests, showing the happy path(s) and featuring intentionally broken cases
- [x] I have run all the tests locally and no test failure was reported
- [x] I did not do unrelated and/or undiscussed refactorings

PR #418, opened 2024-06-14
Implement `fromCell` and `fromSlice` methods for structs

### LINKED ISSUE #344: Consider adding more higher-level approaches to build and parse Cells, which would be compile-time friendly
### Motivation

Currently, one has to use the Builder pattern to construct `Cell`s starting with `beginCell()` and then `storeSmth()` up as needed. Then, for parsing some obtained `Cell` one has to convert it to `Slice` and use `loadSmth()` functions until it's fully read. It is alright for FunC, but we can definitely do better.

### Existing stuff

As the simple construction of messages goes, we already have `SendParameters` Struct alongside with the `send()` function, which are quite easy to reason about — it should be possible to come up with something like that, but now for parsing messages. And for parsing `raw` field of the `Context` Struct too!

### Gas optimization

As a bonus, this may lead to new compile-time optimizations (read: off-chain computation regarding `Cell`s), which are a massive win in terms of gas usage — `beginCell()` on its own costs around 500 gas to begin with, despite not doing much. If we could pre-compute as much of the `Cell` construction as possible, we may achieve massive gas benefits.

Potentially, with new ways to parse common TL-B layouts of `Cell`s (or even introducing neat mappings to generalize work with any TL-B layouts?), `Cell` parsing can also get optimized away too. But that's more of a suggestion, the real thing to focus in message parsing/composing.

Gusarich:
Closes #344 

### DIFF
--- a/src/abi/struct.ts
+++ b/src/abi/struct.ts
@@ -42,4 +42,96 @@ export const StructFunctions: Map<string, AbiFunction> = new Map([
             },
         },
     ],
+    [
+        "fromCell",
+        {
+            name: "fromCell",
+            resolve: (ctx, args, ref) => {
+                if (args.length !== 2) {
+                    throwError("fromCell() expects one argument", ref);
+                }
+                if (args[0].kind !== "ref") {
+                    throwError(
+                        "fromCell() is implemented only a struct type",
+                        ref,
+                    );
+                }
+                const tp = getType(ctx, args[0].name);
+                if (tp.kind !== "struct") {
+                    throwError(
+                        "fromCell() is implemented only a struct type",
+                        ref,
+                    );
+                }
+                if (args[1].kind !== "ref" || args[1].name !== "Cell") {
+                    throwError("fromCell() expects a Cell as an argument", ref);
+                }
+                return { kind: "ref", name: args[0].name, optional: false };
+            },
+            generate: (ctx, args, resolved, ref) => {
+                if (resolved.length !== 2) {
+                    throwError("fromCell() expects one argument", ref);
+                }
+                if (args[0].kind !== "ref") {
+                    throwError(
+                        "fromCell() is implemented only a struct type",
+                        ref,
+                    );
+                }
+                if (args[1].kind !== "ref" || args[1].name !== "Cell") {
+                    throwError("fromCell() expects a Cell as an argument", ref);
+                }
+                return `${ops.readerNonModifying(args[0].name, ctx)}(${writeExpression(resolved[1], ctx)}.begin_parse())`;
+            },
+        },
+    ],
+    [
+        "fromSlice",
+        {
+            name: "fromSlice",
+            resolve: (ctx, args, ref) => {
+                if (args.length !== 2) {
+                    throwError("fromSlice() expects one argument", ref);
+                }
+                if (args[0].kind !== "ref") {
+                    throwError(
+                        "fromSlice() is implemented only a struct type",
+                        ref,
+                    );
+                }
+                const tp = getType(ctx, args[0].name);
+                if (tp.kind !== "struct") {
+                    throwError(
+                        "fromSlice() is implemented only a struct type",
+                        ref,
+                    );
+                }
+                if (args[1].kind !== "ref" || args[1].name !== "Slice") {
+                    throwError(
+                        "fromSlice() expects a Slice as an argument",
+                        ref,
+                    );
+                }
+                return { kind: "ref", name: args[0].name, optional: false };
+            },
+            generate: (ctx, args, resolved, ref) => {
+                if (resolved.length !== 2) {
+                    throwError("fromSlice() expects one argument", ref);
+                }
+                if (args[0].kind !== "ref") {
+                    throwError(
+                        "fromSlice() is implemented only a struct type",
+                        ref,
+                    );
+                }
+                if (args[1].kind !== "ref" || args[1].name !== "Slice") {
+                    throwError(
+                        "fromSlice() expects a Slice as an argument",
+                        ref,
+                    );
+                }
+                return `${ops.readerNonModifying(args[0].name, ctx)}(${writeExpression(resolved[1], ctx)})`;
+            },
+        },
+    ],
 ]);
--- a/src/abi/struct.ts
+++ b/src/abi/struct.ts
@@ -52,14 +52,14 @@ export const StructFunctions: Map<string, AbiFunction> = new Map([
                 }
                 if (args[0].kind !== "ref") {
                     throwError(
-                        "fromCell() is implemented only a struct type",
+                        "fromCell() is implemented only for struct types",
                         ref,
                     );
                 }
                 const tp = getType(ctx, args[0].name);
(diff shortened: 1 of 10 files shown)


Gusarich:
Just realised that I didn't add any negative tests for struct parsing

anton-trunov (changes requested):
We need some tests to make sure the following works as expected.
First, the user defines a `fromCell`:

```
extends mutates fun fromCell(self: Cell, _: Cell): MyStruct1 {
  return MyStruct1 { ... };
}
```

And now the user is trying to do something weird:
```
contract StructsTester {
    receive() { }

    get fun fromCell1(src: Cell): MyStruct1 {
        let MyStruct1: Cell = beginCell().endCell();  // it's a cell here!
        let s: MyStruct1 = MyStruct1.fromCell(src);
        return s;
    }
}
```

What do you expect to happen in this case?

Gusarich:
@anton-trunov do you mean a user-defined function with the same name - `fromCell` or `fromSlice`?

I think it should work the same way as it works for all other built-ins such as `toCell` for structs, and `set`, `get` and `del` for maps.

anton-trunov:
@Gusarich Yep. Because right now this shadowing is allowed and it overrides the generated code.

Gusarich:
> @Gusarich Yep. Because right now this shadowing is allowed and it overrides the generated code.

Should we forbid such shadowing for all built-in functions?

UPD: oh, we actually wanted to do this already, but only implemented throwing errors for global.ts: #351

anton-trunov:
Yeah, we do not allow any form of shadowing in Tact. And if we do, it's a bug

anton-trunov:
btw, there is some merge conflicts

Gusarich:
> Yeah, we do not allow any form of shadowing in Tact. And if we do, it's a bug

In this case maybe we should create a separate issue about that and resolve that in another PR?

anton-trunov:
> In this case maybe we should create a separate issue about that and resolve that in another PR?

Yeah, there are issues like that, for instance #307. There is a special tag for those https://github.com/tact-lang/tact/labels/scope%3A%20var-scope

src/test/e2e-emulated/contracts/structs.tact
```diff
@@ -0,0 +1,50 @@
...
+
+contract StructsTester {    
+    receive() { }
+
+    get fun toCell1(s: MyStruct1): Cell {
+        return s.toCell();
+    }
+
+    get fun fromCell1(src: Cell): MyStruct1 {
+        let s: MyStruct1 = MyStruct1.fromCell(src);
```
anton-trunov:
```suggestion
        let s = MyStruct1.fromCell(src);
```

src/test/e2e-emulated/contracts/structs.tact
```diff
@@ -0,0 +1,50 @@
...
+        return s;
+    }
+
+    get fun fromSlice1(src: Slice): MyStruct1 {
+        let s: MyStruct1 = MyStruct1.fromSlice(src);
+        return s;
```
anton-trunov:
```suggestion
        return MyStruct1.fromSlice(src);
```

src/types/resolveExpression.ts
```diff
@@ -765,6 +765,20 @@ export function resolveExpression(
                         exp.ref,
                     );
                 }
+                // Handle struct calls
```
```suggestion
                // Handle static struct method calls
```

src/abi/struct.ts
```diff
@@ -42,4 +42,96 @@ export const StructFunctions: Map<string, AbiFunction> = new Map([
...
+        "fromCell",
+        {
+            name: "fromCell",
+            resolve: (ctx, args, ref) => {
+                if (args.length !== 2) {
+                    throwError("fromCell() expects one argument", ref);
+                }
+                if (args[0].kind !== "ref") {
+                    throwError(
+                        "fromCell() is implemented only a struct type",
```
anton-trunov:
```suggestion
                        "fromCell() is implemented only for struct types",
```

src/abi/struct.ts
```diff
@@ -42,4 +42,96 @@ export const StructFunctions: Map<string, AbiFunction> = new Map([
...
+                        ref,
+                    );
+                }
+                const tp = getType(ctx, args[0].name);
+                if (tp.kind !== "struct") {
+                    throwError(
+                        "fromCell() is implemented only a struct type",
```
anton-trunov:
```suggestion
                        "fromCell() is implemented only for struct types",
```

src/abi/struct.ts
```diff
@@ -42,4 +42,96 @@ export const StructFunctions: Map<string, AbiFunction> = new Map([
...
+                }
+                return { kind: "ref", name: args[0].name, optional: false };
+            },
+            generate: (ctx, args, resolved, ref) => {
+                if (resolved.length !== 2) {
+                    throwError("fromCell() expects one argument", ref);
+                }
+                if (args[0].kind !== "ref") {
+                    throwError(
+                        "fromCell() is implemented only a struct type",
```
anton-trunov:
```suggestion
                        "fromCell() is implemented only for struct types",
```

src/abi/struct.ts
```diff
@@ -42,4 +42,96 @@ export const StructFunctions: Map<string, AbiFunction> = new Map([
...
+        "fromSlice",
+        {
+            name: "fromSlice",
+            resolve: (ctx, args, ref) => {
+                if (args.length !== 2) {
+                    throwError("fromSlice() expects one argument", ref);
+                }
+                if (args[0].kind !== "ref") {
+                    throwError(
+                        "fromSlice() is implemented only a struct type",
```
anton-trunov:
```suggestion
                        "fromSlice() is implemented only for struct types",
```

src/abi/struct.ts
```diff
@@ -42,4 +42,96 @@ export const StructFunctions: Map<string, AbiFunction> = new Map([
...
+                        ref,
+                    );
+                }
+                const tp = getType(ctx, args[0].name);
+                if (tp.kind !== "struct") {
+                    throwError(
+                        "fromSlice() is implemented only a struct type",
```
anton-trunov:
```suggestion
                        "fromSlice() is implemented only for struct types",
```

src/abi/struct.ts
```diff
@@ -42,4 +42,96 @@ export const StructFunctions: Map<string, AbiFunction> = new Map([
...
+                }
+                return { kind: "ref", name: args[0].name, optional: false };
+            },
+            generate: (ctx, args, resolved, ref) => {
+                if (resolved.length !== 2) {
+                    throwError("fromSlice() expects one argument", ref);
+                }
+                if (args[0].kind !== "ref") {
+                    throwError(
+                        "fromSlice() is implemented only a struct type",
```
anton-trunov:
```suggestion
                        "fromSlice() is implemented only for struct types",
```

anton-trunov (changes requested):
LGTM! Just some minor tweaks in the suggested changes

anton-trunov (approved):

PR #432, opened 2024-06-18
feat/fix/refactor: full rewrite of constant evaluator

### LINKED ISSUE #284: Constant structs are not supported

### LINKED ISSUE #400: Storage vars cannot be initialized by struct instances

anton-trunov:
to fix various issues and support a larger Tact expression fragment. It is implemented as a classical tree-traversing interpreter. We take into account the semantics of TVM operations, including integer overflows.
As an optimizer it is not very efficient, though. But this will be handled separately.
Supported conditional expressions, unboxing operator, struct instances and struct field access.
Slices are not handled.

Closes #400.

Closes #284.

- ~[ ] I did not do unrelated and/or undiscussed refactorings~ Had to do some renaming


novusnota (dismissed):
Splendid work!

I wonder if this interpreter can be slightly adjusted elsewhere for evaluating Tact expressions with `--expr` or similar in the Tact compiler CLI. And this may even lead to a small REPL :)

anton-trunov:
> I wonder if this interpreter can be slightly adjusted elsewhere for evaluating Tact expressions with --expr or similar in the Tact compiler CLI. And this may even lead to a small REPL :)

You just need to expose the `evalConstantExpression` function from `constEval.ts` via CLI. It will parse the input expression and feed the AST into `evalConstantExpression`. A really nice thing to do would be to also compile the expression and evaluate the code with Sandbox in the background to make sure the results are always the same. And if not -- throw an error and ask to report the example as an issue.

anton-trunov:
@novusnota And them we could make a Telegram bot that evaluates those Tact expressions and integrate it to https://t.me/tactlang :)

PR #447, opened 2024-06-21
feat: `throw` and `nativeThrow` in return reachability analysis

### LINKED ISSUE #421: `return`-reachability analysis should take unconditional `throw` into account
For instance, something like this should not result in an error message from the analysis:
```
fun test1(): Int {
  throw(200);
}
```

or, a bit more realistic example:
```
fun test2(c: Bool): Int {
  if (c) {
    throw(200);
  } else {
    return 42;
  }
}
```

anton-trunov:
Closes #421


novusnota (dismissed):
Cool! There are also `nativeThrowWhen()` and `nativeThrowUnless()`, but those aren't checked as easily :)

anton-trunov:
> There are also nativeThrowWhen() and nativeThrowUnless(), but those aren't checked as easily :)

The return reachability analysis we use is really dumb: it won't check if an arbitrary function always throws and it won't try to determine if the condition is always true/false for `nativeThrowWhen()` and `nativeThrowUnless()`, it won't analyze assembly wrapped into a FunC function, etc.

Gusarich (approved):

PR #467, opened 2024-06-22
fix: name clashes with FunC keywords for struct constructor parameters

### LINKED ISSUE #466: Struct constructor parameters clash with FunC keywords
Trying to compile the following
```tat
struct Foo {type: Int}        // or `forall: 42`, `int: 42`, etc.

contract Bar {
    f: Foo = Foo {type: 42};
}
```

would result in this FunC error message:

```
Func compilation error: ... formal parameter name expected instead of `type`
  ((int)) $Foo$_constructor_type(int type) inline;
```

anton-trunov:
Closes #466


novusnota:
Cool, but seems like this ties us to FunC even more. Maintenance of the keyword list may bring occasional bugs in the future. I wonder if we could use different names, like prefixing things with underscores or even UUID's of sort?

anton-trunov:
> Cool, but seems like this ties us to FunC even more.

No, it does not, it does quite the opposite. The test shows you can use all current FunC keywords as your struct field identifiers. And it's highly unlikely FunC will have keywords starting with `$`.

PR #507, opened 2024-06-28
Add one more failing test case for structs parsing

### LINKED ISSUE #472: Incorrect bit length in `.storeUint()` and `.storeInt()` functions fails silently in presence of serialization/storage modifiers when doing parsing via `Struct.fromCell()`
Consider the following snippet:

```tact
import "@stdlib/deploy";

struct Coin {
    first: Int as coins;
    second: Int as uint32;
}

fun directParse(payload: Cell): Coin {
    return Coin.fromCell(payload);
}

contract Example with Deployable {
    receive("example") {
        let coin = directParse(
            beginCell()
            .storeCoins(42)
            .storeUint(35, 31) // here, I only specified 31 bit for storing the Uint, but could've specified 33 bits and get the same issue
            .endCell()
        );
        dump(coin.first); // would display nothing!
        dump(coin.second); // neither would this!
    }
}
```

Although it's up to user to correctly specify serialization modifiers, such as `uint32` and then use `.storeUint(NUM_TO_STORE, 32)` and similar, it would be nice to display an error when mistakes occur.

This may be an edge case of some bigger issue.

Also it's unknown what value `coin` has in the showcased example, because an attempt to `dump(coin)` will be prohibited as `dump()` doesn't support Structs and Messages directly and type of `coin` is still considered to be `Coin`.

---

P.S.: Kinda offtopic, but this reminded me that we may consider adding `.storeUint32()`, `.storeUint64()` and similar functions for convenience. Or, to make a special function for Ints, which would accept serialization modifiers, like so:

```tact
// Option 1
beginCell().storeValue(VALUE as uint32);

// Option 2
beginCell().storeValue(VALUE, uint32);
```

Gusarich:
Closes #472


anton-trunov:
CI failing because of some changed addresses

anton-trunov:
Interestingly enough CI now only fails on Windows and it's not because of timeouts

anton-trunov (approved):

PR #509, opened 2024-06-28
Enhance error handling to support third-party tools

### LINKED ISSUE #262: Enhance error handling in Tact compiler to support third-party tools
Currently, the Tact compiler logs all errors directly to the console. For integration with third-party tools like [Nujan Web IDE](https://ide.nujan.io/), it's necessary to modify the error handling behavior to throw these errors, rather than just logging them.

**Expected Behavior:**
Errors should be thrown programmatically, allowing third-party tools to catch and process these errors accordingly.

**Additional Context:**
This enhancement is crucial for improving the interoperability of the Tact compiler with various third-party development tools, which rely on catching exceptions to handle errors more effectively.

rahulyadav-57:
Closes #262 

- ~[ ] I have added tests to demonstrate the contribution is correctly implemented: this usually includes both positive and negative tests, showing the happy path(s) and featuring intentionally broken cases~


scripts/prepare.ts
```diff
@@ -64,15 +66,15 @@ import { __DANGER__disableVersionNumber } from "../src/pipeline/version";
                             content: code,
                         },
                     ],
-                    logger: consoleLogger,
+                    logger: logger,
                 });
                 if (!c.ok) {
-                    console.error(c.log);
+                    logger.error(c.log);
                     process.exit(1);
```
anton-trunov: My guess would be that all things like `process.exit` should also go

anton-trunov:
Any progress here?

src/logger.ts
```diff
@@ -1,9 +1,87 @@
...
+
+    protected log(level: LogLevel, message: messageType) {
+        if (this.level === LogLevel.NONE) {
+            return;
+        }
+
+        if (message instanceof Error) {
+            message = message.stack || message.message;
+        } else {
+            message = "" + message.toString();
```
byakuren-hijiri: Can we simply use `message.toString()` here?

src/pipeline/build.ts
```diff
@@ -326,16 +336,18 @@ export async function build(args: {
                 bindingsServer,
             );
         } catch (e) {
-            logger.error("Bindings compiler crashed");
-            logger.error(errorToString(e));
-            return false;
+            const message = "Bindings compiler crashed";
+            logger.error(message);
+            logger.error(e as Error);
+            errorMessages.push(new Error(message));
```
byakuren-hijiri: Should we also include `e` to the `errorMessages` structure? It might be useful to third-parties, I think.

src/pipeline/build.ts
```diff
@@ -344,11 +356,13 @@ export async function build(args: {
             );
             project.writeFile(pathBindings, report);
         } catch (e) {
-            logger.error("Report generation crashed");
-            logger.error(errorToString(e));
-            return false;
+            const message = "Report generation crashed";
+            logger.error(message);
+            logger.error(e as Error);
+            errorMessages.push(new Error(message));
```
byakuren-hijiri: See previous

src/node.ts
```diff
@@ -64,33 +65,42 @@ export async function run(args: {
     configPath?: string;
     projectNames?: string[];
     additionalCliOptions?: AdditionalCliOptions;
+    suppressLog?: boolean;
 }) {
     const configWithRootPath = await loadConfig(args.fileName, args.configPath);
     if (!configWithRootPath) {
-        return false;
+        return { ok: false, error: [new Error("Unable to load config")] };
```
byakuren-hijiri: nit: Maybe we should add `args.configPath` to the error message in order to give more context to the user.

src/pipeline/build.ts
```diff
@@ -8,32 +8,32 @@ import { funcCompile } from "../func/funcCompile";
...
 import { idText } from "../grammar/ast";
+import { TactErrorCollection } from "../errors";
 
 export async function build(args: {
     config: ConfigProject;
     project: VirtualFileSystem;
     stdlib: string | VirtualFileSystem;
-    logger?: TactLogger | null | undefined;
-}) {
+    logger?: Logger | null | undefined;
```
byakuren-hijiri: This type definition looks a bit strange, right? What is the different between `null` and `undefined` values? What value will be used if we write `if (!logger) {` in the function body?

src/pipeline/build.ts
```diff
@@ -111,7 +112,7 @@ export async function build(args: {
         let codeEntrypoint: string;
 
         // Compiling contract to func
-        logger.log("   > " + contract + ": tact compiler");
+        logger.info("   > " + contract + ": tact compiler");
```
byakuren-hijiri: nit: Maybe we should use string interpolation in the similar cases instead. AFAIK we prefer that style when refactoring the project, but it's up to you.

src/pipeline/build.ts
rahulyadav-57: Yes, that would be a better.

src/pipeline/build.ts
rahulyadav-57: I think we can remove both `null` and `undefined` since it will be `undefined` by default.

src/node.ts
rahulyadav-57: We can change it to something \`Unable to load config from path: ${args.configPath}\`

src/pipeline/build.ts
rahulyadav-57: Yes, I'll do it. I haven't done the refactoring yet, but I'll take care of it.

anton-trunov (dismissed):
LGTM, but let's wait for @byakuren-hijiri to approve

bin/tact
```diff
@@ -165,9 +165,9 @@ meowModule.then(
         projectNames: cli.flags.projects ?? [],
         additionalCliOptions: { mode },
       })
-      .then((success) => {
+      .then((response) => {
```
novusnota, line 168: Two points here:
1. It would be nice to add a `--quiet` flag, which would be passed to `suppressLog: ...` argument of `run()`
2. If there's no `--quiet` flag, it would be nice to peek into `response.error` and display them onto console before we do `process.exit(...)`

Other than that, splendid work!

anton-trunov: @novusnota can you please help with this flag since you already have experience with meow? just push your commit to this PR

novusnota: Sure!

novusnota: Hmm, seems like adding `response.error` doubles the error output:

```ts
// bin/tact
// ...
main
  .run({
    fileName: cli.input.at(0),
    configPath: cli.flags.config,
    projectNames: cli.flags.projects ?? [],
    additionalCliOptions: { mode },
    suppressLog: cli.flags.quiet
  })
  .then((response) => {
    // Log errors if --quiet is not set
    if (!cli.flags.quiet) {
      response.error.map(err => console.log(err)); // ← this logs errors AFTER the main log output with them showed already, which isn't nice
    }
    // https://nodejs.org/docs/v20.12.1/api/process.html#exit-codes
    process.exit(response.ok ? 0 : 30);
  });
```

Not a big deal though, can omit the `reponse.error.map...`

scripts/prepare.ts
```diff
@@ -15,74 +15,84 @@ import { __DANGER__disableVersionNumber } from "../src/pipeline/version";
...
-        ),
-    )) {
-        const res = await verify({ pkg: fs.readFileSync(pkgPath, "utf-8") });
-        if (!res.ok) {
-            console.error("Failed to verify " + pkgPath + ": " + res.error);
-            process.exit(1);
+    try {
+        // Compile projects
+        const compileResult = await run({
+            configPath: path.join(__dirname, "../tact.config.json"),
```
byakuren-hijiri:
```suggestion
            configPath: path.join(__dirname, "..", "tact.config.json"),
```

scripts/prepare.ts
```diff
@@ -15,74 +15,84 @@ import { __DANGER__disableVersionNumber } from "../src/pipeline/version";
...
-                            content: code,
-                        },
-                    ],
-                    logger: logger,
-                });
-                if (!c.ok) {
-                    logger.error(c.log);
-                    process.exit(1);
+        // Compile func files
+        for (const p of [{ path: path.join(__dirname, "/../func/") }]) {
```
byakuren-hijiri:
```suggestion
        for (const p of [{ path: path.join(__dirname, "..", "func") }]) {
```

scripts/prepare.ts
```diff
@@ -15,74 +15,84 @@ import { __DANGER__disableVersionNumber } from "../src/pipeline/version";
...
-                process.exit(1);
-            }
-            fs.writeFileSync(p.path + r + ".fift", c.fift!);
-            fs.writeFileSync(p.path + r + ".cell", c.output!);
 
-            // Cell -> Fift decompiler
-            const source = decompileAll({ src: c.output! });
-            fs.writeFileSync(p.path + r + ".rev.fift", source);
+                // Precompile
+                logger.info(`Processing ${path.join(p.path + r)}`);
```
byakuren-hijiri: nit: I would create a local variable `p.path+r` and reuse it here and below. It could save us from regression errors when refactoring this path later. But it's up to you.

src/pipeline/build.ts
```diff
@@ -336,10 +336,10 @@ export async function build(args: {
                 bindingsServer,
             );
         } catch (e) {
-            const message = "Bindings compiler crashed";
-            logger.error(message);
-            logger.error(e as Error);
-            errorMessages.push(new Error(message));
+            const error = e as Error;
+            error.message = `Bindings compiler crashed, ${error.message}`;
```
byakuren-hijiri:
```suggestion
            error.message = `Bindings compiler crashed: ${error.message}`;
```

src/pipeline/build.ts
```diff
@@ -356,10 +356,10 @@ export async function build(args: {
             );
             project.writeFile(pathBindings, report);
         } catch (e) {
-            const message = "Report generation crashed";
-            logger.error(message);
-            logger.error(e as Error);
-            errorMessages.push(new Error(message));
+            const error = e as Error;
+            error.message = `Report generation crashed, ${error.message}`;
```
byakuren-hijiri:
```suggestion
            error.message = `Report generation crashed: ${error.message}`;
```

src/node.ts
```diff
@@ -98,7 +115,7 @@ export async function run(args: {
         false,
     ); // Improves developer experience
     for (const config of projects) {
-        consoleLogger.log("💼 Compiling project " + config.name + "...");
+        logger.info("💼 Compiling project " + config.name + "...");
```
byakuren-hijiri:
```suggestion
        logger.info(`💼 Compiling project ${config.name} ...`);
```

src/pipeline/build.ts
```diff
@@ -141,7 +143,7 @@ export async function build(args: {
         }
 
         // Compiling contract to TVM
-        logger.log("   > " + contract + ": func compiler");
+        logger.info("   > " + contract + ": func compiler");
```
byakuren-hijiri:
```suggestion
        logger.info(`   > ${contract}: func compiler`);
```

src/pipeline/build.ts
```diff
@@ -192,39 +196,42 @@ export async function build(args: {
 
         if (config.mode === "fullWithDecompilation") {
             // Fift decompiler for generated code debug
-            logger.log("   > " + contract + ": fift decompiler");
+            logger.info("   > " + contract + ": fift decompiler");
```
byakuren-hijiri:
```suggestion
            logger.info(`   > ${contract}: fift decompiler`);
```

src/pipeline/build.ts
```diff
@@ -192,39 +196,42 @@ export async function build(args: {
...
     const contracts = getContracts(ctx);
     const packages: PackageFileFormat[] = [];
     for (const contract of contracts) {
-        logger.log("   > " + contract);
+        logger.info("   > " + contract);
         const artifacts = built[contract];
         if (!artifacts) {
-            logger.error("   > " + contract + ": no artifacts found");
-            return false;
+            const message = "   > " + contract + ": no artifacts found";
```
byakuren-hijiri:
```suggestion
            const message = `   > ${contract}: no artifacts found`;
```

src/pipeline/build.ts
```diff
@@ -298,17 +307,18 @@ export async function build(args: {
     }
 
     // Bindings
-    logger.log("   > Bindings");
+    logger.info("   > Bindings");
     for (const pkg of packages) {
-        logger.log("   > " + pkg.name);
+        logger.info("   > " + pkg.name);
```
byakuren-hijiri:
```suggestion
        logger.info(`   > ${pkg.name}`);
```

src/pipeline/build.ts
```diff
@@ -298,17 +307,18 @@ export async function build(args: {
...
         if (pkg.init.deployment.kind !== "system-cell") {
-            logger.error(
+            const message =
                 "   > " +
-                    pkg.name +
-                    ": unsupported deployment kind " +
-                    pkg.init.deployment.kind,
-            );
-            return false;
+                pkg.name +
```
byakuren-hijiri: Let's use string interpolation here

byakuren-hijiri (dismissed):
Let's refactor existing sources a little bit and fix the operations with paths to avoid cross-platform problems. After that, LGTM.

anton-trunov:
I just pushed two fixes, let's see if the tests are green and merge it

anton-trunov (approved):

PR #525, opened 2024-07-03
Fix `newAddress` constant evaluation

### LINKED ISSUE #515: Consider making a call to `newAddress(0, 0)` return a null/zero address
The following snippet:

```tact
contract Example {
   receive("example") {
       dump(newAddress(0, 0));
   }
}
```

Will cause a compilation error:

```shell
Tact compilation failed
Error: Invalid address hash length: 0
```

That's not a bug per se, however the expected behaviour is to produce a [null/zero address](https://tonscan.org/address/UQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAJKZ). In the past, that's how that function worked, see [example on the right in Tact by Example](https://tact-by-example.org/02-addresses).

This was probably overlooked in #237 when resolving #214.

---

Alternative solution would be to introduce a `nullAddress()` function, like I mentioned here: https://github.com/tact-lang/tact/issues/214#issuecomment-2032947204

Gusarich:
Closes #515


anton-trunov (approved):
Cool!

Issue #435, opened 2024-06-18
Partial evaluation for const-eval optimization


anton-trunov:
As of now, the constant evaluator's usage is two-fold:
- evaluate compile-time constants' initializers to their values and stop compilation if this cannot be done;
- as an optimization technique: try to evaluate expressions to their values to use the values to produce target code.

The type of the constant evaluator is `Expression -> Value`.

As an optimization technique, the simple interpreter we use is quite limited, e.g. `2 * 3 * 7 * x` where `x` is an arbitrary variable won't be simplified to `42 * x` because evaluation gets stuck on variables. Instead we can try to produce a simplified AST, so we need a partial evaluator of type `Expression -> Expression` where the return expression should have the same semantics but is (non-strictly) smaller.

We can even apply some algebraic transformations, so that even `x * 2 * 3 * 7` (which gets parsed as `((x * 2) * 3 * )7`) gets transformed to `x * 42` or `42 * x`. However, this should be done with caution, because in the presence of overflows some arithmetic operations, for instance `+`, are not associative.

jeshecdom:
Is it OK if potential overflows are removed from the expression? For example, consider this very silly example:
```
const foo: Int =  x + MAX - MAX
```
This may cause an overflow if interpreted as `(x + MAX) - MAX`. But let us suppose that the evaluator cannot evaluate `x` and decides to associate the expression as `x + (MAX - MAX)`. Then, the overflow goes away and `foo = x`. The final value in `foo` is not semantically equivalent to the value of `foo` in the original program (the original program may overflow).

anton-trunov:
I think we could do this, but it should be an opt-in optimization strategy, so by default we should preserve the semantics fully and if the user wants those kinds of more aggressive optimizations that only preserve the semantics morally, they can specify it in the project settings

jeshecdom:
The algorithm will apply several rewriting rules in sequence. These rules ensure that the semantics of the expression are preserved. In particular, for **integer expressions**, the semantics are preserved if the following holds, where `O` denotes the original expression and `T` the transformed expression:
```
(O overflows) IF AND ONLY IF (T overflows)
and
IF (O does not overflow) THEN val(O) = val(T)
```
For **non-integer expressions**, we must have:
```
val(O) = val(T)
```
A more aggressive optimizer weakens the biconditional to a simple implication:
```
IF (O does not overflow) THEN (T does not overflow)
and
IF (O does not overflow) THEN val(O) = val(T)
```
[...]

PR #528, opened 2024-07-04
First version of partial evaluator: support for +,-,* operators and associative rules

jeshecdom:
Towards #435.

- First version of partial evaluator. Currently it is able to simplify expressions involving +, -, and *. However, there are some algebraic rules involving those operators that have not been added yet.
- I decided to keep the partial evaluator separate from the constant evaluator because the partial evaluator still does not support certain language constructions, like structs: the partial evaluator needs to create AST nodes out of values, and there is currently no code to construct ASTStruct nodes out of StructValue. This will be added in the future.
- Currently, to call the partial evaluator, one needs to call the function  `partiallyEvalExpression`. In the future, the constant evaluator will call the partial evaluator, because they should give the same result for expressions that can fully evaluate.
- Many algebraic rules are still pending, including the rest of integer operators and boolean operators.
- Still lacks documentation, specially the explanation of how each rule works.
- Since the partial evaluator is currently isolated from the rest of the code, the tests directly check the partial evaluator instead of going through a contract.


anton-trunov:
@jeshecdom awesome stuff! could you please update your PR and pull in the most recent changes from `main` and also resolve the merge conflicts?

src/constEval.ts
```diff
@@ -97,58 +116,81 @@ function ensureMethodArity(
...
-function evalUnaryOp(
+export function evalUnaryOp(op: AstUnaryOperation, valOperand: Value): Value {
+    return __evalUnaryOp(op, valOperand, DUMMY_LOCATION, DUMMY_LOCATION);
+}
+
+function __evalUnaryOp(
     op: AstUnaryOperation,
-    operand: AstExpression,
+    valOperand: Value,
+    operandRef: SrcInfo,
```
anton-trunov:
```suggestion
    operandLoc: SrcInfo,
```

src/constEval.ts
```diff
@@ -97,58 +116,81 @@ function ensureMethodArity(
...
-    // which is `-(2**256)` we need to have a special case for it
-    if (operand.kind === "number" && op === "-") {
-        // emulating negative integer literals
-        return ensureInt(-operand.value, source);
-    }
-    const valOperand = evalConstantExpression(operand, ctx);
     switch (op) {
         case "+":
-            return ensureInt(valOperand, operand.loc);
+            return ensureInt(valOperand, operandRef);
```
anton-trunov: we should probably just make the location argument of `ensureInt` optional (the default value would be `DUMMY_LOCATION`), so here the function calls would look like `ensureInt(valOperand)`

src/optimizer/types.ts
```diff
@@ -0,0 +1,72 @@
...
+import {
+    AstExpression,
+    AstNumber,
+    AstBoolean,
+    AstNull,
+    AstString,
+    SrcInfo,
+} from "../grammar/ast";
+
+export type ValueExpression = AstNumber | AstBoolean | AstNull | AstString;
```
anton-trunov: Let's rename `ValueExpression` to something different (`AstValue`?), I'm getting headaches from this one :)

src/optimizer/types.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export const DUMMY_INTERVAL: Interval = {
```
anton-trunov: you should be able to reuse the existing dummy definitions from grammar.ts: https://github.com/tact-lang/tact/blob/d8893ca5744772fddabd2071e31931f2b4b83b9b/src/grammar/grammar.ts#L70-L71

src/optimizer/types.ts
```diff
@@ -0,0 +1,72 @@
...
+        };
+    },
+};
+export const DUMMY_LOCATION: SrcInfo = new SrcInfo(
+    DUMMY_INTERVAL,
+    null,
+    "user",
+);
+
+export interface ExpressionTransformer {
```
anton-trunov: let's not use interfaces in our codebase: they have some weird properties related to 'interface merging"

src/optimizer/util.ts
```diff
@@ -0,0 +1,303 @@
...
+        case "number":
+            return ast2.kind === "number" ? ast1.value === ast2.value : false;
+        case "string":
+            return ast2.kind === "string" ? ast1.value === ast2.value : false;
+        case "id":
+            return ast2.kind === "id" ? ast1.text === ast2.text : false;
+        case "method_call":
+            if (ast2.kind === "method_call") {
+                return (
+                    ast1.method.text === ast2.method.text &&
```
anton-trunov: Let's use the identifier comparison function `eqNames` here and elsewhere:
```suggestion
                    eqNames(ast1.method, ast2.method) &&
```

src/optimizer/util.ts
```diff
@@ -0,0 +1,303 @@
...
+
+// precondition: the divisor is not zero
+// rounds the result towards negative infinity
+// Uses the fact that a / b * b + a % b == a, for all b != 0.
+export function modFloor(a: bigint, b: bigint): bigint {
+    return a - divFloor(a, b) * b;
+}
+
+// Test equality of ASTExpressions.
+export function areEqualExpressions(
```
anton-trunov: this should be moved to `src/grammar/ast.ts` and renamed to `eqExpressions` (there is `eqNames`) to compare identifiers

src/optimizer/util.ts
```diff
@@ -0,0 +1,303 @@
...
+    ast1: AstExpression,
+    ast2: AstExpression,
+): boolean {
+    switch (ast1.kind) {
```
anton-trunov: I'm wondering if checks like `ast2.kind === "boolean"` would be redundant if we check `ast1.kind === ast2.kind` before the switch?. That would simplify some code

src/test/e2e-emulated/partial-eval.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    // The following 4 cases should NOT simplify to X * 0.
+    // the reason is that X could be MAX, so that X*3 causes
+    // an overflow, but X*0 does not.
+    { original: "X * 3 * 0", simplified: "X * 3 * 0" },
+    { original: "3 * X * 0", simplified: "3 * X * 0" },
+    { original: "0 * (X * 3)", simplified: "0 * (X * 3)" },
+    { original: "0 * (3 * X)", simplified: "0 * (3 * X)" },
+
+    { original: "X * 0 * 3", simplified: "X * 0" },
```
anton-trunov: wondering why this is not `simplified: 0`, is it because the corresponding rule is not implemented yet?

anton-trunov (changes requested):
At this point I did some superficial code style review mostly, let's merge things, run the linter with the newly introduced checks and then I'll switch to reviewing the algorithm. (from a quick glance it's impressive already)

anton-trunov:
There are merge conflicts

src/grammar/ast.ts
```diff
@@ -679,6 +679,128 @@ export function __DANGER_resetNodeId() {
...
+): boolean {
+    if (ast1.kind !== ast2.kind) {
+        return false;
+    }
+
+    switch (ast1.kind) {
+        case "null":
+            return true;
+        case "boolean":
+            return ast1.value === (ast2 as AstBoolean).value;
```
anton-trunov, line 697: `ast2 as AstBoolean` is a type assertion, so it will fail at run-time if `ast2` is not an actual boolean constant. Basically, you either need to do pattern-matching on both arguments, or catch exceptions and return `false`. Let's refactor this and add some unit tests for `eqExpressions`. We will need this to resolve part of #314 (equivalence checks between ASTs).

src/optimizer/types.ts
```diff
@@ -0,0 +1,20 @@
+import {
+    AstExpression,
+    AstNumber,
+    AstBoolean,
+    AstNull,
+    AstString,
+} from "../grammar/ast";
+
+export type AstValue = AstNumber | AstBoolean | AstNull | AstString;
```
anton-trunov: Would it make sense to move this to `ast.ts`?

src/optimizer/util.ts
```diff
@@ -0,0 +1,167 @@
...
+
+// Checks if top level node is a binary op node
+// with a value node on the left and
+// non-value node on the right
+export function checkIsBinaryOp_Value_NonValue(ast: AstExpression): boolean {
+    if (ast.kind === "op_binary") {
+        return isValue(ast.left) && !isValue(ast.right);
+    } else {
+        return false;
+    }
```
anton-trunov:
```suggestion
    return ast.kind === "op_binary" && isValue(ast.left) && !isValue(ast.right)
```

src/constEval.ts
```diff
@@ -645,25 +702,28 @@ function interpretEscapeSequences(stringLiteral: string) {
     );
 }
 
+function lookupID(ast: AstId, ctx: CompilerContext): Value {
```
anton-trunov:
```suggestion
function lookupName(ast: AstId, ctx: CompilerContext): Value {
```

this is to make it in one style with `eqNames`

btw, thanks for factoring out this as a standalone function: this looks much better than being inlined

src/constEval.ts
```diff
@@ -681,9 +741,9 @@ export function evalConstantExpression(
         case "string":
             return ensureString(interpretEscapeSequences(ast.value), ast.loc);
         case "op_unary":
-            return evalUnaryOp(ast.op, ast.operand, ast.loc, ctx);
+            return fullyEvalUnaryOp(ast.op, ast.operand, ast.loc, ctx);
```
anton-trunov, line 770: maybe we should rename everything to `fullyEvalX` in the `evalConstantExpression` function? for consistency reasons, like `evalMethod` -> `fullyEvalMethod`, etc. We can put a comment that it's ongoing work to make all of those actually fully evaluate

anton-trunov (changes requested):

src/grammar/ast.ts
```diff
@@ -679,4 +679,316 @@ export function __DANGER_resetNodeId() {
...
+            case "static_call":
+                return (
+                    eqNames(ast1.function, (ast2 as AstStaticCall).function) &&
+                    eqExpressionArrays(ast1.args, (ast2 as AstStaticCall).args)
+                );
+        }
+    } catch (e) {
+        // In principle, the assertions "as Ast___" should not fail
+        // because ast1 and ast2 have the same kind inside the switch.
+        return false;
```
anton-trunov: ah, sorry, I missed you had

```
if (ast1.kind !== ast2.kind) {
        return false;
    }
```

above

`try-catch` isn't needed here, my bad

src/grammar/ast.ts
```diff
@@ -679,4 +679,316 @@ export function __DANGER_resetNodeId() {
...
+    for (let i = 0; i < arr1.length; i++) {
+        if (!eqExpressions(arr1[i]!, arr2[i]!)) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+export function traverse(node: AstNode, callback: (node: AstNode) => void) {
```
anton-trunov: `traverse` has been moved to `grammar/iterators.ts` recently

src/optimizer/util.ts
```diff
@@ -0,0 +1,163 @@
...
+    AstExpression,
+    AstUnaryOperation,
+    AstBinaryOperation,
+    createAstNode,
+} from "../grammar/ast";
+import { dummySrcInfo } from "../grammar/grammar";
+import { Value } from "../types/types";
+import { AstValue } from "./types";
+
+export function isValue(ast: AstExpression): boolean {
```
anton-trunov: let's move `isValue` to `ast.ts`

src/test/e2e-emulated/expr-equality.spec.ts
anton-trunov, line 1: this test should be moved to `src/grammar/test`, because it does not use the Sandbox library to run compiled contracts in a simulated environment (that's what `e2e-emulated` means)

anton-trunov (changes requested):

anton-trunov (approved):

PR #535, opened 2024-07-05
fix: unicode code points

### LINKED ISSUE #531: Two small issues with unicodeCodePoint escape sequence in literal strings
### Incorrect regex

Regex here (in constant evaluator):
https://github.com/tact-lang/tact/blob/b119b7baac9f8bfa61484757bd226def0d858ab4/src/constEval.ts#L607

Doesn't match the grammar:
https://github.com/tact-lang/tact/blob/b119b7baac9f8bfa61484757bd226def0d858ab4/src/grammar/grammar.ohm#L286

As it allows more than six hexadecimal digits to be specified, while grammar allows from 1 to 6. In that regex on top, the `+` has to be replaced with `{1,6}`.

### Weird bug with specifying from 1 to 6 zeros

The following:

```tact
dump("\u{00000}"); // any number of zeros between 1 and 6 does this
```

...leads to this ugly error message:

```shell
/playground/build/Playground/tact_Playground.code.fc:25:22: error: undefined function `__gen_slice_string_dbbc7025e89eb9b1ec9e8e2c7a2db6869dbb50fba21bf374c86529dc311cede9`, defining a global function of unknown type
      __tact_debug_str(__gen_slice_string_dbbc7025e89eb9b1ec9e8e2c7a2db6869dbb50fba21bf374c86529dc311cede9(), "File contracts/playground.tact:69:9");
                       ^
```

Same goes for `"\u0000"` and `"\x00"` too.

P.S: The maximum valid Unicode code point is U+10FFFF, which makes `dump("\u{110000}")` (and bigger) fail with an RangeError ­— that error can be handled a bit more gracefully :)

novusnota:
Closes #531

- ~~[ ] I have documented my contribution in Tact Docs: https://github.com/tact-lang/tact-docs/pull/PR-NUMBER~~ Not applicable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Introduced a streamlined error logger for compilation pipeline to support third-party tools: PR [#509](https://github.com/tact-lang/tact/pull/509)
 - Collisions of PascalCase getter names in generated wrappers are now checked: PR [#556](https://github.com/tact-lang/tact/pull/556)
 - Display a clearer error in case the source code file is missing when using Tact's CLI: PR [#561](https://github.com/tact-lang/tact/pull/561)
+- Error messages for unicode code points outside of valid range: PR [#535](https://github.com/tact-lang/tact/pull/535)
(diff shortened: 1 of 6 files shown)


anton-trunov (changes requested):
We need a test demonstrating the FunC compiler can process comments with the null character

anton-trunov (approved):

Issue #533, opened 2024-07-05
Inconsistent ABI generation for projects with multiple contracts


Gusarich:
Found while debugging #516

ABI generation seems to go wrong when the project has multiple contracts.
Small reproducable example is below.

Create a new Tact project (doesn't matter Blueprint or not) with two contracts with separate build output directorie specified.

Contract A:
```tact
import "@stdlib/deploy";

contract A with Deployable {
    receive() {}
}
```

Contract B:
```tact
import "@stdlib/deploy";
import "./a.tact";

contract B with Deployable {
    receive() {
        require(1 == 1, "1 is not equal to 1");
    }

    get fun test(): Address {
        return contractAddress(initOf A());
    }
}
```

Now build that project and look at the build directory, which will have separate `A` and `B` project builds. Since `B` depends on `A`, the `A` contract will be compiled separately for `B` project and will be included in that directory.
It is completely normal behaviour, but if we look at the resulting ABI of `A` in both of these directories, it'll differ in the presence of `1 is not equal to 1` error code. This leads to the different compilation results due to hash change in IPFS link to ABI.
Same behaviour can be achieved for `types` ABI field too by defining some struct in `B` and not in `A`.

The order of elements in `types` ABI field can also be shuffled in some cases, but I didn't come up with a small reproducable example. I believe it has something to do with the order of Object items in JS (Objects are used for storing all types during compilation and not Maps).

Gusarich:
depends on #545

anton-trunov:
An internal discussion with @Gusarich has lead us to the decision of not to change ABI generation.

PR #545, opened 2024-07-10
Refactor AST traversing

### LINKED ISSUE #541: `traverse` function does not perform a complete traverse of AST
For example, traversing `ASTStatementLet` currently looks this way:
https://github.com/tact-lang/tact/blob/cfcb4f04ce5155e0d6638c0b6b1f7a920f5c1792/src/grammar/ast.ts#L759-L761
while it should look like this:
```ts
if (node.kind === "statement_let") {
    traverse(node.name, callback);
    if (node.type) traverse(node.type, callback);
    traverse(node.expression, callback);
}
```
becase `AstStatementLet` includes two other AST nodes apart from `expression`:
https://github.com/tact-lang/tact/blob/cfcb4f04ce5155e0d6638c0b6b1f7a920f5c1792/src/grammar/ast.ts#L188-L195

And this problem exists for almost all ASTNode types in `traverse` function, which can cause problems in `resolveDescriptors`, especially if we want to perform some new traversals as for resolving #533.

Gusarich:
Closes #541 


src/grammar/ast.ts
```diff
@@ -217,14 +217,14 @@ export type AstStatementAssign = {
...
 export type AstAugmentedAssignOperation =
-    | "+"
-    | "-"
-    | "*"
-    | "/"
-    | "%"
-    | "|"
-    | "&"
-    | "^";
+    | { kind: "augmentedassign_operation"; operation: "+" }
```
anton-trunov: I'm actually not at all convinced we should add these tags to operators and attributes here and below. For instance, I need to collect attributes for a function, I'd traverse only the function definitions and native function declarations. And if you need some info on augmented assignment operator you 100% need to look at the AST node that actually includes meaningful information like the path and the right-hand side expression. I propose to revert this part of the refact and only introduce things like this if there is ever a use case

src/grammar/ast.ts
```diff
@@ -626,23 +632,29 @@ export type AstReceiverKind =
...
-    | AstConstantDecl;
+    | AstModuleItem
+    | AstTypeDecl
+    | AstContractDeclaration
+    | AstTraitDeclaration
+    | AstStatement
+    | AstAugmentedAssignOperation
+    | AstType
+    | AstTypeId
+    | AstOptionalType
```
```
    | AstTypeId
    | AstOptionalType
    | AstMapType
    | AstBouncedMessageType
```

should be covered by `AstType`

anton-trunov:
What is the minimal change that is needed to resolve #533?

Gusarich:
> What is the minimal change that is needed to resolve #533?

I think the change of `traverse` function is enough and we don't need other refactors for that.
Should I just revert all the refactors that I did here and only leave `traverse` function rewrite?

anton-trunov:
yeah, let's keep the minimal changes

anton-trunov:
btw, I broke already broken CI for Windows, so please ignore that for now

anton-trunov (commented):
LGTM. Let's also remove `iterators.ts` from the list of files ignored by Knip: https://github.com/tact-lang/tact/blob/eb9ec73920be7830a072b98c87f1e79eb23949ef/knip.json#L7, since it's now being used in other modules within Tact

anton-trunov (dismissed):

anton-trunov:
the unused deps marked by Knip need to be made `@public`, it should fix the failing check:

```
$ knip
Unused exports (4)
foldExpressions    function  src/grammar/iterators.ts:182:17
forEachStatement   function  src/grammar/iterators.ts:393:17
foldStatements     function  src/grammar/iterators.ts:501:17
forEachExpression  function  src/grammar/iterators.ts:8:17
```

PR #557, opened 2024-07-11
Rework generation of `types` and `errors` ABI fields

### LINKED ISSUE #516: `initOf` returns different addresses from same arguments
Hi, 
There is two line of code with same arguments, returning DIFFERENT address of contract, based on initOf:

`        dump(contractAddress(initOf BankJettonWallet(owner_address, self.jetton_master)));`
(here: https://github.com/Architec-Ton/banktoken/blob/a781bd5f580bdb2461a9b8769785895e1ffb6e6b/contracts/bank_jetton.tact#L23   )
and  
`        let senderWallet: StateInit = initOf BankJettonWallet (stake.sender, self.BNKjetton);`
(here https://github.com/Architec-Ton/banktoken/blob/a781bd5f580bdb2461a9b8769785895e1ffb6e6b/contracts/bank_staking.tact#L36)

Arguments, sent that function is same, that was with checked lot of iterations (look to picture):
Screenshot_20240629_185148-1

Whats wrong?

### LINKED ISSUE #533: Inconsistent ABI generation for projects with multiple contracts

Gusarich:
Closes #533 and closes #516


src/types/resolveDescriptors.ts
```diff
@@ -264,6 +265,7 @@ export function resolveDescriptors(ctx: CompilerContext) {
     const types: Map<string, TypeDescription> = new Map();
     const staticFunctions: Map<string, FunctionDescription> = new Map();
     const staticConstants: Map<string, ConstantDescription> = new Map();
+    const structDependencies: Map<string, Set<TypeDescription>> = new Map();
```
anton-trunov, line 268: Can you please add a comment on what "struct dependencies" are?

Gusarich: This is a Map that stores Structs on which contracts and other Structs depend

anton-trunov: do you mean for each contract we get the set of structs that can be actually instantiated in it and not just declared?

if so, what happens if a contract does not ever call a static function that touches a struct? or what if a contract refers statically to a struct but we can never tell if it is ever going to be created at runtime?

what I'm saying is something like: do we actually need to put that struct information into the ABI file?

Gusarich: Hm, good point. There can be cases when struct is used only in runtime and never in sent/received messages - and in such cases there's no sense to put it into ABI.

I guess it'll be right to only analyse message structs used in contract either as receiver selector or in message sending.

anton-trunov: why don't we remove structs from abi altogether?

anton-trunov: > or in message sending

this might happen in unreachable code

Gusarich: > why don't we remove structs from abi altogether?

but they are part of the "interface" of the contract. this is how to interact with it 🤔

Gusarich: > > or in message sending
> 
> this might happen in unreachable code

error throwing could also happen in unreachable code

Gusarich: I believe it's better to have some struct/error in ABI even if it's not actually being used there in real scenarios, than not having them at all.

The original problem was that structs/errors that go into ABI were selected in a wrong way so that **the same contract** in different compiling projects would generate a **different** ABI.

anton-trunov:
An internal discussion with @Gusarich has lead us to the decision of not to change ABI generation.

PR #562, opened 2024-07-12
split serializers for `Cell`, `Slice` and `Builder`

### LINKED ISSUE #414: `Cell`, `Slice` and `Builder` are treated as `Cell` in generated TS wrappers
Each of the below get methods return different type (Cell, Slice and Builder):
https://github.com/tact-lang/tact/blob/7afc87e8b24618301af5fdd58dcf7862ee1d5101/src/test/e2e-emulated/contracts/serialization-3.tact#L47-L55

However, after compilation, generated TS wrapper treats all of these values as Cells:
```ts
async getGetC(provider: ContractProvider) {
    let builder = new TupleBuilder();
    let source = (await provider.get('getC', builder.build())).stack;
    let result = source.readCell();
    return result;
}

async getGetD(provider: ContractProvider) {
    let builder = new TupleBuilder();
    let source = (await provider.get('getD', builder.build())).stack;
    let result = source.readCell();
    return result;
}

async getGetE(provider: ContractProvider) {
    let builder = new TupleBuilder();
    let source = (await provider.get('getE', builder.build())).stack;
    let result = source.readCell();
    return result;
}
```

Both `Slice` and `Builder` have their own types in TS so they should be used instead.

Gusarich:
Closes #414 


src/bindings/typescript/serializers.ts
```diff
@@ -279,49 +279,63 @@ const addressSerializer: Serializer<{ optional: boolean }> = {
...
         } else {
-            return "Cell";
+            return getCellLikeTsType(v.kind);
         }
     },
     tsLoad(v, slice, field, w) {
         if (v.optional) {
             w.append(
-                `let ${field} = ${slice}.loadBit() ? ${slice}.loadRef() : null;`,
+                `let ${field} = ${slice}.loadBit() ? ${slice}.loadRef()${v.kind !== "cell" ? ".as" + getCellLikeTsType(v.kind) + "()" : ""} : null;`,
```
anton-trunov: let's factor out this into a helper producing `asSlice`, `asBuilder`, ...
and also apply it for the cases below

src/test/e2e-emulated/masterchain.spec.ts
```diff
@@ -122,7 +122,9 @@ describe("masterchain", () => {
         await system.run();
         const addr = new Address(-1, Buffer.alloc(32, 0));
         void expect(
-            contract.getParseAddress(beginCell().storeAddress(addr).endCell()),
+            contract.getParseAddress(
```
anton-trunov, line 125: Hmmm, looks like with this change the amount of boilerplate for tests has increased a bit. Can we do something about it? I'd expect that with more precise types it should become easier for the end-user to write tests

anton-trunov (changes requested):

src/test/e2e-emulated/masterchain.spec.ts
Gusarich: > I'd expect that with more precise types it should become easier for the end-user to write tests

In general - yes. And not only tests but also custom code that work with contracts.
But in some cases (such as the one from this review comment) it'll require *more* actions, like transform `Cell` into a `Slice`.

I don't think we can do anything about such cases.

anton-trunov: ok, fair enough

anton-trunov (dismissed):

anton-trunov (approved):

PR #565, opened 2024-07-13
add impure flag to throw functions

### LINKED ISSUE #564: False positive !! check
in the next contract if the sender's address is not in the `self.book` as key the contract should throw `128: Null reference exception`, but it doesn't:
```
contract ContactBook with Ownable  {

    owner: Address;
    book: map<Address, Int as uint16>;
    lenght: Int as uint8;
    
    init(owner: Address, book: map<Address, Int as uint16>, lenght: Int) {
        self.owner = owner;
        self.book = book;
        self.lenght = lenght;
    }

    receive(str: String) {
        let sender: Address = context().sender;
        let stat: Int = self.book.get(sender)!!;
        self.forward(self.owner, str.asComment(), false, null);
    }
}
```
Possible solution: add `impure` flag to the function
```
forall X -> X __tact_not_null(X x) inline {
    throw_if(128, null?(x)); return x;
}
```

apasittacc:
Closes #564


Gusarich:
`impure` modifier tells the compiler not to remove the function call even if it thinks the returned result is never used.
These functions like `pow` or `log2` don't really have to be always executed. If the result is really never being used for some reason in compiled code - so be it.

src/generator/writers/writeStdlib.ts
```diff
@@ -1915,12 +1919,15 @@ export function writeStdlib(ctx: WriterContext) {
...
         ctx.signature(`int __tact_log2(int num)`);
+        ctx.flag("impure");
         ctx.context("stdlib");
         ctx.asm(`asm "DUP 5 THROWIFNOT UBITSIZE DEC"`);
     });
 
     ctx.fun(`__tact_log`, () => {
         ctx.signature(`int __tact_log(int num, int base)`);
+        ctx.flag("impure");
+        ctx.flag("inline");
```
Gusarich, line 1926: and if we want to add `inline` modifiers I think we should run some benchmarks and compare the gas usage with `inline_ref`.

Gusarich (changes requested):
I suggest only leaving the `impure` specifier for `__tact_verify_address` and `__tact_not_null` because it makes sense in context of the original issue.

Gusarich (dismissed):

src/generator/writers/writeStdlib.ts
Gusarich: `inline_ref` isn't used widely in Tact stdlib so I guess we can merge it with this `inline` modifier and maybe do a benchmarking on most of the stdlib functions later

anton-trunov:
@Gusarich let's add the test case from the linked issue and an entry to `CHANGELOG.md` and we are good to go

src/test/e2e-emulated/contracts/maps.tact
```diff
@@ -945,4 +945,9 @@ contract MapTestContract {
     get fun intMap1Del(key: Int): Bool {
         return self.intMap1.del(key);
     }
+
+    get fun checkNullReference(): Int {
+        let m: map<Int, Int> = emptyMap();
+        return m.get(0)!!;
```
anton-trunov, line 953: I just checked with the online IDE and it actually throws exit code 128 in this getter. The IDE runs Tact v1.4.0. This needs to be a receiver, because getters are not supposed to change the contract's state, so it looks like the semantics of `impure` is a bit different in the FunC compiler for these two cases.

anton-trunov (changes requested):

anton-trunov (approved):

PR #567, opened 2024-07-13
Add `process.setMaxListeners(0);` in the beginning of the `main.ts`

### LINKED ISSUE #156: `MaxListenersExceededWarning` when running tests
```
💼 Compiling project wallet-opt...
   > Wallet: tact compiler
   > Wallet: func compiler
(node:513) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 uncaughtException listeners added to [process]. Use emitter.setMaxListeners() to increase limit
(Use `node --trace-warnings ...` to show where the warning was created)
```

In `src/func/funcfiftlib.js` there is the following code for some reason

```
process["on"]("unhandledRejection",function(reason){throw reason});
```

It's executed once per every created `CompilerModule`

https://github.com/tact-lang/tact/blob/e4661498a05fb936dce67af7363290c35a5e14a1/src/func/funcCompile.ts#L85-L91

When number of subscriptions to `unhandledRejection` goes over preset Node limit of 10, Node emits error message, because in usual conditions numerous subscriptions without unsubscriptions mean that there's a memory leak.

We can use `process.setMaxListeners(###)` to increase the limit, but the problem appears to be already fixed upstream [here](https://github.com/ton-community/func-js/commit/159e63aa0ecb0c60fb4f97df313c1c75d9f4c87a) and is not present in the [current version](https://github.com/ton-community/func-js-bin/blob/main/src/funcfiftlib.js).

Gusarich:
Closes #156 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Display a clearer error in case the source code file is missing when using Tact's CLI: PR [#561](https://github.com/tact-lang/tact/pull/561)
 - Error messages for unicode code points outside of valid range: PR [#535](https://github.com/tact-lang/tact/pull/535)
 - Correct regex for unicode code points and escaping of control codes in generated comments: PR [#535](https://github.com/tact-lang/tact/pull/535)
+- The `MaxListenersExceededWarning` warning is resolved: [#567](https://github.com/tact-lang/tact/pull/567)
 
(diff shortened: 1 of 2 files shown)


anton-trunov:
> By default EventEmitters will print a warning if more than 10 listeners are added for a particular event. This is a useful default that helps finding memory leaks. 

I remember @byakuren-hijiri telling me there is a memory leak in the Tact compiler fuzzer, so the fuzzer stops working in about 10 minutes. This might be related and so we shouldn't silence the warning but fix the actual problem

PR #568, opened 2024-07-13
Enhance error message when trying to access storage var without `.self` prefix

### LINKED ISSUE #261: Improve error message when the user forgets to use `self` to refer to storage vars
In case one drops `self.` when referring to the `var` storage variable, we should suggest to use `self`.

Gusarich:
Closes #261 


Gusarich:
@anton-trunov what if we try to find the variable name in all of the structs in the current context, not only `self`?

anton-trunov:
Let's add similar tests for traits

anton-trunov:
> what if we try to find the variable name in all of the structs in the current context, not only self?

hmmm, not sure about this one, because it might return several results (different struct types can have identical field names)

Gusarich:
> Let's add similar tests for traits

done

src/types/stmts-failed/var-does-not-exist3.tact
```diff
@@ -0,0 +1,19 @@
+primitive Int;
+
+trait BaseTrait { }
+
+trait TestTrait {
+    someField: Int;
+}
+
+contract Test with TestTrait {
```
anton-trunov, line 9: I actually meant something like this:

```
trait Ownable {
    owner: Address;

    fun requireOwner() {
        nativeThrowUnless(132, sender() == owner);
    }
}
```

anton-trunov (approved):

PR #571, opened 2024-07-14
Forbid comparing null with non-optional type

### LINKED ISSUE #570: Checking for null a struct throw FunC compilation error
The compilation of the following code:

```
struct A {
    a: Int as uint256;
    b: Int as uint256;
}

contract SampleTactContract {
    tuple: A;

    init() { self.tuple = A{a:0,b:0}; }

    // removing `get` keyword here fixes the error
    get fun check3(): Int {
        return self.check() + self.check2();
    }

    fun check(): Int {
        if (self.tuple != null) {
            return self.tuple.a;
        } else {
            return 0;
        }
    }

    fun check2(): Int {
        if (self.tuple != null) {
            return self.tuple.b;
        } else {
            return 0;
        }
    }
}
```

fails with:

```
sample_SampleTactContract.code.fc:21:49: error: cannot apply function null? : Forall (_Z) _Z -> int to arguments of type (int, int): cannot unify type ??65 with (int, int): cannot unify generic type and tensor
      if ((~ null?(($self'tuple'a, $self'tuple'b)))) {
```

Gusarich:
Closes #570


Gusarich:
Seems that `__tact_load_address_opt` is being used for optional addresses even though the type of the field might be not nullable

Gusarich:
It seems that currently `Address` type can be `null` by default even without marking it with `?` sign and simply refers to `addr_none`:
https://github.com/tact-lang/tact/blob/4ce02f24020035cb5bb1c06b2cdb98f116d27663/src/test/e2e-emulated/contracts/sample-jetton.tact#L69

I believe that this is wrong behavior and `null` should only be allowed in nullable types marked with `?`, e.g. `Address?`.

We should probably move this one to 1.5.0 because resolving this will require adding or changing some functionality related to work with addresses.

anton-trunov:
> We should probably move this one to 1.5.0 because resolving this will require adding or changing some functionality related to work with addresses.

It's ok to change things if this fixes a bug

Gusarich:
Actually, after investigating this issue a bit further I understood that there are no *bugs* in optional addresses and `__tact_load_address_opt` is only used when the address is optional.

The reason for test fail is that they were originally wrong - `response_destination` on the line I mentioned above is always not null but there is still a comparison.

So to finish with this PR and #570 we can just fix the `SampleJetton` test case, and get back to https://github.com/tact-lang/tact/issues/515#issuecomment-2205178642 in the next release.

anton-trunov:
> So to finish with this PR and https://github.com/tact-lang/tact/issues/570 we can just fix the SampleJetton test case, and get back to https://github.com/tact-lang/tact/issues/515#issuecomment-2205178642 in the next release.

let's go for it

anton-trunov:
@Gusarich I fixed some edge cases with `void` types and refactored the implementation a bit, please have a look. Maybe you can break it or add new test cases

Gusarich:
> @Gusarich I fixed some edge cases with `void` types and refactored the implementation a bit, please have a look. Maybe you can break it or add new test cases

Looks good!

PR #577, opened 2024-07-15
feat: Add init parameters to ABI

### LINKED ISSUE #576: ABI for init function in build output

rahulyadav-57:
Closes #576 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `-e` / `--eval` CLI flags to evaluate constant Tact expressions: PR [#462](https://github.com/tact-lang/tact/pull/462)
 - `-q` / `--quiet` CLI flags to suppress compiler log output: PR [#509](https://github.com/tact-lang/tact/pull/509)
 - Markdown report for compiled contracts now includes Mermaid diagrams for trait inheritance and contract dependencies: PR [#560](https://github.com/tact-lang/tact/pull/560)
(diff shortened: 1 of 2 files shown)


anton-trunov:
So, here is the definition of the `ContractABI` type:

```ts
export type ContractABI = {
    name?: Maybe<string>;
    types?: Maybe<ABIType[]>;
    errors?: Maybe<{
        [key: number]: ABIError;
    }>;
    getters?: Maybe<ABIGetter[]>;
    receivers?: Maybe<ABIReceiver[]>;
};
```

The proposed solution does not conform to this type.

There is no `init` in `ContractABI`. And `init` is essentially included in each and every receiver: there is a initialization flag that is checked before executing any receiver. That said, we should either propose to extend the `ContractABI` type in `@ton/core` or go with some other solution. @Gusarich might have some more thoughts on this.

rahulyadav-57:
@Gusarich, what is your opinion on this?

Gusarich:
Well, Anton already explained why we can't just do what you did here.
And I personally don't see any point in adding `init` to ABI.

rahulyadav-57:
I agree, but it might be necessary for third-party tools like Nujan, where input fields are rendered based on the initialization parameters for deployment. While I can obtain this information via the compiler context, it would be helpful for projects that do not have access to the context.

anton-trunov:
We can provide API using a different mechanism, let's discuss this elsewhere

anton-trunov:
@Gusarich btw, the `interfaces` property in the contract ABI object that gets returned from the `createABI` function is also non-standard

PR #584, opened 2024-07-17
Second version of partial evaluator

jeshecdom:
Towards #435 

Second version of partial evaluator.

- Supports further combinations of +/-, specially when - is on the right, i.e., x + (y - z).
- Supports all the following algebraic rules (in file src/optimizer/algebraic.ts):
   - `x + 0 ---> x`
   - `x - 0 ---> x`
   - `0 - x ---> -x`
   - `x * 0 ---> 0`  (`x` an identifier)
   - `x * 1 ---> x`
   - `x - x ---> 0`  (`x` an identifier)
   - `x + x ---> x * 2`

- Next version:
   - Support for integer operators: `/, %, >>, <<`.
   - Boolean expressions
   
 ---


anton-trunov:
> `x + x ---> x * 2`

I'm wondering about the reasoning behind this rule

anton-trunov (approved):

src/grammar/test/partial-eval.spec.ts
anton-trunov, line 1: btw, let's move this file somewhere else: partial evaluator tests should not live in the grammar subdirectory

jeshecdom: Ok. I moved it into a folder Test inside src/optimizer.

jeshecdom:
> > `x + x ---> x * 2`
> 
> I'm wondering about the reasoning behind this rule

So far, I've noticed two rules of thumb:
1. Agglomerate constants as much as possible and put them in one branch of the tree.
For example, `(X + 3) + 5`: here, we should combine 3 and 5 and move them in a branch of the + operator.
2. Exchange variables for constants as much as possible so that other rules will be able to consume the constants.
For example, in `(X + X) * 3`, we should exchange one of the `X` for `2`, so that `(X * 2) * 3`. Then, other rules will combine the constants `2` and `3`.
Notice that `X + X` and `X * 2` are equivalent and have the same overflow behavior.

These rules of thumb combine with the way the optimizer works: it is bottom up and greedy. Each pass of the rules is linear on the size of the tree, but the optimizer could execute several passes of the rules. To give an example, consider the expression `(X + X) * 3`.
First, the optimizer will attempt to simplify the leaves `X`, `X`, and `3` (which are already simplified). Then, it will attempt to simplify `X + X`, so it will replace it with `X * 2` after a pass of the rules on the expression `X + X`. Then, it will give a pass on the tree `X * 2` because there is an opportunity of further simplification there (just in case `X` is actually another expression, for example `X` could be `Y * 4`). And finally, it will execute a pass on the expression `(X * 2) * 3`, which will replace it by `X * (2 * 3)` using the associativity rules, giving `X * 6`.

I am aware that the pattern `X + X` is not very common. Although, I've seen assignments inside loops of the form `idx = f(i,i)`, where `i` is the loop index and `f` could be `+`. Also, I will have to add rules like `x + c*x ---> x * (1 + c)` and `c1 * x + c2 * x ---> x * (c1 + c2)` which are more of a "distributive" kind (with some safety conditions on the constants `c`, `c1`, and `c2` of course).

PR #559, opened 2024-07-12
New Func backend: First iteration

byakuren-hijiri:
Closes #526
Closes #619 (FunC grammar & parser by @novusnota)

### Main roadmap
1. ~~Write AST/formatter, and utilities to generate and emit Func code.~~
3. Rewrite the existing codegen using the API created on the step 1.
   It should be rewritten entirely at this step, since everything is interconnected in the old backend, and there are no any unit tests checking any specific functionality of the generated code.
4. Provide differential testing between two backends achieving good code coverage for the new backend.
    The previous backend is basically untested. It requires to write a number of Tact contracts and setup code coverage to ensure that all the implemented features of the new backend work in the same way as in the old one.
5. Refactor the new backend and remove the old backend.

### Sub-roadmap of @novusnota

The goal is to create a comprehensive FunC grammar and parsed using the Ohm parsing toolkit. Should cover FunC version 0.4.4:
- [x] `#pragma ...`
  - [x] Grammar (`grammar.ohm`)
  - [x] Semantic actions (`semantics.addOperation`)
  - [x] Unit tests (positive and negative)
- [x] `#include ...`
  - [x] Grammar
  - [x] Semantic actions
  - [x] Unit tests
- [x] `global ...`
  - [x] Grammar
  - [x] Semantic actions
  - [x] Unit tests
- [x] `const ...`
  - [x] Grammar
  - [x] Semantic actions
  - [x] Unit tests
- [x] Identifiers & Types (unexpectedly tough!)
  - [x] Grammar
  - [x] Semantic actions
  - [x] Unit tests
- [x] Comments and literals
  - [x] Grammar
  - [x] Semantic actions
  - [x] Unit tests
- [x] Functions (declarations & definitions, including asm ones):
  - [x] Grammar
  - [x] Semantic actions
  - [x] Unit tests
- [x] Statements
  - [x] Grammar
  - [x] Semantic actions
  - [x] Unit tests
- [x] Expressions (just like in C++'s parser, with all its precedences matched!)
  - [x] Grammar
  - [x] Semantic actions
  - [x] Unit tests
- [x] Done?
- [x] (maybe) Add tests with .fc files from `ton-blockchain/ton` monorepo

### Current progress
* [x] FunC syntax tree (AST + comments)
* [x] FunC syntax iterators
* [ ] FunC formatter (we can skip unit tests for now, as it will be tested as a part of the Tact compiler pipeline)
  + [ ] ~~Support JSON dumps for debugging purposes~~
[...]

PR #604, opened 2024-07-22
Fix FunC code generated for `recv_external`

### LINKED ISSUE #610: Externals do not throw exit code `130` if incoming message is unhandled

byakuren-hijiri:
Closes #610.

The external receiver generates the following code:

```
;; Handle operation
int handled = ...;
;; Throw if not handled
throw_unless(handled, 130);
```

Which uses the incorrect order of argument in `throw_unless`.


anton-trunov:
it seems to me `throw_unless(<exit-code>, <condition>)` is the right order

byakuren-hijiri:
> it seems to me throw_unless(<exit-code>, <condition>) is the right order

That's right. I noticed an inconsistency in the generated code while working on https://github.com/tact-lang/tact/pull/559 and just got around to checking it.

The problem is in `recv_external` which generates `throw_unless(handled, 130)`.

anton-trunov (approved):
Approving without a test, because it appears to be hard to test it: the generated TS wrappers do not allow to send an unhandled message at the level of the typechecker and also there is a runtime check for incorrect messages at the TS level (which is pretty weird actually)

byakuren-hijiri:
> Approving without a test, because it appears to be hard to test it: the generated TS wrappers do not allow to send an unhandled message at the level of the typechecker and also there is a runtime check for incorrect messages at the TS level (which is pretty weird actually)

Agreed.

I'm going to test the generated code for this case as well as for other receivers in https://github.com/tact-lang/tact/pull/559.

PR #611, opened 2024-07-23
fix: safety condition fixed in associative rule

### LINKED ISSUE #599: Partial evaluator simplifies some expressions when it should not
There are some edge cases that the Associative3 Rule currently does not consider in its safety conditions, causing it to simplify expressions that should not be simplified.

Examples:
- `(0 - X) + -1` simplifies to `-1 - X`. The problem occurs when `X = MIN`, because `0 - X` overflows, but `-1 - X = MAX` does not.
- `(0 - X) - 1` simplifies to `-1 - X`. This has the same problem as the previous one.
- `(X * 2) * -1` simplifies to `X * -2`. The problem occurs when `X = -(MIN/2) = 2^255`, because `X * 2 = 2^256` overflows, but `X * -2 = -2^256` does not.

The first two examples are caused because the safety condition in the associativity rule does not check when the constant `c` in `(c - X)` is zero. The third example is caused because the safety condition currently has a weak inequality `|c1| <= |c1 * c2|` in the expression `(X * c1) * c2`, when it should be a strict inequality `|c1| < |c1 * c2|` for `c1 > 0` and `c1 * c2 < 0`.

jeshecdom:
Closes #599.

### DIFF
--- a/src/optimizer/test/partial-eval.spec.ts
+++ b/src/optimizer/test/partial-eval.spec.ts
@@ -151,11 +151,8 @@ const mixedExpressions = [
     { original: "(X * -1 * -2) + (X * -1 * -2)", simplified: "X * 4" },
 ];
 
-const previousFailingAssoc3Expressions = [
-    // Previously, the following cases where wrongly simplified
-    // by the partial evaluator. Now, they do not simplify,
-    // as it should be.
-
+// These expressions are edge cases to test the associative rule only.
+const associativeRuleExpressions = [
     // The following three cases should NOT simplify to
     // -1 - X because X could be MIN, so that 0 - X overflows
     // but -1 - X = MAX does not.
@@ -316,7 +313,7 @@ describe("partial-evaluator", () => {
         new AssociativeRule3(),
     ]);
 
-    previousFailingAssoc3Expressions.forEach((test) => {
+    associativeRuleExpressions.forEach((test) => {
         testExpressionWithOptimizer(test.original, test.simplified, optimizer);
     });
 });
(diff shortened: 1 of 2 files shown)


src/optimizer/test/partial-eval.spec.ts
```diff
@@ -151,42 +151,43 @@ const mixedExpressions = [
...
-    { original: "(2 * X) * -1", simplified: "-2 * X" },
+    //{ original: "(2 * X) * -1", simplified: "-2 * X" },
 
     // PROBLEM: Same as before
-    { original: "-1 * (X * 2)", simplified: "X * -2" },
+    //{ original: "-1 * (X * 2)", simplified: "X * -2" },
 
     // PROBLEM: Same as before
-    { original: "-1 * (2 * X)", simplified: "-2 * X" },
+    //{ original: "-1 * (2 * X)", simplified: "-2 * X" },
```
anton-trunov: looks like the commented out code should be removed

anton-trunov (changes requested):

src/optimizer/test/partial-eval.spec.ts
```diff
@@ -151,42 +151,26 @@ const mixedExpressions = [
...
-    { original: "(0 - X) - 1", simplified: "-1 - X" },
-
-    // The following case four cases forgot to check that
-    // c2 * c1 < 0 when c1 > 0 in (X * c1) * c2.
-    // The safety condition requires the strict inequality
-    // |c1| < |c1 * c2| for the case c1 > 0 and c1 * c2 < 0,
-    // while currently has |c1| <= |c1 * c2|
-
-    // PROBLEM: X could be -(MIN/2) = 2^255, so that
+const previousFailingAssoc3Expressions = [
```
anton-trunov: `previousFailingAssoc3Expressions` won't make sense after some time, the name should be descriptive outside the context of the development timeline

anton-trunov (approved):

PR #622, opened 2024-07-26
Improved error messages on missing field/method

### LINKED ISSUE #502: Improving error message on missing field/method
We could implement additional logic before raising an error on a missing field to check if there is a method with the same name.

For example:
```
fun processSlice(s: Slice) {
  let x: Int = s.bits;
}
```

Currently the compiler prints: `Type "Slice" does not have a field named "bits"`.

But there is a method `Slice.bits()`. It is trivial to check if there is a method with such a name and add additional context to the error message, e.g., `Type "Slice" does not have a field named "bits". Did you mean ".bits()"?`.

We could also do the same for missing methods and check if there is a field with that name.

Here are the relevant parts in the source code:
* [missing methods](https://github.com/tact-lang/tact/blob/97ab73df3e93068dfdf2bd3eb883dac80f42fdb0/src/types/resolveExpression.ts#L556)
* [missing fields](https://github.com/tact-lang/tact/blob/97ab73df3e93068dfdf2bd3eb883dac80f42fdb0/src/types/resolveExpression.ts#L419)

Gusarich:
Closes #502 


src/types/resolveExpression.ts
```diff
@@ -544,32 +556,41 @@ function resolveCall(
...
+                }
+            }
+
+            return registerExpType(ctx, exp, f.returns);
         }
 
-        // Check arguments
-        if (f.params.length !== exp.args.length) {
+        // Check if a field with the same name exists
+        const field = srcT.fields.find((v) => eqNames(v.name, exp.method));
```
byakuren-hijiri, line 586: Just to confirm, `srcT.field` contain fields inherited from parent traits, right?

byakuren-hijiri (approved):

anton-trunov (approved):
Great stuff!

PR #630, opened 2024-07-27
Implement `Struct.toSlice()`

### LINKED ISSUE #474: Add `.toSlice()` extension function for Structs and Messages
One can do `Struct.toCell().asSlice()` already, but for consistency reasons we probably should have `.toSlice()` too. For example, `StringBuilder` has it, and it would also mirror the `.fromSlice()` method nicely. 

As mentioned in: https://github.com/tact-lang/tact-docs/pull/271#discussion_r1651512617

Gusarich:
Closes #474 


src/abi/struct.ts
```diff
@@ -103,6 +103,51 @@ export const StructFunctions: Map<string, AbiFunction> = new Map([
...
+                if (args.length !== 1) {
+                    throwCompilationError(
+                        "toSlice() expects no arguments",
+                        ref,
+                    );
+                }
+                const arg = args[0]!;
+                if (arg.kind !== "ref") {
+                    throwCompilationError(
+                        "toSlice() is implemented only a struct type",
```
anton-trunov: We also have `toSlice()` for string builders: https://github.com/tact-lang/tact/blob/5bec4864deccf8a3ee694914f3be0a815818488e/stdlib/std/text.tact#L30

Gusarich: Then should I also change the `toCell() is implemented only a struct type` error message for `toCell()` as there's also that function in stdlib for `StringBuilder`?

anton-trunov: Sure, let's add it here too

anton-trunov:
Let's update this and merge

src/abi/struct.ts
Gusarich: Maybe it'll be better to change these error messages to something like `toSlice() is not implemented for type 'type'`?

anton-trunov: Sure, sounds good

anton-trunov (approved):

PR #636, opened 2024-07-28
fix: improve `funcId` recognition and squash misdetection bugs

### LINKED ISSUE #635: `funcPlainId` doesn't allow identifiers having only `[a-fA-F0-9]`, i.e. having characters only from allowed set for hexadecimals
For example, the following native function won't parse:

```tact
@name(C4)
native C4onPrevLineWontParse();
```

But this one will parse just fine:

```tact
@name(C4g)
native works();
```

This is (fortunately) not a huge problem, since there aren't that many readable function names with only `abcdef` letters in them. But it's still a big whoopsie of mine, which I only discovered during extensive work on the [FunC parser](https://github.com/tact-lang/tact/issues/619) :/

I think we should rollout the fix early, in some kind of hotfix 1.4.2 release.

novusnota:
That took a looooooooot of RegEx iterations.

Closes #635

~~- [ ] I have documented my contribution in Tact Docs: https://github.com/tact-lang/tact-docs/pull/PR-NUMBER~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- FunC function identifiers with characters from hexadecimal set: PR [#635](https://github.com/tact-lang/tact/pull/635)
+- FunC function identifiers with characters from hexadecimal set: PR [#636](https://github.com/tact-lang/tact/pull/636)
 
 ## [1.4.1] - 2024-07-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- FunC function identifiers with characters from hexadecimal set: PR [#635](https://github.com/tact-lang/tact/pull/635)
+
 ## [1.4.1] - 2024-07-26
 
 ### Added
--- a/src/grammar/grammar.ohm
+++ b/src/grammar/grammar.ohm
@@ -284,7 +284,7 @@ Tact {
                   | ("-"? digit+) ")" --notDecimalNumber
                   | ("-"? "0x" hexDigit+) ")" --notHexadecimalNumber
 
-    funcPlainId = ~funcInvalidId (~(whiteSpace | "(" | ")" | "," | "." | ";" | "~") any)+
+    funcPlainId = ~funcInvalidId (~(whiteSpace | "(" | ")" | "[" | "]" | "," | "." | ";" | "~") any)+
 
     funcQuotedId = "`" (~("`" | "\n") any)+ "`"
 
--- a/src/grammar/grammar.ohm
+++ b/src/grammar/grammar.ohm
@@ -263,12 +263,12 @@ Tact {
     id = ~reservedWord #idStart #(idPart*)
 
     // FunC identifiers, where `funcId` stands for FunC function identifier
-    // The plain identifier cannot be a single underscore nor just a number
-    funcPlainId = "_"? "-"? "0x"? hexDigit* ~hexDigit (~(whiteSpace | "(" | ")" | "," | "." | ";" | "~") any)+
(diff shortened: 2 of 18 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- FunC function identifiers with characters from hexadecimal set: PR [#635](https://github.com/tact-lang/tact/pull/635)
```
```suggestion
- FunC function identifiers with characters from hexadecimal set: PR [#636](https://github.com/tact-lang/tact/pull/636)
```

src/grammar/test/items-native-fun-funcid.tact
```diff
@@ -51,3 +51,38 @@ native idTest17();
...
+// I.e., you can have a function with any of those combinations of characters, but only one.
+
+@name(_0x0)
+native idTest22();
+
+@name(_0)
+native idTest23();
+
+// NOTE(novusnota):
+// There is a only small set of cases related to the above that our parser won't parse, but C++ FunC's will:
```
anton-trunov: looks like this note should go to grammar.ohm

anton-trunov (changes requested):

anton-trunov:
To be on the safe side, let's take `stdlib.fc` and `mathlib.fc` as libraries that are commonly used and see if we can add more function identifiers from those libs into our positive tests to make sure people will be able to FFI with those.

src/grammar/grammar.ohm
```diff
@@ -263,12 +263,22 @@ Tact {
...
+                | "-"? digit+ "_" --justDecUnderscore
 
     funcQuotedId = "`" (~("`" | "\n") any)+ "`"
 
-    funcId = ~("\"" | "{-") ("." | "~")? (funcQuotedId | funcPlainId)
+    funcId = ~("\"" | "{-" | "#") ("." | "~")? (funcQuotedId | funcPlainId)
+
+    /*
+      NOTE(novusnota):
+
```
```suggestion
```

anton-trunov: Let's remove this part, there is commit history to query authorship and additional context

anton-trunov (changes requested):
This PR accepts, for instance, `+` as a valid FunC identifier, but it's actually not.

Something like this should work:
```
    funcInvalidId = "_" ")" --notUnderscore
                  | ("+" | "-" | "*" | "/" | "~/" | "^/" | "%" | "~%" | "^%" | "/%") ")" --notArithOperator
      // more operators here
                  | ("-"? digit+) ")" --notDecimalNumber
                  | ("-"? "0x" hexDigit+) ")" --notHexadecimalNumber
    funcPlainId = ~funcInvalidId (~(whiteSpace | "(" | ")" | "," | "." | ";" | "~") any)+
```

novusnota:
Nice observation! I'd try putting lookaheads `&` in front of those ")" so they don't get parsed, but let me try your suggestion first, it should work. UPD: Yeah, it does, clever idea!

The keywords are not allowed and should be excluded too. Basically, contents of https://github.com/ton-blockchain/ton/blob/master/crypto/func/keywords.cpp has to be put in the exceptions list. I'm on it.

UPD2: Order of operators in those inner () alternations really does matter, the ones with bigger prefix shall stand in front of others.

novusnota:
<details>
<summary>There was a suggestion, which didn't pass negative tests, so I hid it.</summary>

The thing we've engineered works, but _possibly_ I've found a cleaner solution, which doesn't require the ")" and can be used in the FunC parser too (the current approach in this PR fails there in absence of ")" and stuff). But this one works:

image

It passes similar set of tests we have for function identifiers, but adapted to FunC code:

image

UPD: Negative tests failed terribly, reverting :)

</details>

anton-trunov:
@novusnota So, what's the current status of this PR?

novusnota:
@anton-trunov It's a working and robust solution. The alternative would be to do invalidation of identifiers during syntactic analysis, which enhances the error messages further, but that seems to be a bit out of scope of this PR. Wdyt?

anton-trunov (approved):

PR #637, opened 2024-07-29
feat: `map.deepEquals` for deep map comparison

### LINKED ISSUE #196: Map comparison can produce false-negative results
I've noticed that Tact allows to compare Maps with `==` and `!=` operators while working on #195:
https://github.com/tact-lang/tact/blob/7ef2ae17baef890d6489688d523bcdf77ce7b561/src/generator/writers/writeExpression.ts#L288-L293
If we look at the implementation of `__tact_cell_eq_nullable` (same for `neq` operation) we can see how exactly it compares maps in this case:
https://github.com/tact-lang/tact/blob/7ef2ae17baef890d6489688d523bcdf77ce7b561/src/generator/writers/writeStdlib.ts#L941-L952

As you can see, it simply compares their hashes. This is fine for `Cell` comparisons, but not for `Map` because the same dictionaries in TVM can have many representations.
In most cases it'll still work, but that's just because most serializers are written identically to the one from TON monorepo. It's still possible to break the comparison to produce **false-negative** results by serializing a map manually or by changing the logic of serializer in some library.

We should either remove the ability to compare maps or highlight that behaviour for non-empty maps.

Gusarich:
Closes #196


anton-trunov:
Let's also add a negative test case based on https://github.com/tact-lang/tact/issues/196#issuecomment-2075088934. It should show that `deepEquals` works fine on it, whereas `==` fails.

Gusarich:
We should also note in documentation that this function is very expensive in terms of gas (iterating a map + using `slice_hash`, both are expensive things), and in most cases it'll be enough to use hash comparison via `==`.

anton-trunov (approved):
Awesome work!

PR #640, opened 2024-07-31
Compile time optimization - send()

cryshado:
At the moment I've noticed that the [`send()`](https://github.com/tact-lang/tact/blob/116a135514aee6d9b9662531ec1184548d7bdd78/stdlib/std/send.tact#L21-L70) function implemented in the Tact standard library works rather sub-optimally, making unnecessary "writes" to the cell builder and performing a number of runtime checks. I believe that this can be fixed by implementing this function natively in the compiler and applying some set of optimizations. The current changes presented in this PR helps to reduce the gas cost of the example contract by _18.43%_.

Group 1

Group 2

### DIFF
--- a/src/abi/global.ts
+++ b/src/abi/global.ts
@@ -5,7 +5,7 @@ import {
     writeExpression,
     writeValue,
 } from "../generator/writers/writeExpression";
-import { throwCompilationError } from "../errors";
+import { throwCompilationError, throwInternalCompilerError } from "../errors";
 import { evalConstantExpression } from "../constEval";
 import { getErrorId } from "../types/resolveErrors";
 import { AbiFunction } from "./AbiFunction";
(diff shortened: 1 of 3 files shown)


anton-trunov:
Ok, I see the tests are failing. This optimization looks great, but it needs to be backwards compatible: we don't break our users' code, unless they rely on some buggy behavior that needs to be fixed. Also, I'm personally not a fan of a having a function with 7 *positional* arguments, I still prefer to have named arguments, so a struct is fine currently.

Gusarich:
I think that such compile-time optimizations will be done automatically for this and many other functions once we have an evaluator that handles cell building.

cryshado:
@anton-trunov 
> Also, I'm personally not a fan of a having a function with 7 _positional_ arguments

I completely agree, but the only way not to accept positional arguments and not to accept optional arguments (to avoid runtime checks) would be to overload structure initializers / to allow anonymous structure Initialization and overload the function itself, which doesn't seem to be implemented yet.

cryshado:
@Gusarich 
> once we have an evaluator that handles cell building

Yes, but I don't think the problem is only with the cell building, rather it makes sense to have another intermediate step before translating to FunC, i.e. building a some sort of FunC AST and then performing a number of optimizations on it.

anton-trunov:
@cryshado We could just process `send(SendParameters { ... })` as a special case in the compiler

PR #647, opened 2024-08-01
chore(build): Separate Feature Enabling Function

### LINKED ISSUE #646: Refactor `build` to Separate Feature Enabling Function
We need a separate function that calls `featureEnable` for each configuration option defined here:

https://github.com/tact-lang/tact/blob/2524aae05b558deaab3ce01086aefc21d65373c8/src/pipeline/build.ts#L44-L69

This will enable third-party tools to use configuration options without copy-pasting a part of the `build` function (see related issue: https://github.com/nowarp/misti/issues/48).

byakuren-hijiri:
Closes #646

### DIFF
--- a/src/pipeline/build.ts
+++ b/src/pipeline/build.ts
@@ -22,6 +22,30 @@ import { getCompilerVersion } from "./version";
 import { idText } from "../grammar/ast";
 import { TactErrorCollection } from "../errors";
 
+export async function enableFeatures(
+    ctx: CompilerContext,
+    logger: Logger,
+    config: ConfigProject,
+): Promise<void> {
+    if (config.options === undefined) {
+        return;
+    }
+    const features = [
+        { option: config.options.debug, name: "debug" },
+        { option: config.options.masterchain, name: "masterchain" },
+        { option: config.options.external, name: "external" },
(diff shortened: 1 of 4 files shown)


src/pipeline/build.ts
```diff
@@ -41,32 +65,7 @@ export async function build(args: {
         entrypoint: posixNormalize(config.path),
         options: config.options ?? {},
     });
-    if (config.options) {
-        if (config.options.debug) {
-            logger.error("   > 👀 Enabling debug");
```
byakuren-hijiri, line 46: Why do we have `logger.error` here? Let's change it to `logger.debug`?
I believe, the user doesn't need an additional confirmation that his configuration has been applied.

anton-trunov (commented):
LGTM, but one suggestion though, let's start defining(centralizing) our API in `index.ts` (see, for instance, https://github.com/ton-org/ton-core/blob/main/src/index.ts). Maybe we can export `enableFeatures` and `build` in `index.ts` to start with. If we do it and add `index.ts` to Knip's entry points, then we won't have to ignore `build.ts` (which is the right thing to do).

anton-trunov (approved):

PR #650, opened 2024-08-01
fix(typechecker): less permissive equality for optional types

### LINKED ISSUE #649: Optional types are always equality types even when the argument type is not
The following example typechecks just fine

```tact
struct S { x: Int; y: Int }

contract Test {
    get fun foo(): Bool {
        let s: S? = S {x: 42, y: 0};
        return s == s;
    }
}
```

but fails with the following error message during code generation:

```
Tact compilation failed
Error: (Internal Compiler Error) Invalid types for binary operation: /Users/anton_1/ton/tact/contract-eq-optional/test.tact:6:16
    at writeExpression (/Users/anton_1/ton/tact/dist/generator/writers/writeExpression.js:272:19)
    at writeCastedExpression (/Users/anton_1/ton/tact/dist/generator/writers/writeFunction.js:21:77)
    at writeStatement (/Users/anton_1/ton/tact/dist/generator/writers/writeFunction.js:53:32)
    at /Users/anton_1/ton/tact/dist/generator/writers/writeFunction.js:468:17
    at WriterContext.body (/Users/anton_1/ton/tact/dist/generator/Writer.js:187:9)
    at /Users/anton_1/ton/tact/dist/generator/writers/writeFunction.js:456:13
    at WriterContext.fun (/Users/anton_1/ton/tact/dist/generator/Writer.js:141:9)
    at writeFunction (/Users/anton_1/ton/tact/dist/generator/writers/writeFunction.js:447:9)
    at writeAll (/Users/anton_1/ton/tact/dist/generator/writeProgram.js:298:47)
    at writeProgram (/Users/anton_1/ton/tact/dist/generator/writeProgram.js:28:5)
```

anton-trunov:
Closes #649


Gusarich (commented):
Why do we forbid comparing `X` and `X?`? It makes sense to allow comparing these two. Generated code would look like this:
1. First we check whether `X?` is null and return `false` in this case because `null` cannot be equal to `X`
2. If `X?` is not null, we can just compare them as `X` and `X`

Same for comparison of `X?` and `X?`: First we check if either both of these are null or both of these are not null and compare them accordingly, and in other cases (first is null and second is not or vice versa) we just return `false`

anton-trunov:
> First we check whether X? is null and return false in this case because null cannot be equal to X

In general, equality comparisons only make sense for terms of the same type. The check you describe should be done by the user, e.g. `if (x != null) { return x!! == y }` where, for instance, `x: Int?` and `y: Int`.

> Same for comparison of X? and X?: First we check if either both of these are null or both of these are not null and compare them accordingly, and in other cases (first is null and second is not or vice versa) we just return false

Look at the tests: we don't forbid equality comparisons for `X?` and `X?` in general, just for non-equality types like `Builder` or structs.

anton-trunov:
> Why do we forbid comparing X and X?? It makes sense to allow comparing these two

wait, we don't actually forbid that -- we check assignability of types

so, my example above is not correct, because `Int` and `Int?` are both equality types

anton-trunov:
@Gusarich Added a test for that: https://github.com/tact-lang/tact/pull/650/files#diff-570728cab178f5e1fa5f1885b32ee9826e95de3a357d341adbc331578591ab1a

Gusarich (approved):

Issue #512, opened 2024-06-28
Basic REPL for evaluating Tact code


anton-trunov:
It should support evaluating expressions, loading modules with top-level functions and using those for evaluating Tact expressions

Issue #455, opened 2024-06-21
Ensure constant evaluator and compiler produce the same results


anton-trunov:
It would be a critical bug if the constant evaluator and the compiler produced different results for an (unevaluated) constant expression. So, we can do both at compile time and report a critical error if both results ever diverge.

@jeshecdom you might want to look into this while working on #435.

PR #664, opened 2024-08-07
First version of interpreter

jeshecdom:
Towards #512 and #455.

It supports expressions and statements. Specifically:
  + Calls to module-level functions without attributes (including recursive calls).
    Unsupported attributes: native, inline, extends, mutates
  + Reference to module-level constant declarations and declared local variables.
  + All expressions supported by the constant evaluator. In fact, I refactored the constant evaluator and moved all its code into the interpreter, since the constant evaluator is basically an expression interpreter (otherwise I was producing a lot of code duplication, and moving the code to the interpreter I have more control, instead of having the code in two separate files.).
  + All the following statements: simple and augmented assignments (but only for identifiers as path expressions), let, conditions, repeat, do-until, while, return.
    Unsupported: simple and augmened assignments for more complex path expressions, try/catch, foreach.

The interpreter assumes that before calling the interpreter, the user has executed the typechecker. The interpreter relies on CompilerContext to extract declarations.


src/test/compilation-failed/contracts/const-eval-div-by-zero-in-func.tact
anton-trunov, line 1: `func` in the file name means "function" and not "FunC", right? Let's change it to either "fun" or "function", otherwise it's confusing

src/test/compilation-failed/contracts/const-eval-div-by-zero-in-func.tact
```diff
@@ -0,0 +1,15 @@
+fun funcTest(): Int {
```
anton-trunov: let's not use "func" as an abbreviation for "function"

```suggestion
fun test(): Int {
```

src/test/e2e-emulated/contracts/constants.tact
```diff
@@ -1,5 +1,14 @@
 const someGlobalConst: Int = 100;
 
+const globalConst1: Int = 1;
+const globalConst2: Int = globalConst1 + 1;  // 2
+const globalConst3: Int = globalFunc1(globalConst1) + 1;  // 4
```
anton-trunov:
```suggestion
const globalConst3: Int = globalFun1(globalConst1) + 1;  // 4
```

src/test/e2e-emulated/contracts/constants.tact
```diff
@@ -10,6 +19,76 @@ struct T {
...
+fun globalFunc5(a: Int, b: Int): Int {
+    if (a <= b) {
+        return b;
+    } else {
+        return a;
+    }
+}
+
+// GCD for non-negative integers
+// test recursion
```
anton-trunov: let's also add the factorial function :)
would be nice to have Fibonacci as well, but the definitions are topologically sorted before codegen, so it will error out

src/test/compilation-failed/contracts/const-eval-div-by-zero-in-func.tact
jeshecdom: Oh, yes, you are right. I did not notice this. Changing it.

src/test/e2e-emulated/contracts/constants.tact
jeshecdom: I did not understand why Fibonacci would fail.

src/interpreter.ts
```diff
@@ -1,13 +1,462 @@
...
+type InterpreterConfig = {
+    // Options that tune the interpreter's behavior.
+
+    // Maximum number of iterations inside a loop before a time out is issued.
+    // This option only applies to: do...until and while loops
+    maxLoopIterations: bigint;
+};
+
+class RuntimeEnvironment {
+    private values: Map<string, Value>;
```
anton-trunov: Looks like you want a data structure which could probably be called something like "persistent multi map", i.e. a map from identifiers to stacks of values: the last key removal would actually keep the key but re-associate it with the previous value, i.e. values: `Map<string, Value[]>` and then you don't need the `enclosing` environment

anton-trunov: Or maybe not, since you are emulating the call stack frames with a linked list of environments. Please ignore the above, since it would require more bookkeeping

src/test/e2e-emulated/contracts/constants.tact
anton-trunov: Ah, sorry Fibonacci would work, I was thinking about mutually recursive even/odd functions but my fingers typed Fibonacci :)

src/interpreter.ts
jeshecdom: Yeah, I was actually thinking how to implement your suggestion, because I would also need to keep track of which variables were modified at each stack frame so that I only drop those when the frame closes.

src/interpreter.ts
```diff
@@ -485,18 +491,46 @@ export class Interpreter {
...
+        }, this);
+    }
+
+    // Closes the current environment and returns to its
+    // parent environment
+    private closeCurrentEnvironment() {
+        const parentEnv = this.currentEnv.getParentEnvironment();
+        if (parentEnv === undefined) {
+            // This is a programmer's error, not even a compilation error.
+            throw new Error(
```
anton-trunov: See PR #669 which uses "internal compiler error" instead of throwing just an `Error`

src/interpreter.ts
```diff
@@ -25,3 +480,886 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+        config: InterpreterConfig = defaultInterpreterConfig,
+    ) {
+        this.currentEnv = new RuntimeEnvironment();
+        this.initialContext = initialContext;
+        this.config = config;
+    }
+
+    // Opens a new environment with the provided param names and
+    // values.
+    private openNewEnvironment(
```
anton-trunov: I'm wondering if it would make sense to make `openNewEnvironment` and  `closeCurrentEnvironment` methods of the `RuntimeEnvironment` class

jeshecdom: Moved the opening and closing of environments into `RuntimeEnvironment` (which I renamed into `EnvironmentStack`, because that is what actually is. The stack is now a linked list of environments).
The opening and closing of environments is now done automatically inside the `EnvironmentStack` by calling `executeInNewEnvironment` method. This removes the potential error of a programmer           
forgetting to open and close environments manually. The code also looks cleaner, i.e., instead of:
```
openNewEnvironment();
try {
   // execute some code
} finally {
  closeCurrentEnvironment()
}
```
it is now:
```
envStack.executeInNewEnvironment(
 () => {
     // execute some code
  }
);
```

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+            case "struct_instance":
+                return this.interpretStructInstance(ast);
+            case "field_access":
+                return this.interpretFieldAccess(ast);
+            case "static_call":
+                return this.interpretStaticCall(ast);
+        }
+    }
+
+    public interpretName(ast: AstId): Value {
```
anton-trunov, line 761: since Tact does not allow identifier shadowing, `interpretName` should check if there are scoping rules violations (variables and constants cannot shadow each other). Yes, it could rely on the typechecker to do that, but it's not mature enough at this point, see, for instance, #676

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+            return letBinding;
+        }
+        throwNonFatalErrorConstEval("cannot evaluate a variable", ast.loc);
+    }
+
+    public interpretMethodCall(ast: AstMethodCall): Value {
+        const methodName = ast.method;
+        const args = ast.args;
+        const source = ast.loc;
+        const object = ast.self;
```
anton-trunov: I'd prefer not to have any local aliases in such cases, because you need to track renames when reading the source code. Also, `loc` is sort of standard, so I don't get the point of renaming it.

```suggestion
```

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+        const valOperand = this.interpretExpression(operand);
+
+        return evalUnaryOp(op, valOperand, operand.loc, source);
+    }
+
+    public interpretBinaryOp(ast: AstOpBinary): Value {
+        const left = ast.left;
+        const right = ast.right;
+        const op = ast.op;
+        const source = ast.loc;
```
anton-trunov: Let's not have these aliases
```suggestion
```

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+        const valLeft = this.interpretExpression(left);
+        const valRight = this.interpretExpression(right);
+
+        return evalBinaryOp(op, valLeft, valRight, left.loc, right.loc, source);
+    }
+
+    public interpretConditional(ast: AstConditional): Value {
+        const condition = ast.condition;
+        const thenBranch = ast.thenBranch;
+        const elseBranch = ast.elseBranch;
```
anton-trunov: same here and elsewhere
```suggestion
```

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+                    selfTypeRef.name,
+                );
+                const foundContractConst =
+                    contractTypeDescription.constants.find((constId) =>
+                        eqNames(fieldId, constId.name),
+                    );
+                if (foundContractConst === undefined) {
+                    // not a constant, e.g. `self.storageVariable`
+                    throwNonFatalErrorConstEval(
+                        `cannot a evaluate non-constant self field access`,
```
anton-trunov:
```suggestion
                        `cannot evaluate non-constant self field access`,
```

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+            resObj[idText(fieldWithInit.field)] = this.interpretExpression(
+                fieldWithInit.initializer,
+            );
+            return resObj;
+        }, resultWithDefaultFields);
+    }
+
+    public interpretFieldAccess(ast: AstFieldAccess): Value {
+        const structExpr = ast.aggregate;
+        const fieldId = ast.field;
```
anton-trunov: I used the term "aggregate" to generalize over structs, messages, contracts, traits
```suggestion
```

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+
+const defaultInterpreterConfig: InterpreterConfig = {
+    // We set the default max number of loop iterations
+    // to the maximum number allowed for repeat loops
+    maxLoopIterations: maxRepeatStatement,
+};
+
+export class Interpreter {
+    private envStack: EnvironmentStack;
+    private initialContext: CompilerContext;
```
anton-trunov: should this be called `context` instead of `initialContext`?

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+    }
+
+    public interpretFieldAccess(ast: AstFieldAccess): Value {
+        const structExpr = ast.aggregate;
+        const fieldId = ast.field;
+
+        // special case for contract/trait constant accesses via `self.constant`
+        if (structExpr.kind === "id" && isSelfId(structExpr)) {
+            const selfTypeRef = getExpType(this.initialContext, structExpr);
+            if (selfTypeRef.kind == "ref") {
```
anton-trunov:
```suggestion
            if (selfTypeRef.kind === "ref") {
```

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+                    throwErrorConstEval(
+                        `cannot evaluate declared contract/trait constant ${idTextErr(fieldId)} as it does not have a body`,
+                        fieldId.loc,
+                    );
+                }
+            }
+        }
+        const valStruct = this.interpretExpression(structExpr);
+        if (
+            valStruct == null ||
```
anton-trunov:
```suggestion
            valStruct === null ||
```

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+            return resObj;
+        }, resultWithDefaultFields);
+    }
+
+    public interpretFieldAccess(ast: AstFieldAccess): Value {
+        const structExpr = ast.aggregate;
+        const fieldId = ast.field;
+
+        // special case for contract/trait constant accesses via `self.constant`
+        if (structExpr.kind === "id" && isSelfId(structExpr)) {
```
anton-trunov: A struct can also be referred to as `self` if it's the self argument of a method. Looks like you need to filter out this case, otherwise a legitimate case would generate an error

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+                    // At this point, the function did not execute a return.
+                } catch (e) {
+                    if (e instanceof ReturnSignal) {
+                        const val = e.getValue();
+                        if (val !== undefined) {
+                            return val;
+                        }
+                        // The function executed a return without a value
+                    } else {
+                        throw e;
```
anton-trunov, line 1217: This should probably be an internal compiler error, a.k.a. ICE

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+                    }
+                }
+                // If execution reaches this point, it means that
+                // the function had no return statement or executed a return
+                // without a value. This is an error only if the return type of the
+                // function is not void
+                if (returns.kind !== "void") {
+                    throwErrorConstEval(
```
anton-trunov: looks like this should be an internal compiler error

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+            case "statement_until":
+                this.interpretUntilStatement(ast);
+                break;
+            case "statement_while":
+                this.interpretWhileStatement(ast);
+                break;
+        }
+    }
+
+    public interpretLetStatement(ast: AstStatementLet) {
```
anton-trunov, line 1281: let's also check here there is no shadowing

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+    public interpretAssignStatement(ast: AstStatementAssign) {
+        const path = ast.path;
+        const exp = ast.expression;
+
+        if (path.kind === "id") {
+            const val = this.interpretExpression(exp);
+            this.envStack.updateBinding(idText(path), val);
+        } else {
+            throwNonFatalErrorConstEval(
+                `only identifiers are currently supported as path expressions`,
```
anton-trunov: I'm wondering why ESLint does not complain here: you are using backticks for a string that does not have any interpreted TS expressions inside

src/interpreter.ts
```diff
@@ -25,3 +499,840 @@ export function parseAndEvalExpression(sourceCode: string): EvalResult {
...
+        }
+    }
+
+    public interpretConditionStatement(ast: AstCondition) {
+        const condition = ensureBoolean(
+            this.interpretExpression(ast.condition),
+            ast.condition.loc,
+        );
+        if (condition) {
+            this.envStack.executeInNewEnvironment(() => {
```
anton-trunov, line 1338: why do you need a new environment here? conditionals in Tact do not introduce new binders

anton-trunov: ah, nevermind, this is to get rid of local vars from the new statement blocks

anton-trunov: I think we could probably define a helper function which could be reused for every block statement execution, basically abstracting this
```
this.envStack.executeInNewEnvironment(() => {
                ast.trueStatements.forEach(this.interpretStatement, this);
            });
```

and then reusing it for the else branch, while, repeat, do loops (and the currently missing block statement)

anton-trunov: or maybe for loops it's not a good idea

src/interpreter.ts
jeshecdom: I only added checks during let declarations and function calls, because those are the only places where the runtime environment grows at this moment.

anton-trunov:
Let's also have a negative test for the case when a let-binding shadows a let-binding:

```
let x = 42;
let x = 43;
```

anton-trunov:
To perform that test you might need to turn off the typechecker.

jeshecdom:
> Let's also have a negative test for the case when a let-binding shadows a let-binding:
> 
> ```
> let x = 42;
> let x = 43;
> ```

jeshecdom:
> Let's also have a negative test for the case when a let-binding shadows a let-binding:
> 
> ```
> let x = 42;
> let x = 43;
> ```

Currently, the interpreter will simply overwrite the variable in the runtime environment because it was relying on the typechecker. Actually, I rely on runtime shadowing for certain uses cases, for example, for modelling recursion.

jeshecdom:
> Let's also have a negative test for the case when a let-binding shadows a let-binding:
> 
> ```
> let x = 42;
> let x = 43;
> ```

Currently, the interpreter will simply overwrite the variable in the runtime environment because it was relying on the typechecker. And actually, I rely on runtime shadowing of local variables for certain uses cases, for example, for modelling recursion in a simple manner. Suppose the following recursive function:
```
fun foo(v: Int) {
   if (v == 0) {
      return;
   }
   foo(v - 1);
}
```
And suppose `foo(4)` is called.

The environment stack will look as follows just before `foo` finishes execution (the arrows point to their parent environment in the stack):
```
v = 4   <------  v = 3  <-------- v = 2  <-------- v = 1  <------ v = 0
```
As you can see, local `v` is being shadowed by the children environments. If I add local variable shadowing checks when a variable is declared, I would need to add special cases for recursion in order to allow the above behavior, and for this, I would need to add state in the environment stack to know if I am currently in a recursive call (probably I will need to add some map telling me the names of all currently called functions in the stack: if I am going to add debugging features in the future, probably I will need to add this anyway, since the debugger needs to show a calling stack).

But regarding the shadowing checks, I am starting to get worried that the interpreter will do code duplication for certain checks done by the typechecker. Do you think the interpreter should call the typechecker before the interpreter executes each command? Probably this is a case of turning on and off the typechecker monad in the collecting semantics? I am not sure how to proceed here.

anton-trunov:
The case of modeling recursion looks more than reasonable, let's rely on the typechecker for this. Also, let's document your reasoning somewhere as a design decision comment, this is really useful.

Btw, we have scoping test files prefixed with `var-scope-`, it would be nice if you checked we are not missing some corner cases there, given that you are relying on those static checks.

jeshecdom:
> The case of modeling recursion looks more than reasonable, let's rely on the typechecker for this. Also, let's document your reasoning somewhere as a design decision comment, this is really useful.
> 

Would it be ok if I document this recursion behavior in the code itself? I will also add comments in the Interpreter class itself, saying that you should call the typechecker before calling the interpreter and pass to the interpreter the `CompilerContext` that the typechecker manipulated.

jeshecdom:
> .... you should call the typechecker before calling the interpreter and pass to the interpreter the `CompilerContext` that the typechecker manipulated.

The reason for passing the `CompilerContext` of the typechecker to the interpreter is that the interpreter should also work on code that the interpreter did not execute itself (this is the use case for the partial evaluator). For example, in this code (I marked inside square brackets [ ] the places where the interpreter gets called to simplify expressions):
```
const C: Int = [1];

contract TestContract {

   get fun test(): Int {
      return [C + 1];
   }
}
```
When the interpreter gets called in the brackets, it does not know what other code is surrounding those brackets, because the interpreter did not execute the code outside the brackets. Hence, it relies on the typechecker to receive the `CompilerContext` that includes the declarations in the code (the constant `C` for example).

Another use case is when the interpreter gets called in a REPL. In this case, it knows the entire code, but I think it just should call the typechecker on the given code and then attempt the interpretation.

anton-trunov:
Got it, thanks for clarifying it. Documenting it would be highly appreciated!

anton-trunov (approved):

PR #666, opened 2024-08-07
Allow returning `self` from getters

### LINKED ISSUE #579: Returning `self` from getters

Gusarich:
Closes #579 


src/bindings/typescript/serializers.ts
```diff
@@ -558,9 +558,9 @@ const struct: Serializer<{ name: string; optional: boolean }> = {
     },
     tsType(v) {
         if (v.optional) {
-            return v.name + " | null";
+            return v.name + "_struct | null";
```
anton-trunov: why this change?

Gusarich: Because I've added `_struct` suffix to struct/contract types in TS, so that they don't overlap with contract names

anton-trunov: Hmmm, this breaks some existing tests (like the ones in `wallet.spec.ts`). I would add a prefix/suffix to a less frequent use case of the contract "struct" (like in the case of a getter returning a contract) to break as little existing test code as possible.

Gusarich: Done!

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Returning `self` from getters is now allowed: PR [#666](https://github.com/tact-lang/tact/pull/666)
```
anton-trunov: Let's move this to `Fixed`

src/types/resolveDescriptors.ts
```diff
@@ -1818,7 +1818,7 @@ export function getType(
     ident: AstId | string,
 ): TypeDescription {
     const name = typeof ident === "string" ? ident : idText(ident);
-    const r = store.get(ctx, name);
+    const r = store.get(ctx, name.endsWith("$Data") ? name.slice(0, -5) : name);
```
anton-trunov: Looks like this check should be moved to the place(s) related to ABI?

anton-trunov (changes requested):

src/types/resolveDescriptors.ts
Gusarich: done, have a look

anton-trunov (approved):
LGTM!

PR #669, opened 2024-08-08
Throw internal/compiler errors instead `Error` when possible

### LINKED ISSUE #645: Throw `TactCompilationError` instead `Error` when possible
Tact compiler throws exceptions when compilation errors occur. Typically, it uses `throwCompilationError` for this.

This enables the third-party tools to understand the reason of the exception by inspecting the error stacktrace, e.g.:
```typescript
  static isCompilationError(stack: string | undefined): boolean {
    return stack !== undefined && stack.includes("at throwCompilationError");
  }
```

The problem is that in some cases we use `throw new Error` when compilation error occurs. For example, when an unresolved import appears:

https://github.com/tact-lang/tact/blob/2524aae05b558deaab3ce01086aefc21d65373c8/src/imports/resolveImports.ts#L50

The idea is to examine all the cases when we `throw` and to replace them with `throwCompilationError` or `throwInternalCompilerError` if possible.

This will be helpful for the tooling and will provide a more unified internal API, fostering debugging of Tact.

jubnzv:
Closes #645

I have replaced `throw Error` with `throwInternalCompilerError` in most cases, as that's essentially how the compiler should behave. If the user encounters an "impossible" case, they should definitely be encouraged to report it.

Otherwise, if we encounter this kind of error while using the compiler API in a tool, we should understand that we are using it incorrectly.

Note that I didn't modify the old backend, as it will be rewritten anyway.

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -5,8 +5,8 @@ import { AstFuncId, AstId, AstTypeId, SrcInfo } from "./grammar/ast";
 import { ItemOrigin } from "./grammar/grammar";
 
 export class TactError extends Error {
-    readonly loc: SrcInfo;
-    constructor(message: string, loc: SrcInfo) {
+    readonly loc?: SrcInfo;
+    constructor(message: string, loc?: SrcInfo) {
         super(message);
         this.loc = loc;
     }
@@ -26,13 +26,13 @@ export class TactSyntaxError extends TactError {
 
 /// This will be split at least into two categories: typechecking and codegen errors
 export class TactCompilationError extends TactError {
-    constructor(message: string, loc: SrcInfo) {
+    constructor(message: string, loc?: SrcInfo) {
         super(message, loc);
     }
 }
 
 export class TactInternalCompilerError extends TactError {
-    constructor(message: string, loc: SrcInfo) {
+    constructor(message: string, loc?: SrcInfo) {
         super(message, loc);
     }
 }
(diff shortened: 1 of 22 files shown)


src/storage/allocator.ts
```diff
@@ -272,7 +289,9 @@ export function getAllocationOperationFromField(
                     size,
                 };
             } else if (src.format !== undefined && src.format !== null) {
-                throw Error("Unsupported struct format " + src.format);
+                throwInternalCompilerError(
```
jubnzv, line 292: ESLint has a false positive for this case.

`throwInternalCompilerError` returns `never`, but it treats it like it doesn't terminate control flow here. If we replace this line with `throw new Error("42")`, there won't be a warning.

In other cases, it doesn't show false positives. I assume the problem is in their logic of handling control flow termination in if-else clauses.

So, I'm suppressing this.

anton-trunov (dismissed):
Yay!

anton-trunov:
Oh, btw, let's add the Tact errors to `index.ts` as well in this PR

anton-trunov (approved):

src/errors.ts
```diff
@@ -80,21 +80,28 @@ export function throwSyntaxError(message: string, source: SrcInfo): never {
...
-export function throwCompilationError(message: string, source: SrcInfo): never {
-    throw new TactCompilationError(
-        `${locationStr(source)}${message}\n${source.interval.getLineAndColumnMessage()}`,
-        source,
-    );
+export function throwCompilationError(
+    message: string,
+    source?: SrcInfo,
+): never {
+    const loc =
```
Gusarich, line 87: It's now less informative in some cases. For example, when trying to compile non-existing project source:
```
💼 Compiling project getters ...
Tact compilation failed
Error
    at throwCompilationError (/Users/daniil/Coding/tact/src/errors.ts:91:11)
    at resolveImports (/Users/daniil/Coding/tact/src/imports/resolveImports.ts:28:30)
    at precompile (/Users/daniil/Coding/tact/src/pipeline/precompile.ts:18:36)
    at build (/Users/daniil/Coding/tact/src/pipeline/build.ts:73:25)
    at run (/Users/daniil/Coding/tact/src/node.ts:130:34)
    at async /Users/daniil/Coding/tact/scripts/prepare.ts:20:31
Error: Tact projects compilation failed
    at /Users/daniil/Coding/tact/scripts/prepare.ts:24:19
error Command failed with exit code 1.
```

while before it would throw `Could not find entrypoint XXX`.

I think the problem is here, because it ignores `message` in case if no source location was provided.

PR #680, opened 2024-08-13
fix: check new variable bindings do not shadow global constants

### LINKED ISSUE #676: Constant shadowing does not produce var-scope error
```
const foo: Int = 42;

contract Test {
    get fun bar(): Int {
        let foo = 43;  // <--- attempt to shadow `foo` constant
        return foo;
    }
}
```

Moreover, the constant substitution algorithm does not respect scoping rules and the `bar` getter returns `42` instead of expected `43`.

anton-trunov:
Closes #676

### DIFF
--- a/src/types/resolveStatements.ts
+++ b/src/types/resolveStatements.ts
@@ -18,6 +18,7 @@ import {
 import {
     getAllStaticFunctions,
     getAllTypes,
+    hasStaticConstant,
     resolveTypeRef,
 } from "./resolveDescriptors";
 import { getExpType, resolveExpression } from "./resolveExpression";
@@ -42,10 +43,20 @@ export function emptyContext(
     };
 }
 
-function checkVariableExists(ctx: StatementContext, name: AstId): void {
-    if (ctx.vars.has(idText(name))) {
+function checkVariableExists(
+    ctx: CompilerContext,
+    sctx: StatementContext,
+    name: AstId,
+): void {
+    if (sctx.vars.has(idText(name))) {
         throwCompilationError(
             `Variable already exists: ${idTextErr(name)}`,
+            name.loc,
+        );
+    }
+    if (hasStaticConstant(ctx, idText(name))) {
+        throwCompilationError(
+            `Variable ${idTextErr(name)} is trying to shadow an existing constant with the same name`,
             name.loc,
         );
     }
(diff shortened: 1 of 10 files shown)


anton-trunov:
Please let me know if there are any issues in this PR and I'll happily fix those in a follow-up, but now I'm going to merge this one to fix the other issues for v1.4.2

src/types/resolveStatements.ts
```diff
@@ -81,15 +92,16 @@ function removeRequiredVariable(
 function addVariable(
     name: AstId,
     ref: TypeRef,
-    src: StatementContext,
+    ctx: CompilerContext,
+    sctx: StatementContext,
 ): StatementContext {
-    checkVariableExists(src, name); // Should happen earlier
+    checkVariableExists(ctx, sctx, name); // Should happen earlier
     if (isWildcard(name)) {
```
jeshecdom, line 99: Ohhh, interesting. I did not check in the interpreter the use of wildcard identifiers (to be honest, I was not aware of them). Where is the wildcard _ allowed in Tact? Anywhere where an identifier is allowed? I looked in the grammar and it seems that _ is allowed by the id rule. _ allows you to ignore that identifier, right? I am asking this for the interpreter :)

src/types/resolveStatements.ts
```diff
@@ -197,7 +209,7 @@ function processStatements(
                     ctx = resolveExpression(s.expression, sctx, ctx);
 
                     // Check variable name
-                    checkVariableExists(sctx, s.name);
+                    checkVariableExists(ctx, sctx, s.name);
```
jeshecdom, line 212: Is there any reason as to why call `checkVariableExists` at this point if later `addVariable` will call it in both branches of the conditional? I guess to give the user the variable duplication and shadowing errors as soon as possible.

src/types/resolveStatements.ts
```diff
@@ -486,11 +498,12 @@ function processStatements(
                     let catchCtx = sctx;
 
                     // Process catchName variable for exit code
-                    checkVariableExists(initialCtx, s.catchName);
+                    checkVariableExists(ctx, initialSctx, s.catchName);
```
jeshecdom, line 501: Again. Why call `checkVariableExists` if `addVariable` is the first thing that it does in the next line?

src/types/resolveStatements.ts
anton-trunov: Right, underscores (`_`) means an identifier that cannot be used (so the shadowing rules do not apply to them). You can use it as a function parameter, in `catch`-blocks, `foreach`-loops, unused `let`-bindings.

src/types/resolveStatements.ts
anton-trunov: I actually didn't think about this, I just added `ctx` mechanically here.

src/types/resolveStatements.ts
anton-trunov: We could try refactor this, although I'm not sure how useful this is going to be since we are going to replace the existing typechecker with a new one

jeshecdom: Agreed. I think there is no need to refactor.

PR #689, opened 2024-08-14
AST Comparison API

### LINKED ISSUE #686: AST Comparison API
We need an API that provides a way to compare any Tact AST:

* [x] `AstComparator`:
  + [x] Provide an API to check AST equivalence without renaming
* [x] `AstHasher`: Write the hashing module that doesn't take names into account
* [ ] `AstRenamer`: Canonicalize binder names introducing something like de Bruijn indices
  + [ ] Rename elements
    - [x] Top-level/contract/trait entries
    - [x] Native functions: rename to match their original names.
    - [ ] Struct/message/contract fields
    - [ ] Function arguments
  + [ ] Handle dependencies
    - [ ] Build dependency graphs for the AST entries. This is required to rename and sort them properly. (TODO: create a separate module for this?)
    - [ ] Provide topological sort of the renamed entries. We have to write additional logic to handle mutually recursive function calls.
  + [ ] Rename uses of renamed elements in the source code
    - [ ] Struct/contract initializers
    - [ ] Function/method calls
    - [ ] Function arguments
    - [ ] Implemented traits in traits/contracts
    - [ ] Fields (it will be easier when introducing types in AST #314)
* [ ] `AstSorter`: Improve the sorting routines
  + [x] Lexicographically sort top-level entries
  + [ ] Imports
  + [x] Function/constant/trait/contract attributes
* [ ] Compare ASTs after partial evaluation

jubnzv:
Towards #686


cspell.json
```diff
@@ -80,6 +80,10 @@
     "привет",
     "PUSHREF",
     "PUSHSLICE",
+    "renamer",
+    "canonizalize",
```
anton-trunov:
```suggestion
    "canonicalize",
```

knip.json
```diff
@@ -4,6 +4,7 @@
   "project": ["src/**/*.ts", "bin/tact.js"],
   "ignore": [
     "src/grammar/ast.ts",
+    "src/grammar/sort.ts",
```
anton-trunov: I think this should not be ignored if we add the API to `index.ts`

anton-trunov (changes requested):
LGTM! Let's fix merge conflicts and resolve a couple comments I left

anton-trunov (approved):

PR #704, opened 2024-08-17
fix(typechecker): verify map type is well-formed

### LINKED ISSUE #701: Unsupported map types are allowed in function parameters
The following Tact module correctly fails to compile because `String` is not currently supported as a map value type.
```tact
contract BlankContract { m: map<Int, String>; }
```

But the following compiles just fine:

```tact
fun foo(m: map<Int, String>) { }
```

anton-trunov:
Closes #701

### DIFF
--- a/src/types/__snapshots__/resolveDescriptors.spec.ts.snap
+++ b/src/types/__snapshots__/resolveDescriptors.spec.ts.snap
@@ -486,6 +486,56 @@ Line 7, col 1:
 "
 `;
 
+exports[`resolveDescriptors should fail descriptors for wf-type-const 1`] = `
+"<unknown>:4:19: Invalid key type for map. Check https://docs.tact-lang.org/book/maps#allowed-types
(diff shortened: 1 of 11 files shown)


src/types/__snapshots__/resolveDescriptors.spec.ts.snap
```diff
@@ -486,6 +486,56 @@ Line 7, col 1:
 "
 `;
 
+exports[`resolveDescriptors should fail descriptors for wf-type-const 1`] = `
+"<unknown>:4:19: Invalid key type for map. Check https://docs.tact-lang.org/book/maps#allowed-types
```
Gusarich, line 490: why does it say `Invalid key type` and points to a value type?

anton-trunov: ah, good catch, forgot to remove `key` during refactoring

Issue #720, opened 2024-08-23
Ability to change semantics of interpreter


jeshecdom:
We should be able to change the semantics of the interpreter. For example, using collecting semantics or a different abstract machine.

PR #721, opened 2024-08-23
"Monadish" refactoring of interpreter

jeshecdom:
Towards #720 

This PR does not introduce new functionality to the interpreter, it only refactors the interpreter to be able to change the semantics or add additional functionality to it.

At first, I was trying to do a monad-style interpreter, but TypeScript does not have do-notation, which can quickly produce unreadable code because of the bind functions. Although, there are some suggestions over the internet on how to emulate do-notation using generators, it looks ad-hoc, hacky, and unnatural. So, I decided to take the idea of monad and adapt it into an imperative flavor. The idea is to define the semantics of an interpreter as an abstract class parametric over two generic types: "V" (the type of expressions' results) and "S" (the type of statements' results).
At the moment, module level declarations do not have an associated generic type, but it will probably be added.

Specific semantics can be represented by implementing the abstract class. For example, to create an interpreter with the standard semantics, just write:
```
const interpreter = new Interpreter(new StandardSemantics(ctx));
```
If we want to change the behavior, for example, by collecting special values when statements execute, just pass another semantics:
```
const interpreter = new Interpreter(new YourCustomSemantics(ctx));
```
This is a first version of the "monadish" interpreter. As such, it probably will suffer changes as more semantics are added since they will probably force changes to the interface in the abstract class. 

You can see the abstract class interface in the file `src/interpreterSemantics/types.ts`.

### DIFF
--- a/src/constEval.ts
+++ b/src/constEval.ts
@@ -18,12 +18,13 @@ import {
 import { ExpressionTransformer } from "./optimizer/types";
 import { StandardOptimizer } from "./optimizer/standardOptimizer";
 import {
-    Interpreter,
     ensureInt,
     evalBinaryOp,
     evalUnaryOp,
-    throwNonFatalErrorConstEval,
-} from "./interpreter";
+    StandardSemantics,
+} from "./interpreterSemantics/standardSemantics";
+import { Interpreter } from "./interpreter";
+import { throwNonFatalErrorConstEval } from "./interpreterSemantics/util";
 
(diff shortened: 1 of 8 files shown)


jeshecdom:
This PR is obsolete. Its code has been merged into PR #852.

verytactical:
> This PR is obsolete. Its code has been merged into PR https://github.com/tact-lang/tact/pull/852.

Please consider closing it :)

Issue #726, opened 2024-08-25
Improve PR template


anton-trunov:
It needs some separation of PR categories so we don't have to manually delete things.

anton-trunov:
#727 deals with it

PR #727, opened 2024-08-25
chore: categorized PR templates

anton-trunov:
GitHub does not have native support for PR templates in its frontend, this employes the hack described here:
https://github.com/kenji-miyake/test-pr-template.

You can see a screencast of how it works here: https://github.com/orgs/community/discussions/4620#discussioncomment-3383383.

## Issue

Closes issue #726.

## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/api.md
+++ b/.github/PULL_REQUEST_TEMPLATE/api.md
@@ -10,7 +10,7 @@ Closes issue #number.
 ## Checklist
 
 - [ ] I have updated CHANGELOG.md
-- [ ] This part of API is exported from `src/index.ts`
+- [ ] This part of API is exported from the corresponding `index.ts`
 - [ ] I have run all the tests locally and no test failure was reported
 - [ ] I have run the linter, formatter and spellchecker
 - [ ] I did not do unrelated and/or undiscussed refactorings
--- a/.github/PULL_REQUEST_TEMPLATE/api.md
+++ b/.github/PULL_REQUEST_TEMPLATE/api.md
@@ -1,0 +1,16 @@
+<!--
+IMPORTANT:
+If your PR doesn't close a particular issue, please, create the issue first and describe the whole context: what you're adding/changing and why you're doing so. And only then open the Pull Request, which would close that issue!
+-->
+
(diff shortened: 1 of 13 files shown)


.github/PULL_REQUEST_TEMPLATE/api.md
```diff
@@ -0,0 +1,16 @@
...
+-->
+
+## Issue
+
+Closes issue #number.
+
+## Checklist
+
+- [ ] I have updated CHANGELOG.md
+- [ ] This part of API is exported from `src/index.ts`
```
jubnzv: I'm a bit skeptical about that kind of export.

Perhaps it would be better to create a few `index.ts` files inside different packages to avoid name clashes and make the meaning of the specific import clearer for the developer. For example, I would definitely suggest having separate `grammar/index.ts` and `func/index.ts` imports, because they have the same function and type names.

Thus, I would suggest changing it like this:

```suggestion
- [ ] This part of API is exported from `index.ts`
```

novusnota: Creating `index.ts` files in each folder inside `src/` where each one simply exports all the contents of other files is a common pattern in JS/TS world for hacking together directory-scoped imports :)

So while I agree about starting to follow the `index.ts` pattern, the actual line in the PR here probably meant the global, publicly exposed API elements, if I got @anton-trunov idea right. Thus, `src/index.ts` probably has to stay as is, and all collisions into it may be resolved by import aliases: `import { parse as parseTact } from ...`, for example

UPD: Or we might just split the API into two: one for FunC parts, and the rest for Tact

anton-trunov: @jubnzv Sure, the multiple exports idea sounds better. Here is the corresponding issue: https://github.com/tact-lang/tact/issues/740

anton-trunov:
```suggestion
- [ ] This part of API is exported from the corresponding `index.ts`
```

anton-trunov: I just slightly changed the wording here, I hope it's fine

PR #764, opened 2024-08-27
Adding readFile function and binary strings support

### LINKED ISSUE #748: Add support for non-UTF8 binary-like strings (or add format for working with binary data)
## The problem 
Right now there is no way to store any binary data in tact, because the language represents any string as  UTF-8. For example, I want to store small JPEG image (few kilobates) inside smart-contract (I want to HARDCODE it) and then return it inside `get_nft_data` contract getter (in other words, hardcode onchain NFT data). Yes, I understand that i may store it as a cell variable and then set it in initialization, but its not as safe as using constant(BTW needs more deployment code).

## Example

Here is my [smart contract](https://github.com/imartemy1524/tact-nft-onchain-data/blob/master/contracts/on_chain_data.tact), which stores on-chain NFT metadata, and returns it in getter; [example of binary data](https://github.com/imartemy1524/tact-nft-onchain-data/blob/master/contracts/img.png.tact) (but indeed tact stores it as UTF-8 ). But unfortunately, I'm not able to hardcode it using tact - the only way to build a "string cell", is to use func with tact wrappers (something like ```slice __gen_binary_string() asm "B{generated_} B>boc <s PUSHSLICE";```

Here are [unit tests](https://github.com/imartemy1524/tact-nft-onchain-data/blob/master/tests/OnChainData.spec.ts#L40), which shows the problem.

## Possible solutions

I see few options, which can help with this situation:

1. Make possible to create strings using binary (or other non-utf8 encodings). For example, like in python, using `b` literal: `b"binary-data-string\x12\xFF"`
2. Make a different "binary string" structure (for example, `BinaryString` instead of string).
3. Create compile-time function `readFile(filename, encoding)`, which would read specified file contents as a string with specified encoding.

#### Why it is convenient to use `Strings` object for binary data?

Tact perfectly encodes/decodes long strings into cells, using standard stringbuilders/toCell methods, which are really useful in case of serializing binary data

imartemy1524:
## Issue

Closes #748.

## Checklist


imartemy1524:
This closes only few problems, because right now there are no reasonable method to **decode** binary strings using standard library (I implemented it ~~myself~~ in [reading-file.spec.ts](./764/files/#diff-e9deb538bee9bf113b10dd4c0d6fbc08cc5ee3c6460e16034ae822745d7615ddR9) - this is actually `@ton/core` issue)

The way I see it: introduce new type, for example `Binary` or `BinaryString`, which in TACT and func be alias for string, but in `TypeScript` generated code, all structures with this method would contains `Buffer` instead of `string`.
Tact:
```rust
...
trait BinaryData{
    get fun binaryData(): Binary{
        return readFile("./metadata.raw", "binary");
    }
}
...
```
TypeScript:
```typescript
class BinaryContract{
    getBinaryData(): Buffer{
        //implementation
    }
}
```

Feel free to completely disagree with me :+1:

anton-trunov:
Closing this for now, we can discuss the design of this feature in the linked issue

PR #769, opened 2024-08-30
feat: functions with `asm` bodies

### LINKED ISSUE #768: Support `asm` blocks in Tact
For instance, when one has to use the Keccak hashing, they need to add a FunC function wrapper with an `asm` block:

```func
/// cryptolib.fc

int keccak256(slice s) asm "1 INT HASHEXT_KECCAK256";
```

and then create a Tact binding:

```tact
/// Tact file

import "./cryptolib.fc";

@name(keccak256)
native keccak256(s: Slice): Int;
```

Instead, it would be more convenient to do it like so (concrete syntax is subject discussion):

```tact
/// Tact file

fun keccak256(s: Slice): Int {
  asm { 1 INT HASHEXT_KECCAK256 }
}
```

Of course, `asm`-blocks should support things like `(-> 1 0 2)`.

anton-trunov:
## Issue

Closes #768.

## Checklist


src/grammar/grammar.ohm
```diff
@@ -93,6 +98,10 @@ Tact {
              | external "(" Parameter? ")" "{" Statement* "}" --externalRegular
              | external "(" stringLiteral ")" "{" Statement* "}" --externalComment
 
+    AsmInstruction = id
+                   | integerLiteralDec
+                   | integerLiteralHex
```
novusnota: AFAIK, asm instructions can also have various symbols in them, including, but not limited to: `[`, `]`, `(`, `)`, `"`, `:`, `}`, etc.

See:
* https://docs.ton.org/learn/tvm-instructions/instructions
* https://github.com/tonkeeper/w5/blob/main/contracts/wallet_v5.fc

So, the (rough!!) suggestion is:
```suggestion
    asmInstruction = ~comment (~space ~"}" any)+
```

Notice the lowercase name to prevent from optional spaces occuring in-between things. The `AsmFunction` rule on line 25 also has to use the lowercase version then :)

The approach above is not perfect and may be improved, since `{` and `}` are also allowed in Fift asm and denote a block of instructions (not always, though)

anton-trunov: @novusnota Nice catch! Please checkout my update. I had to introduce explicit grammar category for data, otherwise there is a conflict with the curly braces from the enclosing block

src/grammar/grammar.ohm
```diff
@@ -20,6 +20,9 @@ Tact {
     PrimitiveTypeDecl = primitive typeId ";"
 
     ModuleFunction = FunctionDefinition
+                   | AsmFunction
+
+    AsmFunction = "asm" AsmShuffle? FunctionAttribute* fun id Parameters (":" Type)? "{" asmInstruction+ "}"
```
novusnota, line 25: Great! Now, to also support blocked instructions I'm thinking of something like this:

```ohm
AsmFunction = "asm" AsmShuffle? FunctionAttribute* fun id Parameters (":" Type)? "{" AsmInstruction* "}"

// ...

AsmInstruction = "{" AsmInstruction+ "}" --block
               | asmInstruction

// ...

asmInstruction = ("x" | "X" | "b" | "B") "{" hexDigit* "}" --data
               | ~comment (~space ~"}" any)+ --any
```

anton-trunov: Getting all the grammar right as a non-goal of this PR: we need a solution which would allow us to port `asm` functions from `stdlib.fc` and `mathlib.fc`.

PR #787, opened 2024-09-03
feat: `slice`, `rawSlice`, `ascii` and `crc32` built-in functions

### LINKED ISSUE #444: Introduce comptime builtins corresponding to FunC string literal types
like `h`, `c`, etc. See https://docs.ton.org/develop/func/literals_identifiers.

I don't think we should extend the syntax of string literals with these suffixes, because those are too cryptic for beginners, we should just introduce those as compile-time builtin functions, like `crc32("foobar")` as a Tact analogue for `"foobar"c` in FunC.

- [x] `"string"s` — defines a raw slice const by its contents (hex-encoded and optionally bit-padded) [added in PR [#787](https://github.com/tact-lang/tact/pull/787) - `rawSlice("string")`];
- [x] `"string"a` — creates a slice const containing MsgAddressInt structure from a specified address [already have it: `address("string")`];
- [x] `"string"u` — creates an int const that corresponds to the hex values of the provided ASCII string [added in PR [#787](https://github.com/tact-lang/tact/pull/787) - `ascii("string")`];
- [x] `"string"h` — creates an int const that is the first 32 bits of the SHA256 hash of the string [no need: can be solved with `sha256` and a bit mask, this will be optimized in compile-time];
- [x] `"string"H` — creates an int const that is all 256 bits of the SHA256 hash of the string [already have it: `sha256("string")`];
- [x] `"string"c` — creates an int const that is crc32 value of the string [added in PR [#787](https://github.com/tact-lang/tact/pull/787): `crc32("string")`].

Gusarich:
## Issue

Closes #444.

* `slice()` isn't actually an equivalent of any string literals, but just a more efficient way to define slice constants from BOC, similarly to `cell()`
* `rawSlice("abcd")` is an equivalent of `"abcd"s` in FunC
* `ascii("hello world")` is an equivalent of `"hello world"u` in FunC
* `crc32("hello world")` is an equivalent of `"hello world"c` in FunC

## Checklist

### DIFF
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -1,4 +1,5 @@
-import { Address, Cell, toNano } from "@ton/core";
+import { Address, beginCell, Cell, toNano } from "@ton/core";
+import * as crc32 from "crc-32";
 import { evalConstantExpression } from "./constEval";
 import { CompilerContext } from "./context";
 import {
@@ -1068,6 +1069,71 @@ export class Interpreter {
                             ast.loc,
                         );
                     }
+                }
+                break;
+            case "slice":
+                {
+                    ensureFunArity(1, ast.args, ast.loc);
+                    const str = ensureString(
+                        this.interpretExpression(ast.args[0]!),
+                        ast.args[0]!.loc,
+                    );
+                    try {
+                        return Cell.fromBase64(str).asSlice();
+                    } catch (_) {
+                        throwErrorConstEval(
+                            `invalid base64 encoding for a cell: ${str}`,
+                            ast.loc,
+                        );
+                    }
+                }
+                break;
+            case "rawSlice":
+                {
+                    ensureFunArity(1, ast.args, ast.loc);
+                    const str = ensureString(
+                        this.interpretExpression(ast.args[0]!),
+                        ast.args[0]!.loc,
+                    );
+                    try {
+                        return beginCell()
+                            .storeBuffer(Buffer.from(str))
+                            .endCell()
+                            .asSlice();
+                    } catch (_) {
+                        throwErrorConstEval(
+                            `invalid slice data: ${str}`,
+                            ast.loc,
+                        );
+                    }
+                }
+                break;
+            case "ascii":
+                {
+                    ensureFunArity(1, ast.args, ast.loc);
+                    const str = ensureString(
+                        this.interpretExpression(ast.args[0]!),
+                        ast.args[0]!.loc,
+                    );
+                    if (str.length > 32) {
+                        throwErrorConstEval(
+                            `ascii string is too long, expected up to 32 characters, got ${str.length}`,
+                            ast.loc,
+                        );
+                    }
+                    return BigInt(
+                        "0x" + Buffer.from(str, "ascii").toString("hex"),
+                    );
+                }
+                break;
+            case "crc32":
+                {
+                    ensureFunArity(1, ast.args, ast.loc);
+                    const str = ensureString(
+                        this.interpretExpression(ast.args[0]!),
+                        ast.args[0]!.loc,
+                    );
+                    return BigInt(crc32.str(str));
                 }
                 break;
             case "address":
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -1135,6 +1135,12 @@ export class Interpreter {
                             ast.loc,
                         );
                     }
+                    if (str.length == 0) {
+                        throwErrorConstEval(
+                            `ascii string cannot be empty`,
+                            ast.loc,
+                        );
+                    }
                     return BigInt(
                         "0x" + Buffer.from(str, "ascii").toString("hex"),
                     );
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -1095,9 +1095,23 @@ export class Interpreter {
                         this.interpretExpression(ast.args[0]!),
                         ast.args[0]!.loc,
                     );
+
+                    if (str.length > 255) {
+                        throwErrorConstEval(
+                            `hex string is too long, expected up to 255 characters, got ${str.length}`,
+                            ast.loc,
+                        );
+                    }
+                    if (!/^[0-9a-fA-F]*$/.test(str)) {
+                        throwErrorConstEval(
+                            `invalid hex string: ${str}`,
+                            ast.loc,
+                        );
+                    }
+
                     try {
                         return beginCell()
-                            .storeBuffer(Buffer.from(str))
+                            .storeBuffer(Buffer.from(str, "hex"))
                             .endCell()
                             .asSlice();
                     } catch (_) {
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -1133,7 +1133,8 @@ export class Interpreter {
                         this.interpretExpression(ast.args[0]!),
                         ast.args[0]!.loc,
                     );
-                    return BigInt(crc32.str(str));
+                    const c = BigInt(crc32.str(str));
+                    return c < 0 ? c + 4294967296n : c;
                 }
                 break;
             case "address":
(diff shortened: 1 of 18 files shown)


anton-trunov (changes requested):
1. What does `slice` and `rawSlice` correspond to in FunC string suffixes?
2. Let's add some negative tests, for instance, `ascii("000000000000000000000000000000000")` is supposed to fail due to overflow (33 zeros * 8 = 264 bits)`

src/test/e2e-emulated/contracts/intrinsics.tact
```diff
@@ -6,6 +6,10 @@ contract IntrinsicsTester {
     d: Cell = cell("te6cckEBAQEADgAAGEhlbGxvIHdvcmxkIXgtxbw=");
     e: Int = pow(2, 9);
     f: Int = sha256("hello world");
+    g: Slice = slice("te6cckEBAQEADgAAGEhlbGxvIHdvcmxkIXgtxbw=");
+    h: Slice = rawSlice("hello world");
+    i: Int = ascii("hello world");
+    j: Int = crc32("hello world");
```
anton-trunov: Let's reuse the test cases from the TON docs (https://docs.ton.org/develop/func/literals_identifiers#string-literals). For instance:

> `"transfer(slice, int)"c` becomes 2235694568 int const

anton-trunov:
cspell

Gusarich:
fixed

src/test/compilation-failed/const-eval-failed.spec.ts
```diff
@@ -163,4 +163,9 @@ describe("fail-const-eval", () => {
         errorMessage:
             "Cannot evaluate expression to a constant: repeat argument must be a number between -2^256 (inclusive) and 2^31 - 1 (inclusive)",
     });
+    itShouldNotCompile({
```
anton-trunov, line 166: Let's also add a negative test with `ascii("")`

Gusarich: and same for `crc32` too?

anton-trunov: I just checked: in Func `""c` results in `0 PUSHINT`

anton-trunov: btw, `ascii("⚡⚡⚡⚡⚡⚡⚡⚡⚡⚡⚡")` should result in overflow, but `ascii("⚡⚡⚡⚡⚡⚡⚡⚡⚡⚡")` should be equal `1563963554659859369353828835329962428465513941646011501275668087180532385` (let's add these two tests too)

src/test/e2e-emulated/contracts/intrinsics.tact
```diff
@@ -6,6 +6,10 @@ contract IntrinsicsTester {
     d: Cell = cell("te6cckEBAQEADgAAGEhlbGxvIHdvcmxkIXgtxbw=");
     e: Int = pow(2, 9);
     f: Int = sha256("hello world");
+    g: Slice = slice("te6cckEBAQEADgAAGEhlbGxvIHdvcmxkIXgtxbw=");
+    h: Slice = rawSlice("abcdef");
```
anton-trunov, line 10: Let's also have a test with `rawSlice("")`

anton-trunov (commented):
Let's add a negative test with `rawSlice(s)` where `s` won't fit into a cell

src/interpreter.ts
```diff
@@ -1133,7 +1133,8 @@ export class Interpreter {
                         this.interpretExpression(ast.args[0]!),
                         ast.args[0]!.loc,
                     );
-                    return BigInt(crc32.str(str));
+                    const c = BigInt(crc32.str(str));
+                    return c < 0 ? c + 4294967296n : c;
```
anton-trunov: https://github.com/SheetJS/js-crc32?tab=readme-ov-file#signed-integers:

> Unconventional for a CRC32 checksum, this library uses signed 32-bit integers. This is for performance reasons. 

> `x >>> 0` converts a number value to unsigned 32-bit integer.

Alternatively, we could use AWS's implementation of CRC-32 which seems to use unsigned numbers: https://www.npmjs.com/package/@aws-crypto/crc32.

anton-trunov: If you decide to keep js-crc32, then let's switch to `x >>> 0` and add this comment about signed numbers, otherwise it's a bit unclear

src/test/compilation-failed/contracts/const-eval-rawslice-overflow.tact
```diff
@@ -0,0 +1,5 @@
+contract AsciiOverflow {
+    get fun getRawSlice_fail(): Slice {
+        return rawSlice("abcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcd");
```
anton-trunov, line 3: weirdly enough, FunC does not complain about this particular string:
```func
#pragma version =0.4.4;
#pragma allow-post-modification;
#pragma compute-asm-ltr;

_ %foo() method_id(0xff) {
    return "abcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcd"s;
}
```

the code above compiles and produces

```
❯ ./func test.fc
// automatically generated from `test.fc`
255 DECLMETHOD %foo
%foo PROC:<{
  x{abcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcd} PUSHSLICE
}>
```

but if I add one more hex character to the string, then I get an error:

```
test.fc:6:12: error: Invalid hex bitstring constant `abcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcda`
      return "abcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabcdabc ...
             ^
```

Gusarich: seems like a FunC problem. when I run the generated fift output - it crashes.

anton-trunov: Oh, interesting. Can you please file an issue to the monorepo?

src/interpreter.ts
Gusarich: done

src/test/compilation-failed/contracts/const-eval-rawslice-overflow.tact
Gusarich: https://github.com/ton-blockchain/ton/issues/1153

anton-trunov (approved):
Great stuff!

PR #799, opened 2024-09-09
feat: rework `rawSlice` to support padding and correctly handle overflows

### LINKED ISSUE #796: `rawSlice()` should support bit padding
This is a follow-up on #787 

String literal type `s` in FunC supports bit padding.
I'm sure that we should also support that for Tact's `rawSlice()`, because otherwise it won't be possible to define slices with binary data of length not divisible by 4.

This padding basically removes all the trailing zeros and also the last `1` bit before them. So, for example, `4a` is `0100 1010` and `4a_` is `0100 10`.

Gusarich:
## Issue

Closes #796.

## Checklist

### DIFF
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -1120,15 +1120,6 @@ export class Interpreter {
                     if (str.endsWith("_")) {
                         const paddedBits = paddedBufferToBits(buffer);
 
-                        console.log({
-                            originalHex: hex,
-                            paddedHex,
-                            bufferLength: buffer.length,
-                            paddedBitsLength: paddedBits.length,
-                            bitStringLength: bits.length,
-                            offset: hex.length % 2 === 0 ? 0 : 4,
-                        });
-
                         // Ensure there's enough length to apply the offset
                         const offset = hex.length % 2 === 0 ? 0 : 4;
                         if (paddedBits.length >= offset) {
--- a/src/interpreter.ts
+++ b/src/interpreter.ts
@@ -1,4 +1,5 @@
-import { Address, beginCell, Cell, toNano } from "@ton/core";
+import { Address, beginCell, BitString, Cell, toNano } from "@ton/core";
+import { paddedBufferToBits } from "@ton/core/dist/boc/utils/paddedBits";
 import * as crc32 from "crc-32";
 import { evalConstantExpression } from "./constEval";
 import { CompilerContext } from "./context";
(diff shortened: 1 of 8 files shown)


novusnota (dismissed):

src/test/e2e-emulated/contracts/intrinsics.tact
anton-trunov, line 1: Let's add a couple of tests groups from the legendary `tvm.pdf`:

> For example, 8A corresponds to binary string 10001010, while 8A_ and 8A0_ both correspond to 100010.

> An empty bitstring may be represented by either ‘’, ‘8_’, ‘0_’, ‘_’, or ‘00_’.

src/interpreter.ts
```diff
@@ -1096,30 +1097,60 @@ export class Interpreter {
...
+                        const paddedBits = paddedBufferToBits(buffer);
+
+                        console.log({
+                            originalHex: hex,
+                            paddedHex,
+                            bufferLength: buffer.length,
+                            paddedBitsLength: paddedBits.length,
+                            bitStringLength: bits.length,
+                            offset: hex.length % 2 === 0 ? 0 : 4,
+                        });
```
anton-trunov: Is it some left-over from debugging?

src/interpreter.ts
```diff
@@ -1096,30 +1097,60 @@ export class Interpreter {
...
 
-                    try {
-                        return beginCell()
-                            .storeBuffer(Buffer.from(str, "hex"))
-                            .endCell()
-                            .asSlice();
-                    } catch (_) {
+                    // Remove underscores from the hex string
+                    const hex = str.replace("_", "");
+                    const paddedHex = hex.length % 2 === 0 ? hex : "0" + hex;
```
anton-trunov, line 1109: We need a test that justifies this `paddedHex` thing. AFAIU, if the padding is removed the tests won't catch it.

anton-trunov (changes requested):

src/test/e2e-emulated/contracts/intrinsics.tact
Gusarich: done

src/interpreter.ts
Gusarich: removed

src/interpreter.ts
Gusarich: there are multiple test cases with an odd number of digits already: `0_`, `1_`, `7`

anton-trunov: got it, thanks

anton-trunov (approved):

Issue #539, opened 2024-07-10
Test language semantics


anton-trunov:
Related docs issue: https://github.com/tact-lang/tact/issues/1163

#### Expressions, statements

- [X] In functions: pass by value of structs, maps, and contracts (i.e., parameters in functions are copies).
- [X] In assignments: structs, maps and contracts are copied (i.e., assigned by value).
- [X] In mutating functions: structs, maps and contracts are modified if they are the first parameter.
- [X] Mutating functions chaining for structs, maps, contracts, integers.
- [X] Nested maps
- [ ] Augment the tests in `try-catch.spec.ts` with structs, maps, and contracts.
- [X] Check that `&&` and `||` short-circuit.
- [X] Check that ternary operator `_ ? _ : _` short-circuits.
- [X] `parseStdAddress` and `newAddress` are inverse functions for each other.
- [X] Equality of addresses, slices, cells.

#### Variable scoping

- Covered in a separate issue: tact-lang/tact-docs#288

#### Contracts

- [ ] Since contract copies can have their own state but still call contract functions, verify if this produces conflicting results with the original contract when calling library functions. This may include obtaining addresses, sending messages, etc.
- [X] Check that it is possible to update the state of the original contract using a contract copy.
- [ ] Since messages are sent in a different phase, carry out tests that produce errors in the second phase for example.
- [X] Check injectivity of addresses: two contracts with the same code but different initial data have different addresses, two contracts with the same initial data but different code have different addresses.
- [ ] Test sending messages with different sending modes.
- [x] Test that `receive("foo")` takes precedence over `receive(s: String) { if (s == "foo") { ... } }`. Done in #1741 

#### Stdlib

- [ ] In the [library documentation](https://docs.tact-lang.org/ref), many functions have notes and warnings describing some semantical behavior, test those.

#### Exit codes

- Covered in a separate issue: tact-lang/tact-docs#174

anton-trunov:
@jeshecdom Can you please write a small plan in my original post (you can edit it)? The first step 'structs' is already done and you already have planned to work with maps, so let's just document the plan here, in this issue

jeshecdom:
A list of planned semantical tests. Please feel free to add any that you think should be tested:

- [X] In functions: pass by value of structs, maps, and contracts (i.e., parameters in functions are copies).
- [X] In assignments: structs, maps and contracts are copied (i.e., assigned by value).
- [X] In mutating functions: structs, maps and contracts are modified if they are the first parameter.
[...]

PR #809, opened 2024-09-12
Test cases of semantical tests of maps

jeshecdom:
## Issue

Towards #539.

Currently it includes test cases for semantics of maps.

Next PR will be contracts.

## Checklist


src/test/e2e-emulated/contracts/semantics.tact
```diff
@@ -40,6 +40,95 @@ extends mutates fun changeSomeFields(self: SA) {
...
+    // Now delete entries 2 and 3
+
+    m.del(2);
+    m.del(3);
+
+    return m;
+}
+
+/*
+The following two functions are not allowed by the compiler, because mutating functions
```
anton-trunov, line 89: Let's open an issue on this, looks like an important language feature to have in Tact. Btw, a workaround for this would be to wrap a map into a struct and then it's suddenly allowed :)

src/test/e2e-emulated/contracts/semantics.tact
```diff
@@ -40,6 +40,95 @@ extends mutates fun changeSomeFields(self: SA) {
...
+
+    m.del(2);
+    m.del(3);
+
+    return m;
+}
+
+/*
+The following two functions are not allowed by the compiler, because mutating functions 
+can only have reference types as first argument.
```
anton-trunov, line 90: btw, let's change "reference types" to "type identifiers" because it's a misnomer in the codebase and the word "reference" in this context sounds *really* confusing

anton-trunov (approved):
Awesome work, just minor comments

src/test/e2e-emulated/contracts/semantics.tact
jeshecdom: Ohhh, good point. I never thought of putting a map inside a struct because I tested structs first. But now I see that I need to add tests that combine things that I previously tested.

src/test/e2e-emulated/contracts/semantics.tact
anton-trunov: Let's fix this minor comment and we are good to go

anton-trunov (changes requested):

anton-trunov (approved):
Let's fix that comment later, I'd like to merge these tests before we make a new release

PR #817, opened 2024-09-13
test: size limits of a single contract

### LINKED ISSUE #586: Add tests for various limits for data types, messages, etc.
For instance, the docs say:

> As the upper bound of the smart contract state size is around 65,000 items of type `Cell`, it constrains the storage limit of maps to be about 30,000 key-value pairs for the whole contract.

We should test doc statements like the above. 

So, we need tests for at least the following items:

- maximum map size
- maximum cell number
- maximum number of struct fields
- how many messages one contract can send (in a loop) before it hits the current 0.4 TON gas limit
- maximum length of a string

novusnota:
WIP:
- [x] (more or less done) map size limits
- [ ] max number of cells — there's an issue with Fift decompiler, blocking progress. Besides, the total limit from TON side is known: $2^{16}$ cells per account state, according to [config param 43](https://docs.ton.org/develop/howto/blockchain-configs#param-43)
- [ ] max number of struct fields

## Issue

Closes #586.

## Checklist


anton-trunov:
> Besides, the total limit from TON side is known

The point of the tests is to see
- if there a failure when we go beyond the limits and
- if we can go very close to the limits *without* a failure

anton-trunov:
ping

anton-trunov:
Changed the assignee of the corresponding issue, closing this now

PR #853, opened 2024-09-19
feat: `&&=`, `||=`, `>>=`, `<<=` operators

### LINKED ISSUE #811: Add logical assignment operators `&&=` and `||=`
It's a follow-up on #92 and #328.

Gusarich:
## Issue

Closes #811.

## Checklist


src/types/resolveStatements.ts
```diff
@@ -350,18 +383,7 @@ function processStatements(
...
-                        expressionType.kind !== "ref" ||
-                        expressionType.name !== "Int" ||
-                        expressionType.optional ||
-                        tailType.kind !== "ref" ||
-                        tailType.name !== "Int" ||
-                        tailType.optional
-                    ) {
+                    if (!isAugmentedAssignExpression(s, ctx)) {
                         throwCompilationError(
                             `Type error: Augmented assignment is only allowed for Int type`,
```
jeshecdom: Probably the error message needs to be changed to "Type error: Augmented assignment is only allowed for Int and Bool types". Although, I wonder if it would be possible to make it more specific like "Augmented assignment &&= is only allowed for Bool type", and "Augmented assignment += is only allowed for Int type".

anton-trunov (changes requested):
There is a bunch of `result = result && ...` expressions in `semantics.tact` -- would be nice to also dog-food here and refactor that code to use `&&=`.

Let's add a few negative tests, like `n &&= true`, where `n` is not a boolean, e.g. `Int`.

src/types/resolveStatements.ts
Gusarich: updated

src/types/__snapshots__/resolveStatements.spec.ts.snap
```diff
@@ -785,8 +785,38 @@ Line 5, col 5:
 "
 `;
 
+exports[`resolveStatements should fail statements for stmt-augmented-assgn-bool-int 1`] = `
+"<unknown>:9:5: Type error: Augmented assignment += is only allowed for matching types
```
anton-trunov: This one should also say:
```
Augmented assignment += is only allowed for Int types
```

src/types/__snapshots__/resolveStatements.spec.ts.snap
```diff
@@ -785,8 +785,38 @@ Line 5, col 5:
...
+Line 5, col 5:
+  4 |     let a: Int = 1;
+> 5 |     a &&= 1;
+          ^~~~~~~~
+  6 |     return a;
+"
+`;
+
+exports[`resolveStatements should fail statements for stmt-augmented-assgn-int-bool 1`] = `
+"<unknown>:9:5: Type error: Augmented assignment &&= is only allowed for matching types
```
anton-trunov: Augmented assignment &&= is only allowed for Bool types

src/types/__snapshots__/resolveStatements.spec.ts.snap
```diff
@@ -785,8 +785,38 @@ Line 5, col 5:
...
+Line 9, col 5:
+   8 | fun test(x: Bool): Bool {
+>  9 |     x += 1;
+           ^~~~~~~
+  10 |     return x;
+"
+`;
+
+exports[`resolveStatements should fail statements for stmt-augmented-assgn-int 1`] = `
+"<unknown>:5:5: Type error: Augmented assignment &&= is only allowed for Bool types
```
anton-trunov: should be singular, not plural
```suggestion
"<unknown>:5:5: Type error: Augmented assignment &&= is only allowed for Bool type
```

src/types/resolveStatements.ts
```diff
@@ -353,20 +353,42 @@ function processStatements(
...
+                        if (expressionType.name !== "Bool") {
+                            throwCompilationError(
+                                `Type error: Augmented assignment ${s.op}= is only allowed for Bool types`,
+                                s.loc,
+                            );
+                        }
+                    } else {
+                        if (expressionType.name !== "Int") {
+                            throwCompilationError(
+                                `Type error: Augmented assignment ${s.op}= is only allowed for Int types`,
```
anton-trunov:
```suggestion
                                `Type error: Augmented assignment ${s.op}= is only allowed for Int type`,
```

src/types/resolveStatements.ts
```diff
@@ -353,20 +353,42 @@ function processStatements(
...
+                        throwCompilationError(
+                            `Type error: Augmented assignment ${s.op}= is only allowed for matching types`,
+                            s.loc,
+                        );
+                    }
+
+                    if (s.op === "&&" || s.op === "||") {
+                        if (expressionType.name !== "Bool") {
+                            throwCompilationError(
+                                `Type error: Augmented assignment ${s.op}= is only allowed for Bool types`,
```
anton-trunov:
```suggestion
                                `Type error: Augmented assignment ${s.op}= is only allowed for Bool type`,
```

src/types/resolveStatements.ts
```diff
@@ -353,20 +353,42 @@ function processStatements(
...
                             s.loc,
                         );
                     }
+
+                    if (expressionType.name !== tailType.name) {
+                        throwCompilationError(
+                            `Type error: Augmented assignment ${s.op}= is only allowed for matching types`,
+                            s.loc,
+                        );
+                    }
```
anton-trunov: Basically, the operator should take priority in error messages. So, we first check if the l-value type and the operator do not match. And only if those match but the right-hand side expression does not, we complain about the expression type.

anton-trunov (changes requested):
Almost there, let's just fix some unintuitive error messages

src/types/resolveStatements.ts
```diff
@@ -353,17 +353,49 @@ function processStatements(
...
+                            throwCompilationError(
+                                `Type error: Augmented assignment ${s.op}= is only allowed for Int type`,
+                                s.loc,
+                            );
+                        }
+                    }
+
+                    if (expressionType.name !== tailType.name) {
+                        throwCompilationError(
+                            `Type error: Augmented assignment ${s.op}= is only allowed for matching types`,
```
anton-trunov: So, at this point we now the l-value and the operator are all good, so we need to complain about the type of the expression. We should be a bit more precise both in the error message and the location. Also, let's add a negative test for this error message.

For instance,

```
let x: Int = 42;
x += true;
```

`x: Int` and `+=` are matching, so we just report that `true` is expected to have type `Int` with the location pointer pointing at `true`, not at the whole statement.

anton-trunov (changes requested):
Just one small step

src/types/resolveStatements.ts
Gusarich: actually it seems that we'll never throw this error, because both types are being checked above already. have a look at the most recent commit.

anton-trunov (approved):

PR #854, opened 2024-09-19
feat: allow optional types for `self` argument in `extends mutates` functions

### LINKED ISSUE #70: Allow nullable types in "extends mutates" functions
TVM dictionaries are nullable cells. To work with them, methods need to accept `Cell?`, and it's currently impossible to mark such function "extends mutates", leading to constructions like `dict = udict_store_ref(dict, 256, "name".sha256(), ...);`.

Gusarich:
## Issue

Closes #70.

## Checklist

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -35,6 +35,7 @@
     "decompiling",
     "dentry",
     "Descr",
+    "DICTUSET",
     "disasm",
     "divmod",
     "dnsresolve",
--- a/cspell.json
+++ b/cspell.json
@@ -35,7 +35,6 @@
     "decompiling",
     "dentry",
     "Descr",
-    "DICTUSET",
     "disasm",
     "divmod",
     "dnsresolve",
--- a/src/test/e2e-emulated/contracts/mutating-methods.tact
+++ b/src/test/e2e-emulated/contracts/mutating-methods.tact
@@ -5,6 +5,13 @@ extends mutates fun multiply(self: Int, x: Int): Int {
     return self;
 }
 
+extends fun multiplyExtends(self: Int?, x: Int): Int? {
+    if (self == null) {
+        return null;
+    }
+    return self!! * x;
+}
+
 struct Foo {
     s: Slice;
 }
@@ -70,5 +77,11 @@ contract Tester {
     get fun test10(dict: Cell?): Cell? {
         dict.nativeUdictStoreUint(8, 123, rawSlice("456"));
         return dict;
+    }
+
+    get fun test11(x: Int?): Int? {
+        x.multiplyExtends(2);
+        x.multiplyExtends(2).multiplyExtends(3);
+        return x.multiplyExtends(2).multiplyExtends(3);
     }
 }
--- a/src/test/e2e-emulated/contracts/mutating-methods.tact
+++ b/src/test/e2e-emulated/contracts/mutating-methods.tact
@@ -1,3 +1,5 @@
+asm(value key self keySize) extends mutates fun nativeUdictStoreUint(self: Cell?, keySize: Int, key: Int, value: Slice) { DICTUSET }
+
 extends mutates fun multiply(self: Int, x: Int): Int {
     self *= x;
     return self;
@@ -63,5 +65,10 @@ contract Tester {
     get fun test9(): Int {
         self.s.loadUint(1);
         return self.s.loadUint(3);
+    }
+
+    get fun test10(dict: Cell?): Cell? {
+        dict.nativeUdictStoreUint(8, 123, rawSlice("456"));
+        return dict;
     }
 }
--- a/src/types/resolveDescriptors.ts
+++ b/src/types/resolveDescriptors.ts
@@ -747,7 +747,7 @@ export function resolveDescriptors(ctx: CompilerContext) {
         // Check virtual
         if (isVirtual) {
             if (self?.kind !== "ref") {
-                throwCompilationError(
+                throwInternalCompilerError(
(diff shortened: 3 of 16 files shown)


src/test/e2e-emulated/contracts/mutating-methods.tact
```diff
@@ -64,4 +66,9 @@ contract Tester {
         self.s.loadUint(1);
         return self.s.loadUint(3);
     }
+
+    get fun test10(dict: Cell?): Cell? {
+        dict.nativeUdictStoreUint(8, 123, rawSlice("456"));
```
anton-trunov, line 78: Let's check `nativeUdictStoreUint` can be chained with itself

anton-trunov (changes requested):
LGTM, just one minor testing suggestion

src/types/resolveExpression.ts
```diff
@@ -562,13 +562,6 @@ function resolveCall(
 
     // Handle ref
     if (src.kind === "ref") {
-        if (src.optional) {
```
jeshecdom, line 565: So, this change will allow optionals in not only `extends mutates` functions, but also in `extends` functions, right? Maybe we should add a test case of an `extends` function as well.

src/test/e2e-emulated/contracts/mutating-methods.tact
Gusarich: how and why should it be chained in this case? `nativeUdictStoreUint` returns `void` here, so it doesnt make sense.

src/types/resolveExpression.ts
Gusarich: resolved

src/test/e2e-emulated/contracts/mutating-methods.tact
anton-trunov: ah, yes, my bad

cspell.json
```diff
@@ -35,6 +35,7 @@
     "decompiling",
     "dentry",
     "Descr",
+    "DICTUSET",
```
anton-trunov:
```suggestion
```

src/types/test/fun-extends-opt-self.tact
```diff
@@ -0,0 +1,46 @@
...
+        return y;
+    }
+    return self!! + y;
+}
+
+extends mutates fun test_mutates(self: Int, y: Int?): Int {
+    if (y == null) {
+        return self;
+    }
+    self += y;
```
anton-trunov: `self += y` should not typecheck, but this should: `self += y!!`

since these tests do not typecheck the function bodies, let's just return some constants right away to avoid confusion

src/types/test/fun-extends-opt-self.tact
```diff
@@ -0,0 +1,46 @@
...
+    return self;
+}
+
+extends mutates fun test_mutates_self(self: Int?, y: Int): Int {
+    if (self == null) {
+        return y;
+    }
+    self!! += y;
```
anton-trunov: this assignment is also not allowed

src/types/test/fun-extends-opt-self.tact
```diff
@@ -0,0 +1,46 @@
...
+    return self;
+}
+
+extends mutates fun test_mutates_self_opt(self: Int?, y: Int): Int? {
+    if (self == null) {
+        return null;
+    }
+    self!! += y;
```
anton-trunov: this as well

src/types/resolveDescriptors.ts
```diff
@@ -746,7 +746,13 @@ export function resolveDescriptors(ctx: CompilerContext) {
 
         // Check virtual
         if (isVirtual) {
-            const t = types.get(self!)!;
+            if (self?.kind !== "ref") {
+                throwCompilationError(
+                    "Virtual functions must have a self parameter",
```
anton-trunov, line 751: Let's add a negative test with this error message

src/types/resolveDescriptors.ts
```diff
@@ -757,7 +763,13 @@ export function resolveDescriptors(ctx: CompilerContext) {
 
         // Check abstract
         if (isAbstract) {
-            const t = types.get(self!)!;
+            if (self?.kind !== "ref") {
+                throwCompilationError(
+                    "Abstract functions must have a self parameter",
```
anton-trunov, line 768: Let's add a negative test with this error message

src/types/resolveDescriptors.ts
```diff
@@ -767,7 +779,13 @@ export function resolveDescriptors(ctx: CompilerContext) {
         }
 
         if (isOverride) {
-            const t = types.get(self!)!;
+            if (self?.kind !== "ref") {
+                throwCompilationError(
+                    "Override functions must have a self parameter",
```
anton-trunov, line 784: Let's add a negative test with this error message

anton-trunov (changes requested):

src/types/resolveDescriptors.ts
Gusarich: Shouldn't these be `internalCompilerErrors` instead? Because those trait functions can't have a non-`ref` type kind, so we just add this condition here for typescript linter to pass

anton-trunov: I see, yeah, yeah it looks like those should be ICEs

jeshecdom (approved):
LGTM

anton-trunov (approved):

PR #855, opened 2024-09-20
feat: enhance `asm` functions to support all* of the Fift-asm syntax

### LINKED ISSUE #837: Better parser for `asm` block bodies
The parser should produce ASTs for `asm`-blocks. This will ensure we are in control of error messages and also we can later implement a type-checker for the `asm`-blocks.

### LINKED ISSUE #857: Current `asm` functions parser does not allow the full range of Fift-asm syntax
For example, any word/instruction with `}` in it aren't allowed and the `{ }` lists aren't processed.

novusnota:
Note, that while it's possible to support the syntax via Ohm without constucting ASTs, it's impossible to support the AST generation without outsourcing that part of the parser to the recursive-descent one, which would also keep track of dictionary of words/instructions and the TVM stack, at least approximating it. [See this for more info](https://github.com/tact-lang/tact/issues/837#issuecomment-2362498612).

P.S.: all* in the title means that we support everything, but with two caveats:
1. Out of place `}` and `{` (when shadowing them, for example) will break the parsing. However, the non-shadowing uses, lists `{ }` and any words containing `}` inside are allowed as long as they are not just a single `}`.
2. Usage of newly created words that use `word` may not parse well, as they might contain the `}` inside of them separated by whitespace, which means that it's another variant of the 1st caveat.

P.P.S: Essentially, our parser only has one flaw, which stems from `{ }` being used as a function block. If we were to use, say, `"""` or `{{{`, the problem would kind of mitigate itself. Just until someone defines the `"""` or `{{{` as a new word in their Fift program, and then we (and FunC), would inevitably fail to parse it. Until we have a [stack-keeping parser](https://github.com/tact-lang/tact/issues/837#issuecomment-2362498612), that is.

## Issue

Closes #857.

## Checklist


src/grammar/grammar.ohm
```diff
@@ -98,8 +98,26 @@ Tact {
...
+                     ("[" | "]" | "{" | "}" | "({)" | "(})") &#whiteSpace
+                     --partiallyDisallowDirectShadowing
+                   | "forget" &#whiteSpace asmInstruction --disallowForget
+                   | "(forget)" &#whiteSpace --disallowIndirectForget
+                   | asmInstruction --custom
+
+    // Instructions exclude word list begin and end words to ensure correct parse
+    asmInstruction = ~(("[" | "]" | "{" | "}" | "({)" | "(})") ~asmWord) asmWord
+
+    // A chuck of non-whitespace characters forms a word in Fift
```
```suggestion
    // A chunk of non-whitespace characters forms a word in Fift
```

anton-trunov (changes requested):
Wait, this PR definitely does not close the linked issue as there is no AST model of `asm`-instructions.

novusnota:
Changed the linked issue to #857. I'm willing to return to #837 later, preferably when we start moving to Tact 2.0 and a new parser for it. Alternatively, can tackle it after other remaining issues on 1.6.0 milestone for me to resolve (both here and for docs)

P.S.: On the positive side of #837, our `asm` ASTs will be typechecked right away :)

anton-trunov (changes requested):
We should definitely add more positive tests from some known Fift-projects.

Can we parse at least some part of `Asm.fif`? 

Some examples can be taken from this list: https://docs.ton.org/develop/smart-contracts/examples#fift-smart-contracts and this list: https://github.com/ton-blockchain/ton/tree/master/crypto/smartcont.

novusnota:
@anton-trunov added Fift library to the test suite and made sure that we don't create unnecessary snapshots for its contents

anton-trunov (approved):
Amazing job!

PR #856, opened 2024-09-20
feat: destructuring of structs and messages

### LINKED ISSUE #298: Simple destructuring of fields of structs and messages

Gusarich:
## Issue

Closes #298.

## Checklist


src/interpreter.ts
```diff
@@ -1409,6 +1413,53 @@ export class Interpreter {
...
+            throwErrorConstEval(
+                `destructuring assignment expected ${ast.identifiers.length} fields, but got ${
+                    Object.keys(val).length - 1
+                }`,
+                ast.loc,
+            );
+        }
+
+        for (const name of ast.identifiers) {
+            const v = val[idText(name)];
```
jeshecdom: Does this mean that in a statement like:
`let { a, b, c, d } = s;`
the variable names `a`, `b`, `c`, and `d` need to match the field names in `s`?

The following program will produce a compilation error, because the interpreter will be called in the return of `testFun2` in order to reduce the call to `testFun` to a value.
```
struct S {
    a: Int;
    b: Int;
    c: Int;
}

fun testFun(): Int {
    let s = S{ a: 1, b: 2, c: 3 };
    let { a1, b1, c1, d1 } = s;    // Variable names do not match field names in s
    return a1 + b1 + c1 + d1;
}

// If you remove testFun2, no compilation error occurs because testFun 
// will never be given to the interpreter unless some expression calls testFun
// somewhere in the program.

fun testFun2(): Int {
   return testFun();
}
```
I think type `StructValue` needs to be changed so that it not only stores the field names, but also the position number of each field in the struct. So that instead of doing this:
`const v = val[idText(name)];`
we could do something like this:
`const v = val.get(i);   // The i-th field in the struct`

src/generator/writers/writeFunction.ts
```diff
@@ -465,6 +465,13 @@ export function writeStatement(
 
             return;
         }
+        case "statement_destruct": {
+            ctx.append(
+                //`var ${resolveFuncTypeUnpack(t, funcIdOf(f.name), ctx)} = ${writeCastedExpression(f.expression, t, ctx)};`,
```
anton-trunov:
```suggestion
```

src/grammar/grammar.ohm
```diff
@@ -143,6 +144,8 @@ Tact {
 
     StatementForEach = foreach "(" id "," id "in" Expression ")" "{" Statement* "}"
 
+    StatementDestruct = let "{" ListOf<id, ","> ","? "}" "=" Expression (";" | &"}")
```
anton-trunov: We need to use an order-independent approach for struct fields, like this:

```tact
struct S {
  field1: Int;
  field2: Int;
}

let S { field1: a, field2: b } = struct;
// `a` and `b` are now bound to the `field1` and `field2` components respectively
```

and also using struct field punning should be allowed too:
```tact
let S { field1, field2 } = struct;
// `field1` and `field2` are bound to the respective struct fields
```

Notice that in the second example the order of fields is not relevant.

This prevents issues with reordering of struct fields in the struct definitions. The currently implemented design would provoke subtle bugs if the user changes the order of fields like so:

```
struct S {
  field2: Int;
  field1: Int;
}
```

Since the fields have the same type the refactoring which fails to change the order in the destructuring let-statements isn't going to be frowned upon by the typechecker.

anton-trunov (changes requested):
See my comment for the `grammar.ohm` file

Gusarich:
@jeshecdom @anton-trunov have a look at the reworked apporach please!

anton-trunov:
@Gusarich There are some merge conflicts

Gusarich:
> @Gusarich There are some merge conflicts

Resolved them, trying to push but knip doesn't let me because for some reason it thinks that `@types/jest` is not used. What should I do?

src/grammar/grammar.ohm
```diff
@@ -143,6 +144,11 @@ Tact {
 
     StatementForEach = foreach "(" id "," id "in" Expression ")" "{" Statement* "}"
 
+    StatementDestruct = let "{" ListOf<DestructItem, ","> ","? "}" "=" Expression (";" | &"}")
+
+    DestructItem = id ":" id --withMap
+                 | id --regular
```
```suggestion
    DestructItem = id ":" id --regular
                 | id --punned
```

src/grammar/ast.ts
```diff
@@ -569,6 +578,14 @@ export const selfId: AstId = {
     loc: dummySrcInfo,
 };
 
+export type AstDestructMapping = {
+    kind: "destruct_mapping";
+    from: AstId;
```
```suggestion
    field: AstId;
```

src/grammar/grammar.ohm
```diff
@@ -143,6 +144,11 @@ Tact {
 
     StatementForEach = foreach "(" id "," id "in" Expression ")" "{" Statement* "}"
 
+    StatementDestruct = let "{" ListOf<DestructItem, ","> ","? "}" "=" Expression (";" | &"}")
```
anton-trunov: We need the constructor name here:

```suggestion
    StatementDestruct = let typeId "{" ListOf<DestructItem, ","> ","? "}" "=" Expression (";" | &"}")
```

anton-trunov:
> Resolved them, trying to push but knip doesn't let me because for some reason it thinks that @types/jest is not used. What should I do?

hmmm, this is weird, because I removed that from ignoreDependencies (see the getter method id pr) precisely because I was getting warnings that it is useless

can you try updating your knip installation locally?

Gusarich:
> can you try updating your knip installation locally?

yeah, just had to run `yarn`. didn't notice the change in `package.json`. have a look now

src/grammar/ast.ts
```diff
@@ -320,6 +321,14 @@ export type AstStatementForEach = {
     loc: SrcInfo;
 };
 
+export type AstStatementDestruct = {
+    kind: "statement_destruct";
+    identifiers: AstDestructMapping[];
```
anton-trunov: Let's use a `Map` here, this will simplify some part of the code, like the codegen

src/grammar/test/stmt-destructuring.tact
```diff
@@ -0,0 +1,18 @@
...
+
+fun testFunc(): Int {
+    let s = S{ a: 1, b: 2, c: 3 };
+    let { a, b, c } = s;
+    let { a1, _, _ } = s;
+    let { _, b1, _ } = s;
+    let { _, _, c1 } = s;
+    let { a2, b2, _, } = s;
+    let { a3, _, c3, } = s;
+    let { _, b4, c4, } = s;
```
anton-trunov: these won't pass the scoping check, right?

src/test/e2e-emulated/contracts/structs.tact
anton-trunov, line 1: Let's also add destruct tests with different order of fields, like `let S {b, a} = s;` and also with fields mappings like `let S {b: b, a: a} = s;` and renames, especially confusing ones like `let S {b: a, a: b} = s;`

anton-trunov (changes requested):
I left a few comments and suggestions for the changes. Since this is a complex feature (yeah, codegen is quite simple, but the static semantics is a bit tricky), we need to test it more thoroughly, so let's add at least the following tests:
- positive interpreter (const-eval) tests (including various combinations of punned and mapped fields);
- negative interpreter tests;
- negative typechecker tests with the scoping rules violations, like `let S {x: x, y: x}` and shadowing of previously defined variables;
- negative typechecker tests with all the new error messages (that are non-ICE ones);
- pretty-printer tests;
- renamer tests;

anton-trunov:
@jeshecdom Please have a look at the interpreter changes and suggest more positive and negative tests. And also, please help with the var-scoping and typechecking rules review

anton-trunov:
1. Also, let's document this feature: look at how it's done in PR #922 ("available since").

2. And let's add some more negative tests for the typechecker in this style: `let S1 {x, y} = S2 {x: 42, y: 43}`

src/grammar/grammar.ohm
jeshecdom: Just to check that I understood correctly. The `S` in the following declaration is just the type of variable `s1`:
```
let S {field1: a, field2: b} = s1;
```
and this will create two variables `a` and `b`, with `a` having `s1.field1` and `b` with `s1.field2`. So, is the `S` really necessary if it could be inferred from the type of `s1`? (But I agree that it is better to be explicit).

anton-trunov: yes, you understand correctly how variables are bound

> So, is the S really necessary if it could be inferred from the type of s1?

We are just following a general syntactical principle here: pattern-matching should correspond to building something. And all our structs need the constructor to be specified. Rust, for instance, uses the same approach for destructuring assignments

src/grammar/test/stmt-destructuring.tact
Gusarich: why?

Gusarich: yes, I forgot to update this one after the rework

src/interpreter.ts
```diff
@@ -1438,6 +1442,53 @@ export class Interpreter {
...
+                    `destructuring assignment expected field ${idTextErr(
+                        item.from,
+                    )}`,
+                    ast.loc,
+                );
+            }
+            this.envStack.setNewBinding(idText(item.name), v);
+        }
+    }
+
```
jeshecdom, line 1494: The interpreter code looks good. For positive tests, I suggest adding tests of the following form:
```
struct S {
  a: Int;
  b: Int;
  c: Int; 
}

fun testFun1(): Int {
    let s = S { a: 1, b: 2, c: 3 };
    let S { c: e, b, a } = s;
    return a + b + e;
}

fun testFun2(): S {
    let s = S { a: 1, b: 2, c: 3 };
    let S { a, c: e, b: b } = s;
    return S {a: e, b: b, c: a};
}

contract TestContract {
    get fun test1(): Int {
        return testFun1();
    }
    
    get fun test2(): S {
        return testFun2();
    }
}
```
In other words, just repeat the tests you already have, but this time call the functions in a getter, since this will force the execution of the interpreter in the function calls inside the getters.

Negative tests for the interpreter are more difficult to write because the errors are caught by the typechecker before they arrive to the interpreter. I have plans for testing the interpreter in isolation, but for the moment, just make sure that in the negative tests the typechecker gets errors for:
- Constant shadowing:
   ```
   const c1: Int = 5;
   .........
   let S { a: c1, b, c } = s;   // c1 is shadowing constant c1
   ```
- Destructuring of non-struct value:
   ```
   let S { a, b, c } = i;    // where i is an integer
   ```
- Missing and extra fields on both sides of the equality (which I think this is already tested):
   ```
   struct A {
     f1: Int;
     f2: Int;
   }

   struct B {
      f1: Int;
      f2: Int;  
   }
   ....... Take each line as an independent test (also, repeat the tests using non-punned versions):
   let A {f1} = s  // s has type A
   let A {f1, f2} = A {f1: 2};
   let A {f1, f2} = s  // s has type B, even though they have the same fields, they differ in types
   let A {f1, f2, f3} = s  // s has type A
   let A {f1, f2} = A {f1: 2, f3: 6};
   ```
- Plus all tests suggested by Anton.

The interpreter makes the assumption that the code it receives has already passed thorough the typechecker. Otherwise, we would repeat a lot of code.

anton-trunov: I would also initialize some constants using functions with `let`-destructuring

novusnota:
> Also, let's document this feature: look at how it's done in PR https://github.com/tact-lang/tact/pull/922 ("available since").

If needed, I can quickly help write the docs once all the other suggestions above are resolved :)

src/grammar/ast.ts
Gusarich: done

src/grammar/test/stmt-destructuring.tact
Gusarich: done

src/test/e2e-emulated/contracts/structs.tact
Gusarich: done

Gusarich:
@anton-trunov @jeshecdom I think I've made all the changes you requested in the review. Please have a look!

Discovered and fixed a few small bugs while writing all these tests 👍

jeshecdom (commented):
LGTM, @anton-trunov

src/generator/writers/writeFunction.ts
```diff
@@ -465,6 +465,26 @@ export function writeStatement(
...
+                    `invalid destruct expression kind: ${t.kind}`,
+                    f.expression.loc,
+                );
+            }
+            const ty = getType(ctx.ctx, t.name);
+            const ids = ty.fields.map((field) => {
+                const id = Array.from(f.identifiers.entries()).find(
+                    ([key]) => key.text === field.name,
+                );
+                return id ? funcIdOf(id[1]) : "_";
```
anton-trunov: I'd say we should also translate Tact's `_` into the same underscore in FunC.

So, for instance, for the following snippet

```tact
struct S { x: Int; y: String; z: Bool }

let S { x: _, y } = ...;
```
we get underscores in the resulting FunC code both for `x` and `z`.

src/test/e2e-emulated/contracts/structs.tact
```diff
@@ -245,6 +245,79 @@ struct OptionalFields {
...
+    return (a ? 1 : 0) + b;
+}
+
+fun destructuringTest2(): Int {
+    let s = S {
+        a: true,
+        b: 42
+    };
+
+    let S {_, b} = s;
```
anton-trunov: wait, `let S {_, b} = s` should be a type error: `_` means `_: _` and there is no field `_`

src/types/__snapshots__/resolveStatements.spec.ts.snap
```diff
@@ -898,6 +898,106 @@ Line 5, col 5:
...
+Line 15, col 22:
+  14 |     let s = S{ a: 1, b: 2, c: 3 };
+> 15 |     let S { a: x, b: x, c: y } = s;
+                            ^
+  16 |     return x + y;
+"
+`;
+
+exports[`resolveStatements should fail statements for stmt-destructuring-fields-non-destructable 1`] = `
+"<unknown>:16:25: Type 'map<Int, Int>' is not destructurable
```
anton-trunov: lets make the error messages more uniform: "... cannot be destructured"

src/types/__snapshots__/resolveStatements.spec.ts.snap
```diff
@@ -898,6 +898,106 @@ Line 5, col 5:
...
+Line 15, col 25:
+  14 |     let s: S? = S{ a: 1, b: 2, c: 3 };
+> 15 |     let S { a, b, c } = s;
+                               ^
+  16 |     return a + b + c;
+"
+`;
+
+exports[`resolveStatements should fail statements for stmt-destructuring-fields-non-destructable2 1`] = `
+"<unknown>:16:29: Type 'Test' is not destructurable
```
anton-trunov: lets make the error messages more uniform: "... cannot be destructured"

src/grammar/grammar.ts
```diff
@@ -1012,6 +1013,31 @@ semantics.addOperation<AstNode>("astOfStatement", {
...
+        _semicolon,
+    ) {
+        return createAstNode({
+            kind: "statement_destruct",
+            type: typeId.astOfType(),
+            identifiers: identifiers
+                .asIteration()
+                .children.reduce((map, item) => {
+                    const destructItem = item.astOfExpression();
+                    map.set(destructItem.field, destructItem.name);
```
anton-trunov: looks like the following 
```tact
struct S { x: Int; y: String; z: Bool }
let S { x: a, x: b } = s;
```

will get past the typechecker (but it should trigger an error)

src/grammar/ast.ts
```diff
@@ -320,6 +321,15 @@ export type AstStatementForEach = {
     loc: SrcInfo;
 };
 
+export type AstStatementDestruct = {
+    kind: "statement_destruct";
+    type: AstTypeId;
+    identifiers: Map<AstId, AstId>;
```
anton-trunov: hmmm, maps from `AstId` makes it sort of unusable (you cannot use it with `.get` methods and field descriptions)

anton-trunov (changes requested):

src/test/e2e-emulated/contracts/structs.tact
novusnota: Oh, so if the user only wants `b`, they should write `let S {b} = s;`, right?

src/grammar/ast.ts
Gusarich: Yes, but you suggested using maps in the previous review. I thought about making it as `Map<string, string>` but it'll miss AST ids this way

src/generator/writers/writeFunction.ts
Gusarich: Isn't it like this already?

src/test/e2e-emulated/contracts/structs.tact
anton-trunov: correct

anton-trunov: or just use `.b`

src/grammar/ast.ts
anton-trunov: yeah, we should keep the location info, but also should make it usable as a map.
looks like a custom map-like data type with a `get` method that can work on strings could be used here

src/generator/writers/writeFunction.ts
anton-trunov: nope, for Tact's `_` you will get `$_`

Gusarich:
@anton-trunov resolved everything, have a look please!

src/grammar/ast.ts
```diff
@@ -320,6 +321,15 @@ export type AstStatementForEach = {
     loc: SrcInfo;
 };
 
+export type AstStatementDestruct = {
+    kind: "statement_destruct";
+    type: AstTypeId;
+    identifiers: Map<string, [AstId, AstId]>; // field name -> [field id, local id]
```
novusnota:
```suggestion
    /** field name -> [field id, local id] */
    identifiers: Map<string, [AstId, AstId]>;
```

A tiny change to make that comment show up in hover info

anton-trunov (approved):

anton-trunov:
> If needed, I can quickly help write the docs once all the other suggestions above are resolved :)

@novusnota your turn :)

novusnota:
@anton-trunov on it!

UPD: https://github.com/tact-lang/tact/pull/964

PR #917, opened 2024-10-06
fix(docs/ci): compilation of examples in `data-structures.mdx` and across Cookbook

### LINKED ISSUE #405: `data-structures.mdx`: Code cannot be compiled
The recent version of the compiler rejects the following code:

https://github.com/tact-lang/tact-docs/blob/d285e25dc2a906cfb43e743b8f8430fd10c480de/pages/cookbook/data-structures.mdx?plain=1#L254

```
Syntax error: src/routes/(examples)/04-arrays/contract.tact:47:13: Parse error: expected not a reservedWord

Line 47, col 13:
  46 |
> 47 |     get fun map(): map<Int, Int> {
                   ^
  48 |         return self.arr;
```

novusnota:
## Issue

Closes tact-lang/tact-docs#405.

## Checklist

## CI

image

### DIFF
--- a/.github/workflows/tact-docs-prod.yml
+++ b/.github/workflows/tact-docs-prod.yml
@@ -6,6 +6,7 @@ on:
   # The check for the 'main' branch is done on the job level
   push:
     tags: ["v[0-9]+.[0-9]+.[0-9]+"]
+    branches: ["main"]
 
   # Allows to run this workflow manually from the Actions tab
   workflow_dispatch:
@@ -29,8 +30,9 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-        # with:
-        #   fetch-depth: 0
+        with:
+          fetch-depth: 0
+
       - name: Install, build and store site artifact
         uses: withastro/action@v3
         with:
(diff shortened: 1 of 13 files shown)


jubnzv (commented):
Do we need to change the getter name?

```tact
// Getter function for obtaining the stack
    get fun map(): map<Int, Int> {
        return self.stack.map;
    }
```

anton-trunov (dismissed):

anton-trunov:
btw, all cookbook examples should be tested in CI, so the source code of the recipes should be either inserted from files into the docs or extracted from docs and compiled separtely

novusnota:
> btw, all cookbook examples should be tested in CI, so the source code of the recipes should be either inserted from files into the docs or extracted from docs and compiled separtely

@anton-trunov implemented just that and discovered a lot of somewhat broken code along the way! All fixed now and the script for checking examples from Cookbook by extracting code blocks off of their pages is now in the CI :)

Also did make tiny enhancements along the way. And enabled the CI for the docs after the successful merge to main (but not before!) — prior to this it would only trigger on pushes of tags to the main OR manually (which I did a lot).

anton-trunov (approved):

PR #922, opened 2024-10-07
feat: specify method ID for getters

### LINKED ISSUE #876: Allow manually specifying `method_id` for getters
To let users still keep their getter names in case of collisions like in issue #859.

I'm proposing the following syntax for this:
```tact
get(0x42) fun bestTelegramChannel(): String {
  return "@gusarich_thoughts";
}
```

`get(method_id)` syntax should be consistent with what we do for message opcodes: `message(0x42)`, however let's discuss alternative proposals.

`method_id` should be an arbitrary compile-time expression of type `Int`, so something like this should be possible too:

```tact
get(crc32("foobar")) fun bestTelegramChannel(): String {
  return "@gusarich_thoughts";
}
```

anton-trunov:
## Issue

Closes #876.

## Related issues

After something like https://github.com/ton-org/ton-core/pull/46 is merged and released, we can remove `as any` from the TypeScript wrappers generator

## Checklist


anton-trunov:
@novusnota I'd need your suggestion on how to make `available since Tact v1.6` work here.

novusnota:
@anton-trunov right away I see couple of options, each using [Badge component](https://starlight.astro.build/components/badges/):

1. For the particular description of the new feature, we add a sub-heading and right after it we place:

```md
### Resolving collisions of method IDs

<Badge text="Available since Tact 1.6" variant="tip" size="medium" /><p/>

Sometimes this can get you the same method ID...(and the following, with a slight re-write of the beginning)
```

2. Alternatively, we place a full stop in this sentence: "If this happens, you can either rename some of the contract's getters or specify the getter's method ID manually.". Then, we place the badge as before, and continue with the example. Code for example can also start with `// runnable since Tact 1.6`

src/types/resolveDescriptors.ts
```diff
@@ -1947,25 +1947,6 @@ export function resolveDescriptors(ctx: CompilerContext) {
         staticConstants.set(idText(a.name), buildConstantDescription(a));
     }
 
-    // Check for collisions in getter method IDs
```
jeshecdom, line 1950: LGTM, just one question: Is there a reason as to why this deleted code was moved to `resolveStatements.ts`?

anton-trunov: Yep, we need contract constants to work in method IDs too. And the `self` parameter only enters the typing context before we typecheck a function's body in `resolveStatemtens.ts`. This is why the code for type-checking a getter's method id is placed *after* introducing `self` into the context and *before* adding the function parameters.

jeshecdom: Got it.

anton-trunov: btw, my guess is that #924 needs to be fixed in the same way

jeshecdom (approved):

PR #938, opened 2024-10-10
feat(docs): `map.exists()`

### LINKED ISSUE #897: Document `map.exists()`
https://github.com/tact-lang/tact/pull/581

novusnota:
## Issue

Closes #897.

## Checklist


docs/src/content/docs/book/maps.mdx
```diff
@@ -133,6 +135,26 @@ fizz = null; // identical to the previous line, but less descriptive
 
 With this approach all previous entries of the map are completely discarded from the contract even if the map was declared as its persistent state variable. As a result, assigning maps to `emptyMap(){:tact}` **does not** inflict any hidden or sudden [storage fees](https://docs.ton.org/develop/smart-contracts/fees#storage-fee).
 
+### Check if entry exists, `.exists()` {#exists}
+
+<Badge text="Available since Tact 1.5" variant="tip" size="medium"/><p/>
+
+The `.exists(){:tact}` [method](/book/functions#extension-function) on maps returns `true{:tact}` if the value under the given key exists in the map and `false{:tact}` otherwise.
```
anton-trunov, line 142: Let's also add a comment saying something like: `m.exists(key)` is equivalent to `m.get(key) != null`. I think it is also supposed to be more gas efficient, but if you could double check it, that would be greatly appreciated

anton-trunov (changes requested):

docs/src/content/docs/book/maps.mdx
novusnota: It's slightly more gas efficient to use `m.exists(key)` indeed, just checked

anton-trunov (approved):
Awesome! Thanks

PR #939, opened 2024-10-10
feat(docs): `map.deepEquals()`

### LINKED ISSUE #898: Document `map.deepEquals`
https://github.com/tact-lang/tact/pull/637#issuecomment-2313189568:

> We should also note in documentation that this function is very expensive in terms of gas (iterating a map + using `slice_hash`, both are expensive things), and in most cases it'll be enough to use hash comparison via `==`.

novusnota:
Didn't go too deep into describing how the long and short key labels might be used to manually construct the HashmapE (FunC's dict) as it's way too low-level for the general audience. But I did link the TL-B page which showcases all the intricacies of that process for the curious :)

## Issue

Closes #898.

## Checklist


docs/src/content/docs/book/maps.mdx
```diff
@@ -151,6 +153,29 @@ if (fizz == null) {
 }
 ```
 
+### Compare with `.deepEquals()` {#deepequals}
+
+
+The `.deepEquals(){:tact}` [method](/book/functions#extension-function) on maps returns `true{:tact}` if all the entries of the map match respective entries of another map and their [underlying serialization logic](https://docs.ton.org/develop/data-formats/tl-b-types#hashmap) is exactly the same. Returns `false{:tact}` otherwise.
```
anton-trunov: Actually, it's the other way around: it returns true even if the serializations are different (as opposed to equality comparisons via hashes). This is called extensional equality. You don't have to explain everything at the level of TL-B, but an example (from the corresponding tests) is definitely needed here to explain the use case of deepEquals. We can also add a note, that if all maps are formed in a controlled way the hash comparisons are fine, but if a map comes from an external sources that does not provide those guarantees, one needs to use deepEquals

anton-trunov (changes requested):

docs/src/content/docs/book/maps.mdx
novusnota: Addressed the suggestions :)

anton-trunov (approved):
LGTM! Thanks

PR #941, opened 2024-10-11
feat: `replace` and `replaceGet` map methods & refactor of codegen and testing

### LINKED ISSUE #538: Replace map operations
We should cover the `_replace` and `_replaceget` families of map operations

Gusarich:
## Issue

Closes #538.

## Checklist


src/abi/map.ts
```diff
@@ -647,4 +647,398 @@ export const MapFunctions: Map<string, AbiFunction> = new Map([
...
         },
     ],
+    [
+        "replace",
+        {
+            name: "replace",
+            resolve(ctx, args, ref) {
+                // Check arguments
+                if (args.length !== 3) {
+                    throwCompilationError("replace expects two arguments", ref); // Should not happen
```
anton-trunov: this can actually happen for `m.replace(42)`

```suggestion
                    throwCompilationError("replace expects two arguments", ref);
```

src/abi/map.ts
anton-trunov, line 1: There is a lot of almost identical code for codegen in this file for the map mutation operations, let's refactor it into some common part parameterized with the operation at hand: this seems to be possible for least the `set`, `replace` and `replaceGet` methods (`get` and `del` are also quite similar, so if you can do it, then it would be nice to have)

anton-trunov: Typechecking is also pretty similar across those functions, but it's more or less straightforward, so let's skip it for now if it's hard

anton-trunov (changes requested):
Cool stuff! Let's do the suggested refactoring. And also add some simple docs, unless @novusnota is willing to volunteer to help with that

novusnota:
> unless @novusnota is willing to volunteer to help with that

Yup, will do ~~tomorrow~~ soon, aside from other doc issues :)

src/test/e2e-emulated/contracts/maps.tact
```diff
@@ -342,6 +512,136 @@ contract MapTestContract {
         self.intMap10_4.del(msg.key);
     }
 
+    receive(msg: ReplaceIntMap1) {
+        self.intMap1.replace(msg.key, msg.value);
```
jeshecdom: At first I thought that `replace` and `replaceGet` had signatures `(map<K, V>, K, V) -> Bool` and `(map<K, V>, K, V) -> V?`, respectively, according to the conversation in issue #538. 

But then, looking into the `map.spec.ts` file, I see that `self.intMap1.replace(k, null)` actually **deletes** the key `k`. The alternative interpretation, where `k` gets updated to `null` but not deleted, does not make sense because the type of the map is `map<Int,Int>`.

So, Is the following interpretation correct?
`m.replace(k, null)`: Delete key `k` if currently exists; do nothing otherwise. Return `true` if `k` exists, `false` otherwise.
`m.replaceGet(k, null)`: Delete `k` if currently exists; do nothing otherwise. Return the old value of `k` if key `k` exists, `null` otherwise. 

And when `v` is not `null`:
`m.replace(k, v)`: Replace value of `k` with `v` if `k` currently exists; do nothing otherwise. Return `true` if `k` exists, `false` otherwise.
`m.replaceGet(k, v)`: Replace value of `k` with `v` if `k` currently exists; do nothing otherwise. Return the old value of `k` if `k` exists, `null` otherwise.

src/test/e2e-emulated/map.spec.ts
```diff
@@ -1307,4 +1307,2282 @@ describe("map", () => {
...
+                expect((await contract.getIntMap10_2()).size).toBe(1);
+                expect((await contract.getIntMap10_3()).size).toBe(1);
+                expect((await contract.getIntMap10_4()).size).toBe(1);
+                expect((await contract.getAddrMap1()).size).toBe(1);
+                expect((await contract.getAddrMap2()).size).toBe(1);
+                expect((await contract.getAddrMap3()).size).toBe(1);
+                expect((await contract.getAddrMap4()).size).toBe(1);
+
+                // Check .replace return value
+                expect(await contract.getIntMap1Replace(k, null)).toBe(true);
```
jeshecdom: So, this will remove key `k` from map `IntMap1`, right?

src/test/e2e-emulated/map.spec.ts
```diff
@@ -1307,4 +1307,2282 @@ describe("map", () => {
...
+                expect(await contract.getIntMap1Replace(k + 1n, null)).toBe(
+                    false,
+                );
+                expect(await contract.getIntMap1Replace(k, 123n)).toBe(true);
```
jeshecdom: But then, why this returns `true` if key `k` was not in the map `IntMap1`? Since `k` is not in the map, `replace` will do nothing and will return `false`.

src/test/e2e-emulated/map.spec.ts
```diff
@@ -1307,4 +1307,2282 @@ describe("map", () => {
...
+                expect(await contract.getIntMap1Replace(k + 1n, 123n)).toBe(
+                    false,
+                );
```
jeshecdom: I think that further tests need to be added here for testing the returned value of `replace` (not using a `receive`, because those cannot test the returned value of `replace`), similar to the tests at lines 1682-1689.

The tests at line 1682-1689 tested:
- The returned value of a `replace(k, null)` when `k` is in the map and so, `k` gets deleted and `replace` returns `true`. A follow up test should be added `get(k)` that should return `null` to confirm that the `k` is deleted.
- The returned value of a `replace(k, null)` when `k` is not in the map and never was in the map. A follow up test should be added `get(k)` to confirm that it returns `null`.
- The returned value of a `replace(k, v)` (for `v` not null) when `k` is not in the map because it was previously deleted by `replace(k, null)`. This returned `true` (which is puzzling to me, because `k` is not in the map). An alternative test should be added in which a key is deleted using `del(k)` and immediately followed by `get(k)` (which returns `null`), then followed by `replace(k, v)` (which should return `false`), and then followed by `get(k)` (which should return `null`).
- The returned value of a `replace(k, v)` (for `v` not null) when `k` is not in the map and never was in the map. This returned `false`.
- A test that is missing is: `replace(k, v)` (for `v` not null) when `k` is actually in the map. This should return `true`.

In theory, similar tests should be added for the rest of maps, not just for `IntMap1`.

src/test/e2e-emulated/map.spec.ts
```diff
@@ -1307,4 +1307,2282 @@ describe("map", () => {
...
+                expect((await contract.getAddrMap4()).size).toBe(1);
+
+                // Check .replaceGet return value
+                expect(await contract.getIntMap1ReplaceGet(k, null)).toBe(
+                    valueInt,
+                );
+                expect(await contract.getIntMap1ReplaceGet(k + 1n, null)).toBe(
+                    null,
+                );
+                expect(await contract.getIntMap1ReplaceGet(k, 123n)).toBe(
```
jeshecdom: Again, similar comment as with `replace`: if line 2803 above deletes key `k`, why this does not return `null`?

src/test/e2e-emulated/map.spec.ts
```diff
@@ -1307,4 +1307,2282 @@ describe("map", () => {
...
+                    valueInt,
+                );
+                expect(await contract.getIntMap1ReplaceGet(k + 1n, 123n)).toBe(
+                    null,
+                );
```
jeshecdom: Similar comment here as with `replace`. Similar follow up tests should be added here, and a positive test with `replaceGet(k,v)` with `v` not null and `k` existing in the map.

src/test/e2e-emulated/contracts/maps.tact
Gusarich: yes, correct

src/test/e2e-emulated/map.spec.ts
Gusarich: yes

src/test/e2e-emulated/map.spec.ts
Gusarich: these are get method calls - they don't alter the state of the smart contract, so `k` is still in map here

src/test/e2e-emulated/map.spec.ts
Gusarich: yes, makes sense, I'll add tests for all maps

src/test/e2e-emulated/map.spec.ts
Gusarich: same reason as for the `replace` (answered above)

src/test/e2e-emulated/map.spec.ts
Gusarich: 👍

src/abi/map.ts
Gusarich: have a look at the latest commit with suggested refactor

Gusarich: @anton-trunov have a look at the latest commit with suggested refactor

anton-trunov: yeah, looks like we are going in the right direction! let's add those tests you discussed with @jeshecdom and merge this PR

Gusarich:
@anton-trunov @jeshecdom have a look please!

map tests are now extremely large and slow so I want to refactor them completely in a separate PR after this one gets merged.
Well, I can do that in this one too if you want it that way

src/test/e2e-emulated/map.spec.ts
```diff
@@ -1307,4 +1307,6345 @@ describe("map", () => {
...
+            expect((await contract.getIntMap9_6()).size).toBe(0);
+            expect((await contract.getIntMap10_1()).size).toBe(0);
+            expect((await contract.getIntMap10_2()).size).toBe(0);
+            expect((await contract.getIntMap10_3()).size).toBe(0);
+            expect((await contract.getIntMap10_4()).size).toBe(0);
+            expect((await contract.getAddrMap1()).size).toBe(0);
+            expect((await contract.getAddrMap2()).size).toBe(0);
+            expect((await contract.getAddrMap3()).size).toBe(0);
+            expect((await contract.getAddrMap4()).size).toBe(0);
+            expect((await contract.getAddrMap6_1()).size).toBe(0);
```
jeshecdom: Is there a reason for the missing getAddrMap5?

src/test/e2e-emulated/map.spec.ts
```diff
@@ -1307,4 +1307,6345 @@ describe("map", () => {
...
+                expect((await contract.getIntMap9_5()).size).toBe(1);
+                expect((await contract.getIntMap9_6()).size).toBe(1);
+                expect((await contract.getIntMap10_1()).size).toBe(1);
+                expect((await contract.getIntMap10_2()).size).toBe(1);
+                expect((await contract.getIntMap10_3()).size).toBe(1);
+                expect((await contract.getIntMap10_4()).size).toBe(1);
+                expect((await contract.getAddrMap1()).size).toBe(1);
+                expect((await contract.getAddrMap2()).size).toBe(1);
+                expect((await contract.getAddrMap3()).size).toBe(1);
+                expect((await contract.getAddrMap4()).size).toBe(1);
```
jeshecdom: Is there a reason for the missing getAddrMap 5, etc.? This happens in several places below as well.

Gusarich: just missed these randomly, will make sure to not miss anything in the refactor (it'll be much easier due to less repetitive code and better structure)

src/test/e2e-emulated/map.spec.ts
Gusarich: same as below

src/generator/writers/writeStdlib.ts
anton-trunov, line 1: I did some refactoring on this file (haven't finished yet though):

For instance, this is how all dict get operations are represented

```ts
const keyTypes = ["slice", "uint", "int"] as const;
type KeyType = (typeof keyTypes)[number];

const valTypes = ["slice", "int", "uint", "cell"] as const;
type ValType = (typeof valTypes)[number];

function genTactDictGet(
    ctx: WriterContext,
    key: KeyType,
    value: ValType,
): void {
    const signatureKeyType = key === "uint" ? "int" : key;
    const signatureRetType = value === "uint" ? "int" : value;
    let dictGet = "";
    const cellSuffix = value === "cell" ? "_ref" : "";
    switch (key) {
        case "slice":
            dictGet = `__tact_dict_get${cellSuffix}`;
            break;
        case "uint":
            dictGet = `udict_get${cellSuffix}?`;
            break;
        case "int":
            dictGet = `idict_get${cellSuffix}?`;
            break;
    }
    let returnExpr = "";
    switch (value) {
        case "slice":
        case "cell":
            returnExpr = "r";
            break;
        case "uint":
            returnExpr = "r~load_uint(vl)";
            break;
        case "int":
            returnExpr = "r~load_int(vl)";
            break;
    }
    let valBitsArg = "";
    switch (value) {
        case "slice":
        case "cell":
            valBitsArg = "";
            break;
        case "uint":
        case "int":
            valBitsArg = ", int vl";
            break;
    }
    ctx.fun(`__tact_dict_get_${key}_${value}`, () => {
        ctx.signature(
            `${signatureRetType} __tact_dict_get_${key}_${value}(cell d, int kl, ${signatureKeyType} k${valBitsArg})`,
        );
        ctx.flag("inline");
        ctx.context("stdlib");
        ctx.body(() => {
            if (dictGet.startsWith("__tact")) ctx.used(dictGet);
            ctx.write(`
                var (r, ok) = ${dictGet}(d, kl, k);
                if (ok) {
                    return ${returnExpr};
                } else {
                    return null();
                }
            `);
        });
    });
}
```

and to actually generate all the versions I used the following:

```
    // generate all dict `get` operations
    for (const key of keyTypes) {
        for (const val of valTypes) {
            genTactDictGet(ctx, key, val);
        }
    }
```

This way it should be easier to add `as coins`, for instance. Let's try refactoring the replace functions in this PR

anton-trunov: Related PR: #993

Gusarich:
@anton-trunov @jeshecdom have a look at the final tests please!

not added a few special cases from the original implementation yet, but all the operations are tested now on multiple sets of key-value pairs. just want to make sure to include all the important scenarios.

src/test/e2e-emulated/map.spec.ts
```diff
@@ -1,1310 +1,2139 @@
...
+            mapConfigs.forEach(({ mapName }) => {
+                const actualValue = getResponse[mapName];
+                expect(actualValue).toBeNull();
+            });
+        }
+    });
+
+    it("get: should retrieve multiple values after setting them", async () => {
+        // Set multiple values
+        for (const { keys, values } of testCases) {
```
jeshecdom, line 921: I am confused regarding the purpose of this test. This for-statement will overwrite the keys in each map several times, so that at the end there is only one value in each key of each map, right? So what "retrieve multiple values after setting them" means here? Could you help me understand this test?

jeshecdom (commented):
The tests are awesome! I just added a small comment below.

src/test/e2e-emulated/map.spec.ts
Gusarich: > This for-statement will overwrite the keys in each map several times, so that at the end there is only one value in each key of each map, right?

no, because keys in testCases are not overlapped and this for-loop just writes all the key-value pairs and then we check that all of them are set correctly

jeshecdom: Ohhhh, I missed that =) 👍

docs/src/content/docs/book/maps.mdx
```diff
@@ -96,6 +96,90 @@ if (gotButUnsure != null) {
...
+// Setting a couple of values under different keys
+fizz.set(7, 70);
+fizz.set(42, 42);
+
+// Overriding one of the existing key-value pairs
+let replaced1 = fizz.replace(7, null); // the entry under key 7 is now deleted
+replaced1; // true
+
+// Trying to replace the value in a non-existing key-value pair will do nothing
+let replaced2 = fizz.replace(8, null); // no key 8, so nothing was altered
```
jeshecdom, line 137: Maybe, instead of using `fuzz.replace(8, null)` here, use key `7` to emphasize that such key is no longer in the map `fizz.replace(7, null);`

The previous examples already showed the example of key `8` which was not originally inserted into the map.

What do you think?

docs/src/content/docs/book/maps.mdx
```diff
@@ -96,6 +96,90 @@ if (gotButUnsure != null) {
...
+// Setting a couple of values under different keys
+fizz.set(7, 70);
+fizz.set(42, 42);
+
+// Overriding one of the existing key-value pairs
+let oldVal1 = fizz.replaceGet(7, null); // the entry under key 7 is now deleted
+oldVal1; // 70
+
+// Trying to replace the value in a non-existing key-value pair will do nothing
+let oldVal2 = fizz.replaceGet(8, null); // no key 8, so nothing was altered
```
jeshecdom, line 179: The same comment here: use key `7`.

jeshecdom (commented):
Tests LGTM.

docs/src/content/docs/book/maps.mdx
anton-trunov: Let's merge as it is and then improve our map testing approach in subsequent PRs

anton-trunov (approved):

PR #956, opened 2024-10-14
feat(docs): add stonfi cookbook

### LINKED ISSUE #149: An example of sending a Ston.fi swap transaction in Cookbook
Related issue: #146 

Also, it would be nice to have an example of adding liquidity (as suggested by @Gusarich)

### LINKED ISSUE #955: An example of sending a Ston.fi swap transaction in Cookbook
Clone of https://github.com/tact-lang/tact-docs/issues/149

a-bahdanau:
## Issue

Closes #955.
Closes https://github.com/tact-lang/tact-docs/issues/149.

## Checklist


anton-trunov:
@novusnota @Gusarich please review this PR

docs/src/content/docs/cookbook/dexes/stonfi.mdx
```diff
@@ -5,10 +5,362 @@ sidebar:
...
+    amount: Int as coins;
+    destination: Address;
+    responseDestination: Address?;
+    customPayload: Cell? = null;
+    forwardTonAmount: Int as coins;
+    forwardPayload: Cell?; // hack for now, equivalent to Slice as remaining with first bit set to 1
+}
+```
+
+Below are shown gas constants. All of them are taken from the sdk [STON.fi sdk](https://github.com/ston-fi/sdk). They are slightly differ from reality, but it is ok for now.
```
Gusarich: Why do they differ from reality?

And overall, this isn't a good statement to have in documentation. Let's explain what these mean, why they are hardcoded in toncoin, note that best practice is to calculate fees dynamically using current config params and leave a related docs page link.

docs/src/content/docs/cookbook/dexes/stonfi.mdx
```diff
@@ -5,10 +5,362 @@ sidebar:
...
+struct ProvideLPAdditionalData {
+    minLpOut: Int as coins; // Minimum required amount of lp tokens to receive
+    receiverAddress: Address; // Address where lp tokens will be sent
+    bothPositive: Bool = true; // Trigger liquidity deposit only if both token amounts are non-zero
+    fwdGas: Int as coins = 0; // Gas used to forward a message in transfer notification after mint if customPayload is present
+    customPayload: Cell? = null;
+}
+```
+
+Below are shown gas constants. All of them are taken from the sdk [STON.fi sdk](https://github.com/ston-fi/sdk). They are slightly differ from reality, but it is ok for now.
```
Gusarich: same fees related comment as I left above

docs/src/content/docs/cookbook/dexes/stonfi.mdx
```diff
@@ -5,10 +5,362 @@ sidebar:
...
+        });
+    }
+}
+```
+
+### TON to Jetton
+
+TON to jetton swap introduces `ProxyTonTransfer` cause all interaction is done with pTON contract.
+
+Jetton `pTON` is a specialized asset within the STON.fi ecosystem on the TON blockchain. It functions as an intermediary within the STON.fi ecosystem to manage and facilitate efficient TON transfers.
```
Gusarich: I'd rework this paragraph a little bit to make sure readers understand that it's just a intermediate token that is used in stonfi smart contracts and that they don't actually have to think about working with pton directly. Also let's leave a link to related stonfi docs page

docs/src/content/docs/cookbook/dexes/stonfi.mdx
```diff
@@ -5,10 +5,362 @@ sidebar:
...
+
+```tact
+message(0xf8a7ea5) JettonTransfer {
+    queryId: Int as uint64;
+    amount: Int as coins;
+    destination: Address;
+    responseDestination: Address?;
+    customPayload: Cell? = null;
+    forwardTonAmount: Int as coins;
+    forwardPayload: Cell?; // hack for now, equivalent to Slice as remaining with first bit set to 1
```
Gusarich: I wouldn't mention `Slice as remaining` thing here. Instead, explain that there's no way to represent `Either` tl-b type directly so we have to stick with this workaround that will only work if we want to store our forwardPayload in a reference (our case)

Gusarich (changes requested):

docs/src/content/docs/cookbook/dexes/stonfi.mdx
```diff
@@ -5,10 +5,362 @@ sidebar:
...
+                forwardPayload: forwardPayload.toCell(),
+            }.toCell(),
+        });
+    }
+}
+```
+
+### Withdraw liquidly
+
+Liquidly is withdrawn from a pool by burning lp tokens, a user then receives both pool tokens based on the current exchange rate.
```
novusnota: Probably has to be modified in the same way similar sentences are expressed in DeDust cookbook

docs/src/content/docs/cookbook/dexes/stonfi.mdx
```diff
@@ -5,10 +5,362 @@ sidebar:
...
+    refundAddress: Address;
+    forwardPayload: Cell?;
+}
+
+const RouterProxyTonWallet: Address = address("kQBbJjnahBMGbMUJwhAXLn8BiigcGXMJhSC0l7DBhdYABhG7"); // Router's pTON wallet Address
+const RouterJettonWallet: Address = address("kQAtX3x2s-wMtYTz8CfmAyloHAB73vONzJM5S2idqXl-_5xK"); // Router's Jetton Wallet Address
+
+contract TonToJetton {
+    receive() {}
+    receive("ton-jetton") {
```
novusnota: Perhaps, it makes sense to rewrite this and similar contracts with String-accepting receivers into global functions, like it's done for DeDust cookbook where we don't assume any contract structure and, basically, provide an API for users to rely on.

Additionally, most repeated Structs/Messages should be explained on top of the page, and then moved to the bottom of each example where they're used, so as to not clutter the initial read

docs/src/content/docs/cookbook/dexes/stonfi.mdx
```diff
@@ -5,10 +5,362 @@ sidebar:
...
+
+```tact
+message(0x01f3835d) ProxyTonTransfer {
+    queryId: Int as uint64 = 0;
+    tonAmount: Int as coins;
+    refundAddress: Address;
+    forwardPayload: Cell?;
+}
+
+const RouterProxyTonWallet: Address = address("kQBbJjnahBMGbMUJwhAXLn8BiigcGXMJhSC0l7DBhdYABhG7"); // Router's pTON wallet Address
```
novusnota: Inline comments like this are tough to read, since the user has to horizontally scroll for those a lot. Especially on mobile.

The suggestion here is to make comments precede what they're describing, like done for DeDust cookbook page

docs/src/content/docs/cookbook/dexes/stonfi.mdx
```diff
@@ -5,10 +5,362 @@ sidebar:
...
+contract JettonToJetton {
+    receive() {}
+    receive(msg: SwapJettonJetton) {
+        // This example directly uses the `myJettonWalletAddress` provided in the received message.
+        // In real-world scenarios, it's more reliable to calculate this address on-chain or save it during initialization to ensure accuracy and trustworthiness.
+        let myJettonWalletAddress: Address = msg.myJettonWalletAddress;
+
+        let offerAmount: Int = 100000; // the quantity of jettons being swapped
+
+        let forwardPayload = StonfiSwap{
```
novusnota: `StonfiSwap` isn't present in this example block, which fails its compilation. See prior suggestions on explaining Structs/Messages on top of the page, then placing those repeated ones by the end of code blocks with examples — both to make them compile correctly and to simplify the initial top-down left-to-right read for the users

docs/cspell.json
```diff
@@ -85,6 +85,8 @@
     "Seamus",
     "Sedov",
     "Stateinit",
+	  "Ston",
+	  "Stonfi",
```
novusnota: Fix formatting

novusnota (changes requested):

docs/src/content/docs/cookbook/dexes/stonfi.mdx
```diff
@@ -5,10 +5,362 @@ sidebar:
...
+            }.toCell(),
+        });
+    }
+}
+```
+
+### Withdraw liquidly
+
+Liquidly is withdrawn from a pool by burning lp tokens, a user then receives both pool tokens based on the current exchange rate.
+You can refer to examples of jetton burning [here](/cookbook/jettons#burning-jetton).
```
Shvandre: Please, mention, that burning LP jettons requires more TONs for gas. (Same as in dedust)

docs/src/content/docs/cookbook/dexes/stonfi.mdx
```diff
@@ -5,10 +5,362 @@ sidebar:
...
+
+message SwapJettonTon {
+    myJettonWalletAddress: Address;
+}
+
+contract JettonToTon {
+    receive() {}
+    receive(msg: SwapJettonTon) {
+        // This example directly uses the `myJettonWalletAddress` provided in the received message.
+        // In real-world scenarios, it's more reliable to calculate this address on-chain or save it during initialization to ensure accuracy and trustworthiness.
```
Shvandre: Maybe it's worth clarifying that the point is that anyone can send a JettonNotify, so you either need to save the token-wallet during initialization, or, for example, use TEP-89, if possible https://github.com/ton-blockchain/TEPs/blob/master/text/0089-jetton-wallet-discovery.md

docs/src/content/docs/cookbook/dexes/stonfi.mdx
```diff
@@ -5,10 +5,362 @@ sidebar:
...
+    }
+}
+```
+
+### Jetton to TON
+
+Swapping jetton to TON is similar to jetton/jetton swap. The only one difference is that `RouterJettonWallet` address should be replaced with `RouterProxyTonWallet`.
+
+```tact
+const RouterAddress: Address = address("kQALh-JBBIKK7gr0o4AVf9JZnEsFndqO0qTCyT-D-yBsWk0v"); // CPI Router v2.1.0
```
Shvandre: Okay, Ston.fi v2 is deployed in testnet, so using such addresses is justified. Perhaps it is worth adding an explanation that these are testnet addresses, and a link to the documentation? https://docs.ton.org/v3/concepts/dive-into-ton/ton-blockchain/smart-contract-addresses

novusnota:
Working on suggested fixes while reading STON.fi docs. Will try to have all suggestions resolved by Monday noon UTC.

UPD: Getting better at estimates overall, but I'm still working on this.
UPD2: Whew, finally resolved everything.

novusnota (dismissed):

Gusarich (dismissed):

anton-trunov:
@novusnota Do you want to make any last minute changes here?

novusnota (approved):

PR #964, opened 2024-10-16
feat(docs): destructuring statement

novusnota:
## Issue

A follow-up docs PR for issue #298 and PR #856.

## Checklist


docs/src/content/docs/book/statements.mdx
```diff
@@ -96,6 +98,131 @@ value += 5;     // augmented assignment (one of the many, see below)
...
+    //        ------  -----
+    //        ↑       ↑
+    //        |       this variable will be of type Int,
+    //        |       same as the "first" field on Struct Two
+    //        this variable will be of type String,
+    //        same as the "second" field on Struct Two
+}
+```
+
+Destructuring assignment is exhaustive and requires specifying all the fields as variables. To deliberately ignore some of the fields, use an underscore `_{:tact}`, which would make fields value considered unused and discarded. Note, that such wildcard variable name `_{:tact}` cannot be accessed:
```
anton-trunov: It's not exhaustive, you can say `let Two { first } = s` which is equivalent to `let Two { second: _, first } = s` (as in the example below)

anton-trunov (changes requested):
Awesome stuff, I especially liked the note about the unsupported nested structs. Let's fix one small inaccuracy and we are good to go

docs/src/content/docs/book/statements.mdx
novusnota: Curiosly, that depends. The following example (taken from docs):

```tact
struct Two { first: Int; second: String }

fun discard(s: Two) {
    let Two { first } = s;
    // ^^^^^^^^^^^^^^^^^^^
}
```

Will report an error: "Cannot evaluate expression to a constant: destructuring assignment expected 2 fields, but got 1".

But if one would place the same destructuring statement in the `receive()` function, there would be no such error:

```tact
struct Two { first: Int; second: String }
message HasTwo { s: Two }

contract Example {
   receive() {}
   receive(msg: HasTwo) {
       // Almost the same line as in the previous example
       let Two { first } = msg.s;
       
       // No errors before or here
       dump(first);
   }
}
```

docs/src/content/docs/book/statements.mdx
```diff
@@ -96,6 +98,131 @@ value += 5;     // augmented assignment (one of the many, see below)
...
+
+The syntax is derived from the [`let` statement](#let), and instead of specifying the variable name directly it involves specifying the structure type on the left side of the [assignment operator `={:tact}`](/book/operators#assignment), which corresponds to the structure type of the value on the right side.
+
+```tact {6}
+// Definition of Example
+struct Example { number: Int }
+
+fun basic() {
+    // Basic syntax of destructuring assignment (to the left of "="):
+    let Example { number } = Example { number: 42 };
```
Gusarich: Maybe we should split these two types of syntax here so that readers won't mistakenly think that these should always be used together? Like make a variable struct and then destructure it.

Gusarich: we actually have the variable below, so we can just move its definition above.

docs/src/content/docs/book/statements.mdx
Gusarich: Oh! I forgot to remove this error from the interpreter:
https://github.com/tact-lang/tact/blob/9522f818503acd253b0328d897b774134eca2548/src/interpreter.ts#L1468-L1475

I'll open a PR with a fix.

Gusarich (changes requested):

docs/src/content/docs/book/statements.mdx
Gusarich: Oh! I forgot to remove this error from interpreter. I'll open a fix PR:
https://github.com/tact-lang/tact/blob/9522f818503acd253b0328d897b774134eca2548/src/interpreter.ts#L1468-L1475

docs/src/content/docs/book/statements.mdx
anton-trunov: Or we can even have even more complex expression instead of a variable, like a function call of a function returning a struct

anton-trunov (dismissed):

Gusarich (dismissed):
LGTM!

docs/src/content/docs/book/statements.mdx
```diff
@@ -96,6 +98,157 @@ value += 5;     // augmented assignment (one of the many, see below)
...
+```tact
+// "first" goes first, then goes "second"
+struct Two { first: Int; second: String }
+
+fun order(s: Two) {
+    let Two { second, first } = s;
+    //        ------  -----
+    //        ↑       ↑
+    //        |       this variable will be of type Int,
+    //        |       same as the "first" field on Struct Two
```
jeshecdom, line 200: Which preposition is the correct one? 
"field on Struct Two"   or
"field in Struct Two"
=)

docs/src/content/docs/book/statements.mdx
```diff
@@ -96,6 +98,157 @@ value += 5;     // augmented assignment (one of the many, see below)
...
+    //        this variable will be of type String,
+    //        same as the "second" field on Struct Two
```
jeshecdom: The same here.

docs/src/content/docs/book/statements.mdx
```diff
@@ -96,6 +98,157 @@ value += 5;     // augmented assignment (one of the many, see below)
...
+}
+```
+
+Destructuring assignment is exhaustive and requires specifying all the fields as variables. To deliberately ignore some of the fields, use an underscore `_{:tact}`, which would make fields value considered unused and discarded. Note, that such wildcard variable name `_{:tact}` cannot be accessed:
```
jeshecdom: The statement "would make fields value considered unused and discarded" is confusing. What about rewording it like this? "which discards the considered field's value".

jeshecdom (commented):
Love the explanation. It is very clear.

docs/src/content/docs/book/statements.mdx
novusnota: I think we can further simplify the sentence to get this: "To deliberately ignore some of the fields, use an underscore `_{:tact}` to discord the value of those fields."

jeshecdom (commented):
Once the associated PR #969 is approved, I will approve this PR.

anton-trunov (approved):

PR #969, opened 2024-10-16
fix: remove the destructuring fields quantity check in interpreter

### LINKED ISSUE #968: destructuring of less fields than are in struct throws an error from interpreter
Curiosly, that depends. The following example (taken from docs):

```tact
struct Two { first: Int; second: String }

fun discard(s: Two) {
    let Two { first } = s;
    // ^^^^^^^^^^^^^^^^^^^
}
```

Will report an error: "Cannot evaluate expression to a constant: destructuring assignment expected 2 fields, but got 1".

But if one would place the same destructuring statement in the `receive()` function, there would be no such error:

```tact
struct Two { first: Int; second: String }
message HasTwo { s: Two }

contract Example {
   receive() {}
   receive(msg: HasTwo) {
       // Almost the same line as in the previous example
       let Two { first } = msg.s;
       
       // No errors before or here
       dump(first);
   }
}
```

_Originally posted by @novusnota in https://github.com/tact-lang/tact/pull/964#discussion_r1803814496_

Gusarich:
## Issue

Closes #968.

## Checklist

- ~~[ ] I have added tests to demonstrate the contribution is correctly implemented: this usually includes both positive and negative tests, showing the happy path(s) and featuring intentionally broken cases~~


anton-trunov (changes requested):
Let's add the previously failing test case from the linked issue

src/interpreter.ts
```diff
@@ -1465,14 +1465,6 @@ export class Interpreter {
                 ast.expression.loc,
             );
         }
-        if (ast.identifiers.size !== Object.keys(val).length - 1) {
```
jeshecdom, line 1468: I think the solution is not to remove this code from the interpreter, but to add this number check in the typechecker (`resolveStatements.ts` in the `statement_destruct` case of the switch). Let me explain why.

The interpreter is only called when attempting to reduce an expression to a value. So, for example, in this code:

```
struct Two { first: Int; second: String }

fun discard(s: Two) {
    let Two { first } = s;  // A
}

contract TestContract {

    get fun test1(): Int {
        discard(Two {first: 5, second: "10"});   // B
        return 0;
    }
.............
```

Line `B` will trigger the interpreter, because `test1` is calling the `discard` function. In other words, the interpreter will execute the code of `discard`, detecting the error at line `A`. If line `B` is removed, the interpreter is never called, hence it will not detect the error at line `A` in that case. Therefore, the error at line `A` should be detected by the typechecker (i.e., the interpreter is just an optimization phase). 

At compile-time, there is nothing in the code triggering a `receive`, which means that the interpreter will never be called in the `receive` example (again, this should be caught by the typechecker):

```
struct Two { first: Int; second: String }
message HasTwo { s: Two }

contract TestContract {

    receive(msg: HasTwo) {
       // Almost the same line as in the previous example
       let Two { first } = msg.s;
       
       // No errors before or here
       dump(first);
   }
```

Just to give an example that the interpreter does catch an example similar to the previous one, in the following example it says the correct error at line `A` (just changed `message HasTwo` to a struct):
```
struct Two { first: Int; second: String }
struct HasTwo { s: Two }

fun discard2(m: HasTwo) {
    let Two { first } = m.s;   // A
}

contract TestContract {

    get fun test2(): Int {
        discard2(HasTwo {s: Two {first: 5, second: "10"}});
        return 0;
    }
.............
```

Comming back to the code in `resolveStatements.ts`, I see that the code at line 736 will only detect that the left side of a destructuring statement is contained in the right-hand side. But it will not detect the case when the right-hand side has more fields that the left-hand side, as in the examples above. The number check needs to be added before line 736.

jeshecdom (commented):
Regarding issue

src/interpreter.ts
novusnota: To add to your point — doesn't removing such check defeat the purpose of having `field: _` syntax in the first place? Like, if there's no requirement to unpack all fields, the only purpose of `_` is to temporarily stop binding the certain field to a certain variable name. But in this case, it would have been easier to just remove the field from the destructuring statement altogether, wouldn't it?

Because the type and number of elements in any Struct/Message are known at compile-time, I'm guessing it should be possible to add that check when resolving `receive()` functions and alike too :)

jeshecdom: Yes. I think when the check is added to `resolveStatements.ts`, it will handle all the cases, including those in `receive` declarations.

anton-trunov:
@Gusarich Let's finish this PR asap, please

src/interpreter.ts
Gusarich: so do you suggest adding this check back (to typechecker)? @anton-trunov what do you think?

anton-trunov: I would say we should make the user specify all the fields of a struct unless they explicitly say they do not want to do that. For instance, Rust has the `..` syntax for this case, so for the example above this will be `let Two { first, .. } = struct;`

Gusarich: so should we also go with the `..` syntax?

anton-trunov: Yeah, let's go for it

novusnota (dismissed):
LGTM, I've [updated the docs to include `..`](https://github.com/tact-lang/tact/pull/964)

src/grammar/grammar.ohm
```diff
@@ -158,7 +158,8 @@ Tact {
 
     StatementForEach = foreach "(" id "," id "in" Expression ")" "{" Statement* "}"
 
-    StatementDestruct = let typeId "{" ListOf<DestructItem, ","> ","? "}" "=" Expression (";" | &"}")
+    StatementDestruct = let typeId "{" ListOf<DestructItem, ","> ","? "}" "=" Expression (";" | &"}") --noRest
+                      | let typeId "{" ListOf<DestructItem, ","> "," ".." ","? "}" "=" Expression (";" | &"}") --withRest
```
anton-trunov: I'd say that allowing trailing comma after `..` makes little sense, since you cannot add more stuff later

```suggestion
    StatementDestruct = let typeId "{" ListOf<DestructItem, ","> (","? | "," "..") }" "=" Expression (";" | &"}")
```

src/grammar/ast.ts
```diff
@@ -326,6 +326,7 @@ export type AstStatementDestruct = {
     type: AstTypeId;
     /** field name -> [field id, local id] */
     identifiers: Map<string, [AstId, AstId]>;
+    rest: boolean;
```
```suggestion
    ignoreUnspecifiedFields: boolean;
```

src/grammar/test/stmt-destructuring.tact
```diff
@@ -12,12 +12,12 @@ message M {
...
     let S { a, b, c } = s;
-    let S { a: a1 } = s;
-    let S { b: b1 } = s;
-    let S { c: c1 } = s;
-    let S { a: a2, b: b2 } = s;
-    let S { a: a3, c: c3 } = s;
-    let S { b: b4, c: c4 } = s;
+    let S { a: a1, .. } = s;
+    let S { b: b1, .. } = s;
+    let S { c: c1, .., } = s;
```
anton-trunov: this should fail

anton-trunov (changes requested):

jeshecdom (commented):
Tests LGTM. 👍

src/grammar/test/stmt-destructuring.tact
Gusarich: because of `..,` on line 17?

anton-trunov: yep

src/grammar/grammar.ohm
novusnota: Better do `("," ".." | ","?)`, otherwise the ".." will always be omitted in the parse.

Or just `("," ".."?)?`

novusnota: image

https://ohmjs.org/editor/

anton-trunov: `("," ".."?)?` -- this is harder to understand

`("," ".." | ","?)` -- this means the list of field patterns ends either ends with an optional trailing comma or the ignore unspecified fields pattern

Gusarich:
@anton-trunov have a look please!

src/prettyPrinter.ts
```diff
@@ -697,7 +697,9 @@ export class PrettyPrinter {
                 acc.push(id);
                 return acc;
             }, []);
-        return `${this.indent()}let ${this.ppAstTypeId(statement.type)} {${ids.join(", ")}} = ${this.ppAstExpression(statement.expression)};`;
+        const restPattern = statement.ignoreUnspecifiedFields ? ", .." : "";
+        console.log(restPattern);
```
anton-trunov: debugging leftovers

anton-trunov (changes requested):

anton-trunov:
Looks good to me. Let's merge the current version and then make one more iteration for v1.6 with the generalized destructuring, so we can do nested destructs

src/prettyPrinter.ts
Gusarich: removed

anton-trunov (approved):

PR #980, opened 2024-10-23
feat(config): Export Zod schemas

### LINKED ISSUE #979: `parseConfig.ts`: Export Zod Schema Definitions
We should exports entries like these:

https://github.com/tact-lang/tact/blob/b7ba7e19565c5ded0b8a95aa1d43c92dbb81cecb/src/config/parseConfig.ts#L3

In order to make them available for third-party tools.

Since Tact always requires a configuration file for execution, tools might want to generate this file. Exporting these definitions will enable them to make their code more type-safe.

Instead this:
```
    const tactConfig: ConfigProject = {
      name: projectName,
        path: projectRoot,
        output: "/tmp/misti/output",
        options: {
            debug: false,
            external: true,
        },
    };
```

We would be able to write this:
```
projectSchema.parse({
        name: projectName,
        path: projectRoot,
        output: "/tmp/misti/output",
        options: {
            debug: false,
            external: true
        }
    });
```

jubnzv:
Closes #979

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,6 +8,11 @@ export {
     TactConstEvalError,
     TactErrorCollection,
 } from "./errors";
+export {
+    optionsSchema,
+    projectSchema,
+    configSchema,
+} from "./config/parseConfig";
 export { AstSorter } from "./grammar/sort";
 export { AstRenamer } from "./grammar/rename";
 export { AstHasher } from "./grammar/hash";
--- a/src/config/parseConfig.ts
+++ b/src/config/parseConfig.ts
@@ -1,6 +1,6 @@
 import { z } from "zod";
 
-const optionsSchema = z
+export const optionsSchema = z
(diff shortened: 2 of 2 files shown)


novusnota (approved):
Nice, I needed that change too :)

src/index.ts
```diff
@@ -8,6 +8,11 @@ export {
     TactConstEvalError,
     TactErrorCollection,
 } from "./errors";
+export {
+    optionsSchema,
+    projectSchema,
+    configSchema,
```
novusnota, line 14: While we're at it, we might also export `parseConfig()` and `verifyConfig()` here

anton-trunov (approved):

src/index.ts
jubnzv: Let's do it later, I believe we should create a separate index file

PR #984, opened 2024-10-24
feat(api): export `precompile()` and allow passing `AstModule[]` in

### LINKED ISSUE #983: `src/pipeline/precompile`: allow passing parsed `AstModule[]` to `precompile()`
The `precompile()` function calls the `openContext()`, which can optionally receive a `parsedModules?: AstModule[]` parameter:

https://github.com/tact-lang/tact/blob/a90b7af2c3117410187cc22b9990d0bc07707645/src/grammar/store.ts#L67-L72

But the `precompile()` function itself doesn't have that parameter (yet), which would be SUPER useful in language server contexts, and for other tools that work with incomplete or invalid code in the current file/module, but presumably complete code in imported modules.

novusnota:
## Issue

Closes #983.

## Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,5 @@
 export { enableFeatures, build } from "./pipeline/build";
+export { precompile } from "./pipeline/precompile";
 export {
     TactError,
     TactParseError,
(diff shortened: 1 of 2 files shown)


anton-trunov:
This won't make it to 1.6

anton-trunov:
Actually, let's get it into 1.6, since we have agreed to postpone the release

anton-trunov (approved):

PR #986, opened 2024-10-25
feat: new functions in stdlib from `stdlib.fc` and `math.fc`

### LINKED ISSUE #770: Port `stdlib.fc` and `mathlib.fc`'s asm functions to Tact's stdlib
This is going to be very easy after #768 is resolved.

It is partially addressed in these issues:
- #496
- #618

Gusarich:
## Issue

Closes #770
Towards #1029 (only `myCode()`)

## Checklist

## Current State

- [x] All stdlib.fc functions are present in Tact stdlib in some form, with some exceptions:
  - no tuple manipulation functions, because there is no tuple support in Tact. Do we really need them? @anton-trunov 
  - no `minmax` and no `slice_bits_refs` because we don't have tensor types in Tact and it would be useless: #1116
  - no `set_data` and `get_data` because Tact manages persistent storage by itself
  - no `bless` and other continuation functions because Tact manages control flow itself and for running some arbitrary code it's better to use RUNVM which doesn't require continuations parameters
  - no `raw_reserve_extra` because it requires an `uint32 -> varuint32` dictionary and we only support `varuint16` for values currently: #1117
- [x] All mathlib.fc functions are present in Tact stdlib in some form, with some exceptions:
  - no fixed-point arithmetic related functions: #1118
  - no `nan` and `is_nan` because we don't support NaN in Tact. Do we really need them? @anton-trunov 
  - no `sub_rev` because it is useless in our case
  - no `...MOD` instructions because they return two values as a tensor and it won't be very developer-friendly in Tact to make them return tuples in my opinion: #1116
- [x] All new functions are covered with tests


stdlib/stdlib.fc
```diff
@@ -645,3 +645,22 @@ builder store_builder(builder to, builder from) asm "STBR";
...
+int get_storage_fee(int workchain, int seconds, int bits, int cells) asm(cells bits seconds workchain) "GETSTORAGEFEE";
+int get_forward_fee(int workchain, int bits, int cells) asm(cells bits workchain) "GETFORWARDFEE";
+int get_precompiled_gas_consumption() asm "GETPRECOMPILEDGAS";
+
+int get_simple_compute_fee(int workchain, int gas_used) asm(gas_used workchain) "GETGASFEESIMPLE";
+int get_simple_forward_fee(int workchain, int bits, int cells) asm(cells bits workchain) "GETFORWARDFEESIMPLE";
+int get_original_fwd_fee(int workchain, int fwd_fee) asm(fwd_fee workchain) "GETORIGINALFWDFEE";
+int my_storage_due() asm "DUEPAYMENT";
+
+tuple get_fee_cofigs() asm "UNPACKEDCONFIGTUPLE";
```
novusnota:
```suggestion
tuple get_fee_configs() asm "UNPACKEDCONFIGTUPLE";
```

stdlib/stdlib.fc
```diff
@@ -440,7 +440,7 @@ builder store_ref(builder b, cell c) asm(c b) "STREF";
 ;;; Stores `slice` [s] into `builder` [b]
 builder store_slice(builder b, slice s) asm "STSLICER";
 
-;;; Stores (serializes) an integer [x] in the range `0..2^120 − 1` into `builder` [b].
+;;; Stores (serializes) an integer [x] in the range `0..2^128 − 1` into `builder` [b].
```
novusnota: These things keep bubbling up :)

stdlib/stdlib.fc
```diff
@@ -645,3 +645,22 @@ builder store_builder(builder to, builder from) asm "STBR";
 
 ;;; Retrieves code of smart-contract from c7
 cell my_code() asm "MYCODE";
+
```
novusnota: Maybe, it would be handy to visually separate the regular `stdlib.fc` (from TON monorepo) and the additions you make specifically for Tact's use. Like, `;;; --- Tact-adjusted ---`, or something

stdlib/std/math.tact
```diff
@@ -59,4 +59,6 @@ native log(num: Int, base: Int): Int;
 @name(__tact_pow)
 native pow(base: Int, exp: Int): Int;
 
-asm fun pow2(exp: Int): Int { POW2 }
\ No newline at end of file
+asm fun pow2(exp: Int): Int { POW2 }
+
+asm fun sgn(x: Int): Int { SGN }
```
novusnota: Maybe, name it `sign()` for it to look less cryptic and be simpler to type

```suggestion
asm fun sign(x: Int): Int { SGN }
```

stdlib/std/cells.tact
```diff
@@ -147,4 +157,19 @@ inline fun emptyCell(): Cell {
...
+    cells: Int;
+    bits: Int;
+    refs: Int;
+}
+
+asm fun computeDataSize(cell: Cell, maxCells: Int): DataSize { CDATASIZE TRIPLE }
+
+asm fun computeSliceDataSize(slice: Slice, maxCells: Int): DataSize { SDATASIZE TRIPLE }
+
+asm fun cellDepth(cell: Cell): Int { CDEPTH }
```
novusnota: This should probably be an extension function

```suggestion
asm extends fun depth(self: Cell): Int { CDEPTH }
```

stdlib/std/cells.tact
```diff
@@ -105,6 +107,14 @@ asm extends mutates fun skipBits(self: Slice, l: Int) { SDSKIPFIRST }
 
 asm extends fun endParse(self: Slice) { ENDS }
 
+asm extends fun skipLastBits(self: Slice, len: Int): Slice { SDSKIPLAST }
+
+asm extends fun firstBits(self: Slice, len: Int): Slice { SDCUTFIRST }
+
+asm extends fun sliceLast(self: Slice, len: Int): Slice { SDCUTLAST }
+
+asm fun sliceDepth(slice: Slice): Int { SDEPTH }
```
novusnota: This should probably be an extension function

```suggestion
asm extends fun depth(self: Slice): Int { SDEPTH }
```

stdlib/std/cells.tact
```diff
@@ -48,6 +48,8 @@ asm extends fun refs(self: Builder): Int { BREFS }
 
 asm extends fun bits(self: Builder): Int { BBITS }
 
+asm fun builderDepth(builder: Builder): Int { BDEPTH }
```
novusnota: This should probably be an extension function

```suggestion
asm extends fun depth(self: Builder): Int { BDEPTH }
```

stdlib/std/cells.tact
```diff
@@ -147,4 +157,19 @@ inline fun emptyCell(): Cell {
 
 inline fun emptySlice(): Slice {
     return emptyCell().asSlice();
-}
\ No newline at end of file
+}
+
+asm fun stringHash(data: Slice): Int { SHA256U }
```
novusnota: This one doesn't make much sense, since in Tact `String` and `Slice` are considered different types, while the `Slice` already has the `.hash()` method which uses the same TVM instruction. And for both that method and this function, the user would have to do explicit conversion `theirString.asSlice()`, so this function `stringHash()` is probably redundant and should be removed :)

novusnota (commented):
Sorry for reviewing the draft, I just wanted to keep those notes around :)

stdlib/stdlib.fc
Gusarich: oh, again, haha 😆
it's from the copy-pasting the latest stdlib from Core team projects

stdlib/stdlib.fc
Gusarich: well, there's nothing adjusted specifically for Tact. I took the "newer" stdlib from one of the recent core projects and removed the stuff that they put there for their FunC code (helper functions for building cells and some constants already existing in Tact etc)

stdlib/std/cells.tact
Gusarich: yes, theres `.hash` method for `String` type as well. Removed the function.

stdlib/stdlib.fc
Gusarich: fixed

Gusarich:
@novusnota @anton-trunov Implemented all functions. I'd like to hear some intermediate feedback on them before I proceed with writing tests.

I made sure to include all missing functions from both stdlib.fc and mathlib.fc with some exceptions (see PR description).

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `replace` and `replaceGet` methods for the `Map` type: PR [#941](https://github.com/tact-lang/tact/pull/941)
 - Utility for logging errors in code that was supposed to be unreachable: PR [#991](https://github.com/tact-lang/tact/pull/991)
 - Docs: `preloadRef` method for the `Slice` type: PR [#1044](https://github.com/tact-lang/tact/pull/1044)
+- New functions in stdlib from `stdlib.fc` and `math.fc`: PR [#986](https://github.com/tact-lang/tact/pull/986)
```
anton-trunov: Let's list all the new functions here -- this is super-useful for editor plugins

stdlib/std/cells.tact
```diff
@@ -191,3 +201,16 @@ inline fun emptyCell(): Cell {
...
     return emptyCell().asSlice();
 }
+
+struct DataSize {
+    cells: Int;
+    bits: Int;
+    refs: Int;
+}
+
+asm fun computeDataSize(cell: Cell, maxCells: Int): DataSize { CDATASIZE TRIPLE }
```
anton-trunov: Would `TRIPLE` work here? Structures are returned as tensors, right?

anton-trunov (commented):
afair, `mathlib.fc` has lots of functions for fixed-arithmetic as well, I think we should add those too (perhaps as a separate library, like `fixed_point_math.tact`).

Also, I left a couple comments in the source code

stdlib/std/cells.tact
Gusarich: `TRIPLE` makes three elements into a tuple of length 3. I think it should work

anton-trunov: yes, but our structs are tensors, not tuples, right?

CHANGELOG.md
Gusarich: done

stdlib/std/cells.tact
Gusarich: structs are implemented as tuples in Tact (that's why we had problems with long ones in the first place, remember?)

anton-trunov: ah, I must have mistaken it with the unbundled approach in the storage

stdlib/std/math.tact
```diff
@@ -59,4 +59,18 @@ native log(num: Int, base: Int): Int;
...
 native pow(base: Int, exp: Int): Int;
 
-asm fun pow2(exp: Int): Int { POW2 }
\ No newline at end of file
+asm fun pow2(exp: Int): Int { POW2 }
+
+asm fun sign(x: Int): Int { SGN }
+
+// floor(x*y/2^z)
+asm fun mulrshift(x: Int, y: Int, z: Int): Int { MULRSHIFT }
```
novusnota:
```suggestion
asm fun mulShiftRight(x: Int, y: Int, z: Int): Int { MULRSHIFT }
```

Not sure if that'll make it less recognizable to users that know about `MULRSHIFT` instruction, but it'll definitely make the function more [accessible to others](https://docs.tact-lang.org/book/operators/#binary-bitwise-shift-right) :)

Similar functions can be renamed in similar fashion

NaNOteh: let two: Int = 2;
two >> 1; // 1
4 >> 1;   // 2
5 >> 1;   // 2, due to flooring of integer values

pow(2, 254) >> 254; // 1

src/generator/writers/writeStdlib.ts
```diff
@@ -1133,6 +1133,32 @@ export function writeStdlib(ctx: WriterContext): void {
         });
     });
 
+    ctx.fun(`__tact_log2_floor_unsafe`, () => {
+        ctx.signature(`int __tact_log2_floor_unsafe(int num)`);
+        ctx.context("stdlib");
+        ctx.asm("", "UBITSIZE");
+    });
+
+    ctx.fun(`__tact_sqrt`, () => {
```
anton-trunov: how about we implement this in Tact?

Gusarich: and what about other stdlib functions?

anton-trunov: if a function can be implemented efficiently in Tact, we should implement it in Tact or open an issue and fix it

src/generator/writers/writeStdlib.ts
```diff
@@ -1133,6 +1133,32 @@ export function writeStdlib(ctx: WriterContext): void {
...
+        ctx.flag("inline");
+        ctx.context("stdlib");
+        ctx.body(() => {
+            ctx.write(`
+                if (num == 0) { return 0; }
+                int s = ${ctx.used("__tact_log2_floor_unsafe")}(num);
+                int x = (s == 1 ? (num - 1) / 2 + 1 : 1 << ((s + 1) / 2));
+
+                do {
+                    int q = (muldivc(num, 1, x) - x) / 2;
```
anton-trunov: looks like `muldivc(num, 1, x)` can be replaced with `divc`

stdlib/std/math.tact
Gusarich: done

src/generator/writers/writeStdlib.ts
Gusarich: done

src/generator/writers/writeStdlib.ts
Gusarich: done

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `replace` and `replaceGet` methods for the `Map` type: PR [#941](https://github.com/tact-lang/tact/pull/941)
 - Utility for logging errors in code that was supposed to be unreachable: PR [#991](https://github.com/tact-lang/tact/pull/991)
 - Docs: `preloadRef` method for the `Slice` type: PR [#1044](https://github.com/tact-lang/tact/pull/1044)
+- New functions in stdlib from `stdlib.fc` and `math.fc`: `Builder.depth`, `Slice.skipLastBits`, `Slice.firstBits`, `Slice.lastBits`, `Slice.depth`, `Cell.computeDataSize`, `Slice.computeDataSize`, `Cell.depth`, `curLt`, `blockLt`, `setGasLimit`, `getSeed`, `setSeed`, `myCode`, `sign`, `mulrshift`, `mulrshift_round`, `mulrshift_ceil`, `sqrt`, `addressNone`: PR [#986](https://github.com/tact-lang/tact/pull/986)
```
novusnota:
```suggestion
- New functions in stdlib from `stdlib.fc` and `math.fc`: `Builder.depth`, `Slice.skipLastBits`, `Slice.firstBits`, `Slice.lastBits`, `Slice.depth`, `Cell.computeDataSize`, `Slice.computeDataSize`, `Cell.depth`, `curLt`, `blockLt`, `setGasLimit`, `getSeed`, `setSeed`, `myCode`, `sign`, `divc`, `muldivc`, `mulShiftRight`, `mulShiftRightRound`, `mulShiftRightCeil`, `sqrt`, `addressNone`: PR [#986](https://github.com/tact-lang/tact/pull/986)
```

Neat!

CHANGELOG.md
NaNOteh: > done

Gusarich:
@anton-trunov @novusnota covered all new functions with tests!

Found and fixed a few little mistakes in types (for `addressNone` and `computeDataSize` functions). Also added comments with formulas for new math functions

anton-trunov:
@Gusarich Let's
- resolve the merge conflicts
- move the notes from "current state" to their own separate issues

anton-trunov (changes requested):
We should document the new functions in https://docs.tact-lang.org/ref/core-math and other relevant doc pages

novusnota:
> We should document the new functions in https://docs.tact-lang.org/ref/core-math and other relevant doc pages

Will do

Gusarich:
@anton-trunov rebased

src/test/e2e-emulated/contracts/stdlib.tact
```diff
@@ -58,4 +58,70 @@ contract StdlibTest {
...
+
+    get fun lastBits(sc: Slice, n: Int): Slice {
+        return sc.lastBits(n);
+    }
+
+    get fun sliceDepth(sc: Slice): Int {
+        return sc.depth();
+    }
+
+    get fun addressNone(): Address? {
```
Shvandre: Why does it return "Maybe Address" and not "Address" ?

src/test/e2e-emulated/contracts/stdlib.tact
```diff
@@ -58,4 +58,70 @@ contract StdlibTest {
...
+    get fun blockLt(): Int {
+        return blockLt();
+    }
+
+    get fun setGasLimit(gl: Int): Int {
+        setGasLimit(gl);
+        let x = 0;
+        repeat (100) {
+            x += 1;
+        }
```
Shvandre, line 126: Maybe I just do not understand something, but why do we need this loop here?

Shvandre: Ah, okay, just for testing purposes.

stdlib/std/cells.tact
```diff
@@ -210,3 +220,18 @@ inline fun emptyCell(): Cell {
 inline fun emptySlice(): Slice {
     return emptyCell().asSlice();
 }
+
+asm fun addressNone(): Address? { b{00} PUSHSLICE }
```
Shvandre: Same comment as above "Why Maybe Address instead of Address"

stdlib/std/math.tact
```diff
@@ -49,14 +49,52 @@ asm fun abs(x: Int): Int { ABS }
...
+
+    let s: Int = log2(num);
+    let x: Int = (s == 1 ? (num - 1) / 2 + 1 : 1 << ((s + 1) / 2));
+
+    let q: Int = 0;
+
+    do {
+        q = (divc(num, x) - x) / 2;
+        x += q;
+    } until (q == 0);
```
Shvandre: What is the exact method used here? Googled and didn't find anything similar to it

src/test/e2e-emulated/contracts/stdlib.tact
Gusarich: Because the way Tact currently work with addresses is that "null" as an address is actually addressNone

stdlib/std/math.tact
Gusarich: https://github.com/tact-lang/tact/issues/770#issuecomment-2481863626

Shvandre: Perfect

src/test/e2e-emulated/contracts/stdlib.tact
Shvandre: Well, I checked the sources and you are 100% right, but maybe we should change it?
Or change the tlb schemes generation, because 
```tact
struct Test {
    my_addr: Address?;
}

```

results
```
## Test
TLB: `_ my_addr:Maybe address = Test`
Signature: `Test{my_addr:Maybe address}`
```
In .md file (and maybe in some other places)

Shvandre (dismissed):

novusnota:
> We should document the new functions in https://docs.tact-lang.org/ref/core-math and other relevant doc pages

Halfway there, will be finishing this soon.

UPD: 2/3rds are done, got to complete the cell related functions
UPD: done!

src/test/e2e-emulated/contracts/stdlib.tact
Gusarich: it's an old problem in Tact, which was blocking multiple issues in the past. I remember some old comments in address-related issues about that.

My opinion is that we should rework addresses, because currently the way they work is not intuitive and doesn't fit into how other types work.
Basically, `addr_none` should be a valid part of `Address` type and should be checked by devs against `addressNone()` (and not `null`). And `Address?` should mean `Maybe Address`, which also could contain `addr_none`.
@anton-trunov what do you think?

anton-trunov: > addr_none should be a valid part of Address type and should be checked by devs against addressNone() (and not null).

yeah, this is how it should work, indeed

> And Address? should mean Maybe Address

also agreed

> which also could contain addr_none.

this one I'm not sure about, I'd say addr_none in the optional version should be `000`, what do you think?

Gusarich: > this one I'm not sure about, I'd say addr_none in the optional version should be `000`, what do you think?

well yes, that's what I meant. I was talking about the possibility of the case when value of `Address?` is equal to `addressNone()`.

anton-trunov: @Gusarich Maybe you could take care of this change first (allowing `addr_none` as an inhabitant of the `Address` type) and then we merge this PR with the corresponding changes

anton-trunov: We could also probably add a config option to turn off address verification along the way

Gusarich:
> move the notes from "current state" to their own separate issues

@anton-trunov have a look at the updated notes. there are few of them that I don't think we need to cover in Tact

anton-trunov:
Looks good! Thanks for adding those

novusnota:
@anton-trunov @Gusarich 

Added documentation for all of the new functions from this PR, except for the `addressNone()`, since its implementation is still in the works. Didn't add doc comments yet — once we agree on the descriptions in the documentation, then I'll quickly add the doc comments mirroring them.

docs/src/content/docs/ref/core-math.mdx
```diff
@@ -59,13 +61,169 @@ abs(-42);       // 42
...
+
+Usage examples:
+
+```tact
+muldivc(4, 1, 2);  // 2
+muldivc(3, 1, 2);  // 2
+muldivc(-4, 1, 2); // -2
+muldivc(-3, 1, 2); // -1
+muldivc(-3, 0, 2); // 0
+muldivc(-3, 0, ); // ERROR! Exit code 4: Integer overflow
```
Shvandre: Why there is no 3rd parameter?

novusnota: A typo, thanks for spotting it!

docs/src/content/docs/ref/core-math.mdx
```diff
@@ -59,13 +61,169 @@ abs(-42);       // 42
...
 
+## sign
+
+<Badge text="Available since Tact 1.6" variant="tip" size="medium"/><p/>
+
+```tact
+fun sign(x: Int): Int;
+```
+
+Computes and returns the sign of the [`Int{:tact}`][int] value `x`. Produces $1$ is the `x` is negative, $-1$ if the `x` is positive, and $0$ if the `x` is $0$.
```
Shvandre: I think it produces 1 if x is positive, -1 if x is negative ....

novusnota: Yup, but examples were correct :)

src/test/e2e-emulated/contracts/math.tact
anton-trunov, line 1: do we want to also add `muldiv`?

anton-trunov:
Let's bring this up to date, resolve the review comments and merge

verytactical:
Maybe we should split it into smaller PRs? Given the amount of discussion it created, it would probably be faster to merged if we didn't block merging the code with discussions on an unrelated part of stdlib.

An uneducated guess would be 3-4 PRs:

image

novusnota:
> Maybe we should split it into smaller PRs? Given the amount of discussion it created, it would probably be faster to merged if we didn't block merging the code with discussions on an unrelated part of stdlib.

Well, the only unresolved thing here besides merge conflics is the `addressNone()` function — other functions were reviewed and even have public docs written for them right in this PR already.

Therefore, I suggest @Gusarich you (or I) fix merge conflics, add the doc comments to stdlib based on the documentation I've added prior, and then the `addressNone()` will wait for completion as per #1029

@verytactical @anton-trunov

Gusarich:
@novusnota @anton-trunov @Shvandre have a look please!

docs/src/content/docs/ref/core-cells.mdx
```diff
@@ -84,6 +84,64 @@ let c: Cell = emptyCell();
 let fizz: Slice = c.beginParse();
 ```
 
+## Cell.depth
+
+<Badge text="Available since Tact 1.6" variant="tip" size="medium"/><p/>
+
+```tact
+extends fun depth(self: Cell): Int;
```
Shvandre: Maybe the argument should be optional? `extends fun depth(self: Cell?)`. I haven't tested, but TVM.pdf (and TVM spec also) states

> D765 — CDEPTH (c – x), returns the depth of Cell c. If c has no
references, then x = 0; otherwise x is one plus the maximum of depths
of cells referred to from c. If c is a Null instead of a Cell, returns zero.

docs/src/content/docs/ref/core-cells.mdx
```diff
@@ -84,6 +84,64 @@ let c: Cell = emptyCell();
...
+
+Extension function for the [`Cell{:tact}`][cell].
+
+Computes and returns the [`Int{:tact}`][int] [depth][std-repr] of the [`Cell{:tact}`][cell]. Produces $0$ if the [`Cell{:tact}`][cell] has no references, otherwise $1$ plus the maximum of the depths of the referenced cells.
+
+Usage example:
+
+```tact
+let c: Cell = beginCell().storeInt(42, 7).endCell();
+let depth: Int = c.depth;
```
Shvandre: Are round braces missing here? Or it is more like a `field`, not `method`?

novusnota: Yes, they are missing, good catch!

docs/src/content/docs/ref/core-cells.mdx
novusnota: Good point, but calling `null.depth()` in order to simply obtain 0 doesn't make much sense. And if the user has something like `let x: Cell? = ...` at any point, I doubt calling `x.depth()` will be an efficient way to determine if `x` is null or not, since the `!!` would still be needed for later uses that require `Cell` and not `Cell?` type.

But I might be totally wrong :)

Shvandre: I'd say it makes sense if your Cell? is fro, incoming message, and your logic is somehow based on that. I haven't seen those examples, but I would add that as there is no overhead, and internal TVM instruction provides support for it.

docs/src/content/docs/ref/core-cells.mdx
```diff
@@ -84,6 +84,64 @@ let c: Cell = emptyCell();
...
+```
+
+## Cell.computeDataSize
+
+<Badge text="Available since Tact 1.6" variant="tip" size="medium"/><p/>
+
+```tact
+extends fun computeDataSize(self: Cell, maxCells: Int): DataSize;
```
Shvandre: Same comment here. What are the benefits of narrowing the application of a function? `CDATAZISE` works fine with `null`, and returns $x=y=z=0$

docs/src/content/docs/ref/core-cells.mdx
```diff
@@ -367,6 +425,25 @@ let b: Builder = beginCell();
...
+
+Extension function for the [`Builder{:tact}`][builder].
+
+Computes and returns the [`Int{:tact}`][int] [depth][std-repr] of the [`Builder{:tact}`][builder]. Produces $0$ if the [`Builder{:tact}`][builder] has no references stored so far, otherwise $1$ plus the maximum of the depths of the referenced cells.
+
+Usage example:
+
+```tact
+let b: Builder = beginCell().storeInt(42, 7);
+let depth: Int = b.depth;
```
Shvandre: same comment about b.depth

docs/src/content/docs/ref/core-common.mdx
```diff
@@ -125,6 +127,10 @@ Creates a new [`Address{:tact}`][p] based on the [`chain` id](https://ton-blockc
 
 This function tries to resolve constant values in [compile-time](/ref/core-comptime) whenever possible.
 
+Attempts to specify an invalid `chain` ID (not $-1$ or $0$) throw an exception with [exit code 136](/book/exit-codes#136): `Invalid address`.
```
Shvandre: Is this behavior remained the same after address-verification removal?

docs/src/content/docs/ref/core-common.mdx
```diff
@@ -125,6 +127,10 @@ Creates a new [`Address{:tact}`][p] based on the [`chain` id](https://ton-blockc
...
+
+Attempts to specify an `chain` ID pointing to the masterchain ($-1$) without [masterchain support](/book/masterchain) enabled throw an exception with [exit code 137](/book/exit-codes#137): `Masterchain support is not enabled for this contract`.
```
Shvandre: Same question, but regarding masterchain flag removal

novusnota:
@Gusarich is `addressNone()` done? That is, may I document it in its current state, returning `Address?`?

If not, we can remove it from the changelog, docs and sources for now. And then make a follow-up PR that'll re-introduce `addressNone()` with changes to the `Address` type itself, as suggested earlier :)

docs/src/content/docs/ref/core-common.mdx
novusnota: Nope, although the const-eval throws an error if it can detect an invalid `chain` ID, i.e. not -1 or 0

novusnota: Edited the texts

docs/src/content/docs/ref/core-common.mdx
novusnota: Such runtime errors don't occur anymore, although the const-eval still throws a compilation error if it can detect an invalid `chain` ID, i.e. not -1 or 0

novusnota:
I've applied most of the suggestions from the code review and also added doc comments to everything, but `addressNone()`.

docs/src/content/docs/ref/core-cells.mdx
Gusarich: changed

docs/src/content/docs/ref/core-cells.mdx
Gusarich: @Shvandre changed

novusnota:
There's an issue for `addressNone()` already: https://github.com/tact-lang/tact/issues/1029. Thus, I've marked this PR as "Towards" it, but not "Closing" it, and removed the `addressNone()` from all places.

Now we can finally merge this PR :tada: 

@Gusarich @anton-trunov @Shvandre

src/test/e2e-emulated/contracts/math.tact
novusnota: Probably not, since it seems to be the same as just doing `(x * y) / z` with regular operators

novusnota (dismissed):

anton-trunov:
merge conflicts

anton-trunov:
@novusnota merge conflicts still

novusnota:
New ones it seems — I've resolved previous ones last time :)
One moment

novusnota:
macOS CIs exceed some time limits even after restarting those in isolation. Other than that, this PR seems to be done, I've even added the couple of "gas-expensive" badges :)

> We should document the new functions in https://docs.tact-lang.org/ref/core-math and other relevant doc pages

:heavy_check_mark:

docs/src/content/docs/ref/core-math.mdx
```diff
@@ -59,13 +61,169 @@ abs(-42);       // 42
...
+
+## sqrt
+
+<Badge text="Available since Tact 1.6" variant="tip" size="medium"/><p/>
+
+```tact
+fun sqrt(num: Int): Int;
+```
+
+Computes the [square root](https://en.wikipedia.org/wiki/Square_root) of the [`Int{:tact}`][int] value `num`. Returns the result rounded to the nearest integer. If there are two equally close integers, rounding is done toward the even one.
```
anton-trunov, line 93: maybe https://en.wikipedia.org/wiki/Integer_square_root is a better link here?

docs/src/content/docs/ref/core-math.mdx
```diff
@@ -59,13 +61,169 @@ abs(-42);       // 42
...
+
+## mulShiftRight
+
+<Badge text="Available since Tact 1.6" variant="tip" size="medium"/><p/>
+
+```tact
+fun mulShiftRight(x: Int, y: Int, z: Int): Int;
+```
+
+Computes and returns the [rounded down][round-down] result of `(x * y) / z{:tact}`. It is a more gas-efficient equivalent of doing the [bitwise shift right](/book/operators#binary-bitwise-shift-right) on the result of multiplication of [`Int{:tact}`][int] `x` times [`Int{:tact}`][int] `y`, where [`Int{:tact}`][int] `z` is the right operand of the shift.
```
anton-trunov:
```suggestion
Computes and returns the [rounded down][round-down] result of `(x * y) / 2^z{:tact}`. It is a more gas-efficient equivalent of doing the [bitwise shift right](/book/operators#binary-bitwise-shift-right) on the result of multiplication of [`Int{:tact}`][int] `x` times [`Int{:tact}`][int] `y`, where [`Int{:tact}`][int] `z` is the right operand of the shift.
```

docs/src/content/docs/ref/core-math.mdx
```diff
@@ -59,13 +61,169 @@ abs(-42);       // 42
...
+
+If the value in calculation goes beyond the range from $-2^{256}$ to $2^{256} - 1$ inclusive, an exception with [exit code 4](/book/exit-codes#4) is thrown: `Integer overflow`.
+
+Attempts to specify a negative value of `z` throw an exception with [exit code 5](/book/exit-codes#5): `Integer out of expected range`.
```
anton-trunov: Let's check if there are other restrictions on `z` like the ones for the shift operations

anton-trunov (changes requested):
Almost done, just one final push

docs/src/content/docs/ref/core-math.mdx
novusnota: Hmm, our current algo rounds upwards to the nearest even integer, but the integer square root is supposed to floor and we don't do that per se.

For example, our `sqrt(3)` gives 2, but according to integer square root the result should've been 1. See `python3 -c "import math; print(math.isqrt(3))"`

docs/src/content/docs/ref/core-math.mdx
novusnota: Yup, not just negative, but also values of 257 and onward aren't allowed.

The allowed range is from 0 to 256 inclusive

docs/src/content/docs/ref/core-math.mdx
novusnota: Applied the fix, and in the code comments too

docs/src/content/docs/ref/core-math.mdx
novusnota: Added that info to docs and code comments

novusnota:
Let's finally merge this :)

src/stdlib/stdlib/std/internal/contract.tact
```diff
@@ -240,3 +240,106 @@ asm fun parseStdAddress(slice: Slice): StdAddress { REWRITESTDADDR }
...
+///     nativeRandomize(lt); // equivalent to calling nativeRandomizeLt()
+/// }
+/// ```
+///
+/// See:
+/// * https://docs.tact-lang.org/ref/core-advanced#curlt
+/// * https://docs.tact-lang.org/ref/core-advanced#nativerandomize
+/// * https://docs.tact-lang.org/ref/core-advanced#nativerandomizelt
+///
+asm fun curLt(): Int { LTIME }
```
i582, line 260: Let's call it `currentLogicalTime`

src/stdlib/stdlib/std/internal/contract.tact
```diff
@@ -240,3 +240,106 @@ asm fun parseStdAddress(slice: Slice): StdAddress { REWRITESTDADDR }
...
+///
+/// ```tact
+/// fun example() {
+///     let time: Int = blockLt();
+/// }
+/// ```
+///
+/// See: https://docs.tact-lang.org/ref/core-advanced#blocklt
+///
+asm fun blockLt(): Int { BLOCKLT }
```
i582, line 274: `blockLogicalTime`

src/stdlib/stdlib/std/internal/math.tact
```diff
@@ -49,14 +49,174 @@ asm fun abs(x: Int): Int { ABS }
 
 asm fun now(): Int { NOW }
 
+// floor(log2(num))
```
i582: Please add more information about what you are describing here

src/stdlib/stdlib/std/internal/math.tact
```diff
@@ -6,8 +6,8 @@
 // the second with the big-endian representation of x.
 asm fun nativeRandomize(x: Int) { ADDRAND }
 
-/// Equivalent to `nativeRandomize(cur_lt())`,
-/// where `cur_lt()` is the logical time of the current transaction
+// Equivalent to `nativeRandomize(cur_lt())`,
```
i582: We (and you too, actually) use `///` for documentation comments :D
Also we don't have `cur_lt()` function in stdlib

src/test/e2e-emulated/math.spec.ts
```diff
@@ -507,6 +507,64 @@ describe("math", () => {
...
+        expect(await contract.getMulShiftRightCeil(2n, 4n, 1n)).toBe(4n);
+        expect(await contract.getMulShiftRightCeil(2n, 4n, 0n)).toBe(8n);
+        expect(await contract.getMulShiftRightCeil(1n, 6n, 0n)).toBe(6n);
+        expect(await contract.getMulShiftRightCeil(1n, 7n, 3n)).toBe(1n);
+        expect(await contract.getMulShiftRightCeil(1n, 4n, 3n)).toBe(1n);
+        expect(await contract.getMulShiftRightCeil(1n, 3n, 3n)).toBe(1n);
+
+        // sqrt: round(sqrt(num))
+        for (let num = 0n; num <= 100n; num++) {
+            // console.log(`sqrt(${num}) = ${await contract.getSqrt(num)}`);
```
i582: Please, remove commented code

src/stdlib/stdlib/std/internal/cells.tact
```diff
@@ -382,3 +474,95 @@ inline fun emptyCell(): Cell {
...
+}
+
+/// Extension function for the `Cell`. Available since Tact 1.6.0.
+///
+/// Computes and returns the number of distinct cells, bits and refs in the `Cell` by using a depth-first search (DFS) algorithm, recursively traversing each referenced cell. This function is computationally expensive and can consume a lot of gas. If `self` is `null`, returns `DataSize` with all fields set to 0.
+///
+/// The results are packed into a `DataSize` Struct consisting of:
+/// * `cells: Int`, the total number of nested cells, including the starting one
+/// * `bits: Int`, the total number of bits in all nested cells, including the starting one
+/// * `refs: Int`, the total number of refs in all nested cells, including the starting one
```
i582, line 502: These comments should be in the structure itself (for hover documentation)

src/stdlib/stdlib/std/internal/math.tact
novusnota: Yeah, those comments were not touched at all by intent of this PR and I've only explicitly made them `//` since those comments didn't follow the doc comment guidelines :)

But I'll update those too, sure

src/stdlib/stdlib/std/internal/cells.tact
novusnota: Good point!

UPD: Wait, but wouldn't the structure itself simply have its own declaration present in the hover? That said, I'll add the `///` comments for each of the fields, yeah.

docs/src/content/docs/ref/core-math.mdx
novusnota: https://github.com/tact-lang/tact/issues/1710

novusnota:
Uno momento

src/stdlib/stdlib/std/internal/contract.tact
novusnota: Hmm, let's not, since we have the `nativeRandomizeLt()` already. And in general people are more or less aware of the meaning of "lt" or at least accustomed to seeing it. 

Besides, the function `now()` is not called `currentTime()`, but just `now()` :)

src/stdlib/stdlib/std/internal/contract.tact
novusnota: Same reasoning as here: https://github.com/tact-lang/tact/pull/986#discussion_r1942642896

novusnota:
@i582 Resolved your suggestions (hopefully), please take a look

i582 (approved):
LGTM

anton-trunov (approved):

novusnota:
Hooray :tada:

PR #991, opened 2024-10-28
refactor: add utility to err on unreachable branches

### LINKED ISSUE #915: Improve exhaustiveness checks
We could improve compile-time checks provided by the TypeScript compiler introducing [the `unreachable` function](https://github.com/nowarp/misti/blob/466e2663d76eecd7db38efe90f0d45065077f867/src/internals/util.ts#L18):
```typescript
export function unreachable(value: never): never {
  throw InternalException.make(`Reached impossible case`, { node: value });
}
```

It accepts the parameter `never`, so it could be used to perform a static exhaustiveness checks for `switch` clauses:
```
switch (node.kind) {
  case "id":
    // ...
  // other cases
  default:
    unreachable(node);
}
```

If some of the node kinds are not covered by the `case` clauses, we'll get a compilation error:
picture

That's quite useful especially to avoid regressions when changing enum types implementations.

verytactical:
## Issue

Closes #915.

## Checklist

### DIFF
--- a/src/generator/writers/writeStdlib.ts
+++ b/src/generator/writers/writeStdlib.ts
@@ -1,7 +1,7 @@
 import { contractErrors } from "../../abi/errors";
 import { maxTupleSize } from "../../bindings/typescript/writeStruct";
 import { enabledMasterchain } from "../../config/features";
-import { throwInternalCompilerError } from "../../errors";
+import { match } from "../../utils/tricks";
 import { WriterContext } from "../Writer";
 
 export function writeStdlib(ctx: WriterContext): void {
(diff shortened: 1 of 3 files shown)


anton-trunov:
It would be nice to add some usage examples for the new utilities

src/interpreter.ts
```diff
@@ -765,36 +766,28 @@ export class Interpreter {
...
-                return this.interpretStructInstance(ast);
-            case "field_access":
-                return this.interpretFieldAccess(ast);
-            case "static_call":
-                return this.interpretStaticCall(ast);
-        }
-    }
+    public interpretExpression = (ast: AstExpression): Value =>
+        match(ast)
+            .on({ kind: "id" })((ast) => this.interpretName(ast))
```
novusnota: Cool, but doesn't this create a ton of small `{ kind: whatever }` objects each time `interpretExpression()` is called?

anton-trunov:
Perhaps let's try applying it to the map key/value pair types once #993 is merged

src/interpreter.ts
verytactical: Indeed it does, but there's not much we can do about it.

src/interpreter.ts
```diff
@@ -765,36 +766,28 @@ export class Interpreter {
...
-                return this.interpretConditional(ast);
-            case "struct_instance":
-                return this.interpretStructInstance(ast);
-            case "field_access":
-                return this.interpretFieldAccess(ast);
-            case "static_call":
-                return this.interpretStaticCall(ast);
-        }
-    }
+    public interpretExpression = (ast: AstExpression): Value =>
```
anton-trunov: I think this change is not strictly necessary, since a missing case in this `switch` should be caught by ESLint.

src/generator/writers/writeStdlib.ts
```diff
@@ -1172,39 +1172,28 @@ function genTactDictGet(
...
-        switch (value) {
-            case "slice":
-            case "cell":
-            case "coins":
-                return "";
-            case "uint":
-            case "int":
-                return ", int vl";
-        }
+        return match(key)
```
anton-trunov: I would say this change is not strictly necessary, ESLint would catch it.

anton-trunov (changes requested):
Let's revert changes on `switch` which are not strictly necessary and keep them only for pairs of key/value types where exhaustivity is desired

anton-trunov:
CI is not all green

anton-trunov (approved):

PR #994, opened 2024-10-28
feat(docs): automatic links to Web IDE from code blocks in Cookbook

### LINKED ISSUE #913: Docs: introduce automatic or semi-automatic links to Web IDE from code snippets
Since just recently, the [Nujan IDE](https://github.com/nujan-io/nujan-ide) have been deployed to https://ide.ton.org (Web IDE). And today it also got [contract file sharing with Base64-encoded URLs](https://github.com/nujan-io/nujan-ide/commit/20c21527d4aa2baf8346e8a8147837abc35f8aa7).

With that feature, it's possible to form links like `https://ide.ton.org/?code=tact-code-encoded-as-base64&lang=tact`, which would open the decoded piece of Tact code in the IDE. The limit is 32779 characters per file, which is enough to share contracts in the order of ~850 lines including comments (even if those comments contain a lot of characters, like mine in [experimental lab setup](https://docs.tact-lang.org/book/debug/#lab) do).

Thus, the `remark`, `rehype` or `starlight` plugin can be made in order to get the code block contents and then conditionally produce the link to Web IDE to run it.

Things to consider:
1. Limit of characters
2. Whether this is a full snippet (i.e. contains any module-level definitions) or not

The created link combined with the code block may look something like this:

image

novusnota:
## Issue

Closes #913.

## Checklist

## Preview

See the small "Open in Web IDE" buttons on bottom right side of each (!) code block:

image_2024-10-28_23-02-25


anton-trunov:
Is is possible to move the links somewhere to the top of the code snippets?

novusnota:
> Is is possible to move the links somewhere to the top of the code snippets?

Yes, but I believe it's a bit worse since it kinda breaks the flow of the text. I can change the z-index or make it float, but that would overlap with text sometimes.

image

I've also tried extending the code blocks themselves and introducing another button adjacent to the copy button, but because the button's HTML/CSS were so rigid that attempt failed. They (expressive code plugin) have [great extensibility and API](https://expressive-code.com/guides/developing-plugins/) for various cool transformations on code blocks, but have no robust way to add another button :)

novusnota:
Right now I'm removing such buttons from code blocks with signatures in the Reference section mainly

anton-trunov:
I would also say that opening links in Web IDE mostly makes sense for complete contracts.

novusnota:
That was my original thought too, but then I decided to allow all code samples but the ones in signatures — that way users can quickly tinker with any examples. But we may only allow such links for Cookbook examples, that's also fine

anton-trunov:
> But we may only allow such links for Cookbook examples, that's also fine

yeah, let's start with Cookbook first and if there is popular demand, we will add more links

anton-trunov (approved):

PR #1009, opened 2024-10-29
feat: implement `Set.isSubsetOf()` for Node <22, which defaults to builtin implementation whenever possible

### LINKED ISSUE #1008: Add a polyfill or a replacement for `.isSubsetOf()` to support Node.js <22
See: https://github.com/tact-lang/web-ide/blob/716ccb163b7eda0046f5fede9563291be3db4f6f/src/lib/polyfills.ts

novusnota:
## Issue

Closes #1008.

## Checklist


verytactical (changes requested):
Please don't patch any `prototype`s. Other libraries might expect builtin ES methods to work exactly to the letter of specification, and this implementation doesn't really [look like it](https://github.com/es-shims/Set.prototype.isSubsetOf/blob/main/implementation.js).

In case something goes wrong, we'll be up to hours of debug, because `prototype` patching changes behavior of other code remotely and without any trace.

It would be much easier both for implementer and for future users to just define `isSubsetOf` function, and optionally default it to builtin `.isSubsetOf` method for increased performance on newer JS engines.

verytactical:
I should also mention that the concept of "polyfill" relies on the process around it. Polyfills are supplied by well-tested libraries, their authors are well-versed in spec legalese, and, most importantly, polyfills are applied by build infrastructure (for example, `babel`). Due to wider adoption of those polyfills, we reduce detection time for inevitable bugs in their implementations, and get centralized updates and security audits.

In this context neither the word "polyfill" nor its implementation appears as something end user might encounter in production codebase. At least not under regular circumstances.

novusnota:
@verytactical I generally agree against prototype patching, but thought this particular case might be fine. Plus, I simply took the working polyfill from [Web IDE repo](https://github.com/tact-lang/web-ide/blob/f55c9d62d36a50e67b4e8568ec9b3398bb09e632/src/lib/polyfills.ts). Which, I suppose, was in turn taken from [core-js](https://github.com/zloirock/core-js/blob/master/packages/core-js/internals/set-is-subset-of.js). Or not :)

Anyways, I can re-write one or another implementation as a global function with optional fallback to the builtin method if/when it's present

verytactical:
Please, define it as a function.

In case of WebIDE it's apparently either used in one of the libraries, or not used at all. Probably it's there, because it compiles code with `tact`, and with this PR it won't be needed anymore. Unfortunately I don't know circumstances of when it was added.

cc @rahulyadav-57

novusnota:
@verytactical all good now — the function is global, but your introduction of `fs.glob` in `scripts/copy-files.ts` has [tied us to Node.js 22](https://nodejs.org/api/fs.html#fspromisesglobpattern-options) :)

**UPD:** Turns out we do have the `glob` package used elsewhere, so I just replaced `fs.glob()` with `glob.sync()`

.github/workflows/tact.yml
```diff
@@ -22,6 +22,27 @@ jobs:
...
+        run: |
+          # Temporarily ignore engines
+          yarn config set ignore-engines true
+          # Install dependencies, gen and build the compiler
+          yarn install
+          yarn clean
+          yarn gen
+          yarn build
+          # Test some specific things for backwards compatibility
+          yarn jest -t 'isSubsetOf'
```
verytactical, line 43: nit: I'd rather create a "backCompat" folder or something. It's odd to have such a specific thing in CI code.

scripts/copy-files.ts
```diff
@@ -1,8 +1,10 @@
 import * as fs from "node:fs/promises";
 import * as path from "node:path";
+import * as glob from "glob";
 
 const cp = async (fromGlob: string, toPath: string) => {
-    for await (const file of fs.glob(fromGlob)) {
+    const files = glob.sync(fromGlob);
```
verytactical, line 6: nit: Now that I look at it... this is `async` function that does `glob.sync`, and then we have `for await`. Not that it will change anything, given there's only a few files, but I'd rather make it async :)

src/utils/isSubsetOf.ts
```diff
@@ -0,0 +1,39 @@
+/** Taken from TypeScript collection lib to perfectly match the .isSubsetOf signature */
+export interface ReadonlySetLike<T> {
```
verytactical, line 2: nit: But do we actually use it on setlikes? If it's always another set, I'd rather leave it at `other: Set<T>`.

src/utils/isSubsetOf.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * @returns a boolean indicating whether all the elements in Set `one` are also in the `other`.
+ */
+export function isSubsetOf<T>(
+    one: Set<T>,
+    other: ReadonlySetLike<unknown>,
+): boolean {
+    // If the builtin method exists, just call it
+    /* eslint-disable @typescript-eslint/no-explicit-any */
+    if ((Set.prototype as any).isSubsetOf) {
```
verytactical:
```suggestion
    if ('isSubsetOf' in Set.prototype) {
```

verytactical: We'll add eslint rules that disable both `as` and `any` (except for one only file) to ensure types are at least somewhat sound. It's hard to disable them right now, but it'd be nice to not add any more (pun probably intended).

src/utils/isSubsetOf.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+            return includes(elements, value);
+        },
+        keys(): Iterator<T> {
+            return createIterator(elements);
+        },
+    };
+}
+
+function includes<T>(target: T[], wanted: T) {
+    for (const element of target) if (wanted === element) return true;
```
verytactical: nit: Does our prettier config consider `{}` omission and same-line body legal? Oh wow, I should take a look there :>

verytactical:
```suggestion
    return target.some((element) => element === wanted);
```

verytactical (dismissed):
I think I should stop bikeshedding this :)

src/utils/isSubsetOf.spec.ts
novusnota: Well, I just took those helper functions as is, modulo adding type hints. But alright :)

src/utils/isSubsetOf.ts
novusnota: Regardless of what we use that on, the method signature in Node uses setlike, thus this function's signature replicating that method must also use setlikes. Or so I believe, at least :)

.github/workflows/tact.yml
novusnota: Hmm, good point, although my idea is to keep such list of things small up until Node.js 22 [goes into maintenance LTS state](https://nodejs.org/en/about/previous-releases) by the end of October 2025. This semi-arbitrary requirement is just for users to catch up on Node.js 22+ and not stumble on `.isSubsetOf is not a function` errors. Therefore, a whole new folder seems like fostering usage of features from v22+, which is opposite to the goal.

scripts/copy-files.ts
novusnota: There's only a callback API there as far as I can tell, so I just changed `for await` to `for`

.github/workflows/tact.yml
verytactical: Now that you mention it, I realized we have `"lib": ["ESNext"]` in `tsconfig.json`, and probably we could prevent any further issues by downgrading it to the lowest possible version of standard library. Can't use a method from too new of a library if it's not there, can we? :)

I tried it before, and unfortunately the code has quite a few places where not-yet-standard (ESNext-only) features are used, so it might take a bit more effort.

verytactical (approved):
.

Thug337 (commented):
Hi

Issue #879, opened 2024-09-27
Constants in default struct fields are not accessible


anton-trunov:
The following snippet should compile
```tact
const C: Int = 0b0001;

message Foo {
    asset0type: Int as uint4 = C;
}
```

but it fails with the following error message:

```
Syntax and type checking failed
Error: test.tact:4:32: Cannot evaluate expression to a constant: cannot evaluate declared constant "C" as it does not have a body
Line 4, col 32:
  3 | message Foo {
> 4 |     asset0type: Int as uint4 = C;
                                     ^
  5 | }
```

Kudos to @a-bahdanau for uncovering this

i582:
Looks like we have circular dependencies here:

https://github.com/tact-lang/tact/blob/b8c07e1c48373d69bbb77956a57ba4217f106124/src/types/resolveDescriptors.ts#L2177-L2180

On the one hand, we need to process fields first and then constants, and on the other hand (this issue) we need to process constants first 🤔

i582:
This code is too weird for me:
```tact
const MY_CONST: Int = S{}.f

struct S { f: Int = 42 }
```

The correct code should probably be:

```tact
const MY_CONST: Int = 42

struct S { f: Int = MY_CONST }
```

The first example uses weird reverse logic and it seems it should be banned in favor of the second example, which is much more logical.

i582:
Moreover, `const MY_CONST: Int = S{}.f` is invalid syntax right now, so I don't really understand this comment

verytactical:
[Why though?](https://www.typescriptlang.org/play/?#code/HYQwtgpgzgDiDGEAEBBJBvAUEnSIA8YB7AJwBcl4AbEKKJAZQyRgFcAjKgS3iQDMAXEmCsw7CCSQBeJABYATEgC+mbLgLFylIsCgUAsgE0A+gGEA8gDkGAFSEixE6cIgB3RgAoAlADo+AbkwVTFBIWARkACEMNRwNUgp4HT0kIzMrW3tRcUkZBVVcPEIEyho6RmY2Th5+LMdc1JMLaxtlIMwgA)

```typescript
namespace A {
    export class S { public f: number = 42 }

    export const MY_CONST: number = new S().f;
}

namespace B {
    export const MY_CONST: number = 42

    export class S { public f: number = MY_CONST }
}
```

i582:
I meant Tact:

```tact
import "@stdlib/deploy";

const MY_CONST: Int = S{}.f

[...]

PR #1010, opened 2024-10-30
feat(stdlib): Add `SendDefaultMode`

### LINKED ISSUE #988: `SendParameters`: Constant for `mode: 0`
We can define a constant for the `0` mode value to improve readability and describe the default behavior more clearly. This constant will replace any direct usage of `0` in the `mode` field throughout the codebase, as commonly seen in real-world contracts.
This change will make choosing the default mode more explicit and could be leveraged by `SuspiciousMessageMode`, which detects such cases: https://github.com/nowarp/misti/pull/199.

Context: https://docs.tact-lang.org/book/message-mode/

jubnzv:
Closes #988


Gusarich (changes requested):
I'd also suggest to change the `mode: Int = 0;` to use the new constant in `stdlib/std/send.tact`

Gusarich (dismissed):

jubnzv:
The badge looks this way:

image

docs/src/content/docs/book/message-mode.mdx
```diff
@@ -11,7 +11,7 @@ It's possible to use raw [`Int{:tact}`][int] values and manually provide them fo
 
 Mode value | Constant name                 | Description
 ---------: | :---------------------------- | -----------
-$0$        | -                             | Ordinary message (default).
+$0$        | `SendDefaultMode{:tact}`      | Ordinary message (default). <Badge text="Available since Tact 1.6" variant="tip" size="medium"/><p/>
```
novusnota:
```suggestion
$0$        | <Badge text="Available since Tact 1.6" variant="tip"/> 
 `SendDefaultMode{:tact}`      | Ordinary message (default).
```

Cool, but I'd do this, since the mode was there all this time, but the constant has just been introduced

novusnota: And the `import { Badge } from '@astrojs/starlight/components';` has to be added after the frontmatter (`---` stuff)

novusnota:
Add the `import { Badge } from '@astrojs/starlight/components';` after the frontmatter (`---` stuff). And also update the stdlib snapshot locally :)

jubnzv:
It will look better this way:

image

novusnota:
Hmm, but this one seems cleaner:

image

novusnota:
Or even

image

And the line in the table looks like this in .mdx:

```mdx
$0$        | <Badge text="Since Tact 1.6" variant="tip"/> `SendDefaultMode{:tact}` | Ordinary message (default).

```

jubnzv:
You can suggest and commit this change. In my opinion, it clutters the table.

docs/src/content/docs/book/message-mode.mdx
```diff
@@ -11,7 +13,7 @@ It's possible to use raw [`Int{:tact}`][int] values and manually provide them fo
 
 Mode value | Constant name                 | Description
 ---------: | :---------------------------- | -----------
-$0$        | -                             | Ordinary message (default).
+$0$        | `SendDefaultMode{:tact}`      | <Badge text="Since Tact 1.6" variant="tip"/> Ordinary message (default).
```
novusnota:
```suggestion
$0$        | <Badge text="Since Tact 1.6" variant="tip"/> `SendDefaultMode{:tact}` | Ordinary message (default).
```

There are like 3 entries in this table, and won't be any more in the foreseeable future. A bit of clutter on source side while clearly displaying things for the end-users is the way to go, I believe

jubnzv:
The CI issue cannot be reproduced locally:

image

Any ideas? Why do we need this check at all?

novusnota:
The CI issue is NOT related to package.json. Instead, you just forgot to run `yarn gen` to re-generate the stdlib cache

docs/src/content/docs/book/message-mode.mdx
```diff
@@ -11,7 +13,7 @@ It's possible to use raw [`Int{:tact}`][int] values and manually provide them fo
 
 Mode value | Constant name                 | Description
 ---------: | :---------------------------- | -----------
-$0$        | -                             | Ordinary message (default).
+$0$        | <Badge text="Since Tact 1.6" variant="tip"/> `SendDefaultMode{:tact}` | Ordinary message (default).
```
anton-trunov, line 16: maybe move the badge into Description?

novusnota: We've discussed that a bit with jubnzv, and my position is that the mode itself was present there all along while only the constant name for it is new. And that the table layout in the source matters less since this table isn't going to grow big and user-facing UX matters more

novusnota:
Also, before we've merged this PR, I'd like to add a small change to it — to exclude the links to web ide from site search as they can really cripple its speed. I'm working on it.

UPD: The fix is to add `data-pagefind-ignore` in the `links-to-web-ide.js` like so:

```js
// line 78
            '<a data-pagefind-ignore="all"',
```

anton-trunov (changes requested):
I agree with @Gusarich about `mode: 0` in the `SendParameters` structure. And also, we should use `mode: SendDefaultMode` in all the examples and tests instead of `mode: 0` (but let's not change the implicit modes)

jubnzv:
@anton-trunov using it in the structure definition leads to the compilation error.

anton-trunov:
@jubnzv this one #879?

jubnzv:
@anton-trunov yes

PR #1018, opened 2024-11-07
Pretty-printer refactored

### LINKED ISSUE #1037: Wrong priority for shift operators in pretty-printer

### Tact source code

```tact
contract Test {
    a: Int;

    init() {
        self.a = (1 << 2) + 3;
    }
}
```

```shell
contract Test {
    a: Int;

    init() {
        self.a = 1 << 2 + 3;
    }
}
```

### What happened?

`getPrecedence` doesn't have a case for `<<` or '>>', and their priority is inferred to be `11`, while by the grammar it should have been `7.5`.

After printing the code `self.a` is set to `32`, while previously it was `7`.

### What did you expect?

Pretty-printer should not change behavior.

### Steps to reproduce

_No response_

### How do you run Tact?

_No response_

### Anything else?

_No response_

### LINKED ISSUE #1038: `struct` and `message` are printed without indents in front of closing bracket

### Tact source code

_No response_

_No response_

### What happened?

_No response_

### What did you expect?

This is mostly a code consistency thing. While `struct` and `message` are never nested, and will always be displayed at indent 0, this might change in the future.

### Steps to reproduce

_No response_

### How do you run Tact?

_No response_

### Anything else?

_No response_

### LINKED ISSUE #1042: Postfix `!!` becomes prefix after printing

### Tact source code

```tact
x!!
```

_No response_

### What happened?

After printing becomes `!!x`.

### What did you expect?

`x!!`

### Steps to reproduce

_No response_

### How do you run Tact?

_No response_

### Anything else?

_No response_

### LINKED ISSUE #1043: Expression printer is broken

### Tact source code

_No response_

_No response_

### What happened?

_No response_

### What did you expect?

`1 * (2 + 3).foo()` becomes `1 * 2 + 3.foo()`

Printer doesn't consider which argument of ternary operator it's in

Printer doesn't consider nesting of prefix and postfix unary operators

### Steps to reproduce

_No response_

### How do you run Tact?

_No response_

### Anything else?

_No response_

verytactical:
## Issue

This was a demo for `makeVisitor`, but then it unfortunately worked a bit better than expected, and several bugs were found.

If new AST node types and operators are added, but printer is not updated, there will be an error message now.

Architecture was ruggedized to enforce correct handling of indents and precedences whenever code changes (`Precedence`, `ExprPrinter`, `Printer`). Further details are in code comments.

Closes #1037
Closes #1038
Closes #1042
Closes #1043

## Checklist

### DIFF
--- a/src/prettyPrinter.ts
+++ b/src/prettyPrinter.ts
@@ -248,7 +248,7 @@ export const ppAstExpressionNested = makeVisitor<A.AstExpression>()({
     method_call: ppAstMethodCall,
     field_access: ppAstFieldAccess,
 
-    op_unary: ppAstOpUnary, // x2
+    op_unary: ppAstOpUnary,
 
     op_binary: ppAstOpBinary,
 
@@ -339,7 +339,7 @@ const concat = ([head, ...tail]: readonly ContextModel[]): ContextModel => {
     if (isUndefined(nextHead)) {
         return head;
     }
-    // Otherwise merge results by setting
+    // Otherwise concatenate results, leaving indent only in front of the merged line
     return [...init, (level) => last(level) + nextHead(0), ...nextTail];
 };
 
--- a/src/prettyPrinter.ts
+++ b/src/prettyPrinter.ts
@@ -45,138 +45,205 @@ export const ppAstType = makeVisitor<A.AstType>()({
 // Expressions
 //
 
-export const precedenceMap: Readonly<Record<A.AstBinaryOperation, number>> = {
-    "||": 1,
-    "&&": 2,
-    "|": 3,
-    "^": 4,
-    "&": 5,
-    "==": 6,
-    "!=": 6,
-    "<": 7,
-    ">": 7,
-    "<=": 7,
-    ">=": 7,
-    "+": 8,
-    "-": 8,
-    "*": 9,
-    "/": 9,
-    "%": 9,
-    "<<": 11,
-    ">>": 11,
+export const unaryOperatorType: Record<A.AstUnaryOperation, 'post' | 'pre'> = {
+    "+": 'pre',
+    "-": 'pre',
+    "!": 'pre',
+    "~": 'pre',
+
+    "!!": 'post',
+};
+
+export const checkPostfix = (operator: A.AstUnaryOperation) => unaryOperatorType[operator] === 'post';
+
+
+/**
+ * Description of precedence of certain type of AST node
+ */
+export type Precedence = {
+    /**
+     * Add parentheses around `code` if in this `parent` position we need brackets
+     * @param check Position-checking function from parent
+     * @param code Code to put parentheses around
+     * @returns 
+     */
+    brace: (position: (childPrecedence: number) => boolean, code: string) => string;
+    /**
+     * Create a position-checking function for positions where grammar rule recurses
+     * For example in `foo = "+" foo | bar` this should be used for `foo`
+     * In case of `"+" "+" bar` input, this won't put parentheses around inner
+     * `"+" bar`
+     */
+    self: (childPrecedence: number) => boolean;
+    /**
+     * Create a position-checking function for positions where grammar rule uses
+     * a higher-priority grammar rule
+     * For example in `foo = foo "-" bar | bar` this should be used for `bar`
+     * This enforces parentheses in expressions such as `x - (y - z)` where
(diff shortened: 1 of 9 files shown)


src/prettyPrinter.ts
```diff
@@ -1,804 +1,889 @@
...
+     * For example in `foo = "+" foo | bar` this should be used for `foo`
+     * In case of `"+" "+" bar` input, this won't put parentheses around inner
+     * `"+" bar`
+     */
+    self: (childPrecedence: number) => boolean;
+    /**
+     * Create a position-checking function for positions where grammar rule uses
+     * a higher-priority grammar rule
+     * For example in `foo = foo "-" bar | bar` this should be used for `bar`
+     * This enforces parentheses in expressions such as `x - (y - z)` where
```
anton-trunov: I'm not sure I understand this comment on precedence here, since left/right associativity (which you seem to be hinting at) and precedence are pretty much orthogonal things

src/prettyPrinter.ts
```diff
@@ -1,804 +1,889 @@
...
-    }
+export const conditionalPrecedence = makePrecedence(5);
 
-    ppAstPrimitiveTypeDecl(primitive: AstPrimitiveTypeDecl): string {
-        return `${this.indent()}primitive ${this.ppAstId(primitive.name)};`;
-    }
+export const binaryPrecedence: Readonly<
+    Record<A.AstBinaryOperation, Precedence>
+> = {
+    "||": makePrecedence(10),
```
anton-trunov: do you think we could reuse the precedence levels from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence#precedence_and_associativity (or is it documented in the standard?), since we model this part of Tact's syntax after JS

anton-trunov: If you think we should stick with our own levels, then it should be documented in the Tact docs. @novusnota should be able to help with that

anton-trunov (commented):
Looks fabulous, just a few minor comments/questions

src/prettyPrinter.ts
verytactical: Well, precedence and associativity are cases of a more general thing: the grammar itself.

Rules can reference other rules or self-reference, and that's precisely what these methods are for.

anton-trunov: > Well, precedence and associativity are cases of a more general thing: the grammar itself

haha, I would argue that the grammar is only a formal expression of what we want intuitively :)
so precedence and associativity come first and then get expressed in the grammar :)

src/prettyPrinter.ts
verytactical: > since we model this part of Tact's syntax after JS

Sadly Tact's grammar is different from JS. Ternary operator and assignment are not on the same priority 2 in Tact, and priorities 15/17 are merged together.

src/prettyPrinter.ts
anton-trunov: but let's consider a concrete example: 

> both operators have same precedence, but right-hand size is allowed to be an operator of the same precedence only if there were `()`

what if we introduce the exponentiation operator (`**`) in a future Tact version? those are normally right-associative, so `2 ** 3 ** 4` would have unnecessary parens when pretty-printed, wouldn't it?

src/prettyPrinter.ts
anton-trunov: this is a bug and it should be fixed then

anton-trunov: and it wouldn't be even a violation of backwards compatibility: users should not rely on buggy behavior

anton-trunov: wait, assignment is not an expression in Tact

verytactical: Yes, because assignment is not an expression, it shouldn't change anything from user's perspective, but they do have different priority. Same with 15/17: we don't have a prefix `new`, so user can never distinguish whether they have different priorities anyway.

src/prettyPrinter.ts
verytactical: > what if we introduce the exponentiation operator

It will have a precedence of 105. In order to make it right-associative we'll have to swap `self` and `child` passed to `ppAstExpressionNested` of its arguments.

I'll try to improve wording of these comments, they're indeed confusing.

anton-trunov (approved):

PR #1047, opened 2024-11-15
refactor: add optimization phase

### LINKED ISSUE #970: Call interpreter even if a contract is not present in the module

### Tact source code

```tact
fun blowup(v: Int) {
    1 / v;
}

fun test() {
    blowup(0);   // Should produce a div by zero error when interpreter calls the function
}
```

_No response_

### What happened?

The interpreter is not activated. But, if we add a dummy contract, like so:
 ```
fun blowup(v: Int) {
    1 / v;       // A
}

fun test() {
    blowup(0); 
}

contract DummyContract {
}
```

The interpreter now produces the div by zero at line `A`.

### What did you expect?

I expected a div by zero at compile-time.

### Steps to reproduce

_No response_

### How do you run Tact?

Tact CLI

### Anything else?

_No response_

### LINKED ISSUE #1449: Add optimization phase
Add an optimization phase in between the typechecking and FunC code generation phases. As a first approximation, the optimization phase could carry out expression simplification by running the interpreter over all expressions in the AST. Later, other optimization techniques could be added; for example, constant propagation.

Calls to the interpreter that were previously being carried out during FunC code generation should be removed. The FunC generation phase should NOT be calling constant evaluation (this is a job of the optimization phase).

### PR DESCRIPTION
**Pull request by jeshecdom**
## Issue

Closes #1449.
Closes #970. (As a byproduct, a test was added).

This PR adds an optimization phase in between the typechecking and FunC code generation phases. Currently, the optimization phase only carries out expression simplification by running the interpreter over all expressions in the AST (later, this will be replaced by the partial evaluator). I did this in preparation for other optimization techniques that will be added in the future; for example, constant propagation will be moved to the optimization phase later.

IMPORTANT: This PR removes calls to the interpreter that were previously being carried out during the FunC code generation phase. The FunC generation phase should NOT be calling constant evaluation (this is a job of the optimization phase).

The PR adds two CLI compiler options:

- skipTactOptimizationPhase: Default is false. If true, skips the tact optimization phase. Skipping the phase is useful, for example, to carry out tests in which you want to deactivate the interpreter so that you work with the raw program. Simply add the option in your project's json:
```
"projects": [
    {
      "name": "test",
      "path": "./test.tact",
      "output": "./output",
      "options": { 
          "skipTactOptimizationPhase": true 
      }
    }
  ]
```

- dumpCodeBeforeAndAfterTactOptimizationPhase: Default is false. If true, it will produce two files: one with the program before the optimization phase, and another one after the optimization phase, so that you can make a diff on both files and see what the optimization phase did to the program.

NOTE: This PR does not modify the AST one gets from calling `getRawAST` on the `CompilerContext` object `ctx`. The FunC generation part seems to make use of the functions `getAllTypes`, `getAllStaticFunctions` and the like. This PR DOES modify the ASTs produced by those functions. The FunC generator part seems to use only the "sources" from `getRawAST`.

## Checklist

### CODE REVIEW

File: schemas/configSchema.json [hunk 1]
```diff
@@ -56,6 +56,16 @@
...
               },
+              "skipTactOptimizationPhase": {
+                "type": "boolean",
+                "default": false,
+                "description": "False by default. If set to true, skips the Tact code optimization phase."
+              },
+              "dumpOptimizedTactCode": {
+                "type": "boolean",
+                "default": false,
+                "description": "False by default. If set to true, dumps the code produced by the Tact code optimization phase. In case the optimization phase is skipped, this option is ignored."
```
verytactical: "Smooth" (in math sense) behavior would be to dump the code anyway, yet unoptimized. "Dump the optimized code" is just another separate optional compilation step that happens to be in the list after "optimize" step.

src/interpreter.ts
```diff
@@ -960,7 +960,7 @@ export class Interpreter {
                 if (foundContractConst.value !== undefined) {
                     return foundContractConst.value;
                 } else {
-                    throwErrorConstEval(
+                    throwNonFatalErrorConstEval(
```
verytactical: Sorry I don't understand this change, and I suspect a future reader wouldn't either.

When do we throw non-fatal errors in general?

src/optimizer/optimization_phase.ts
```diff
@@ -0,0 +1,40 @@
...
+export function optimize_tact(ctx: CompilerContext): CompilerContext {
+    // Call the expression simplification phase
+    ctx = simplify_expressions(ctx);
+
+    // Here, we will call the constant propagation analyzer
+
+    return ctx;
+}
+
+export function dump_tact_code(ctx: CompilerContext, file: string) {
```
verytactical: It's odd that the function calls `prettyPrint` several times.

`prettyPrint` is supposed to print the whole AST node, whatever its type is.

Either we're missing another type of AST node, or this is actually something in the lines of `prettyPrint({ kind: 'module', ... })`.

src/optimizer/optimization_phase.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    for (const f of getAllStaticFunctions(ctx)) {
+        program += `${prettyPrint(f.ast)}\n\n`;
+    }
+
+    for (const t of getAllTypes(ctx)) {
+        program += `${prettyPrint(t.ast)}\n\n`;
+    }
+
+    writeFileSync(file, program, {
```
verytactical: `writeFile` does `w` by default, doesn't it?

Also `*Sync` functions are synchronous and stop the whole JS virtual machine until they're done with what they were doing. It's fine to use a few of these in tests or in build scripts, but I think they can never be used from any of `src/**/*.ts`.

Eventually we might want to read several files during compilation, and in case of Web IDE probably even download those files over wire. `*Sync` is impossible to emulate in web (barring some unfortunate deprecated APIs that make the whole browser tab unresponsive), and there'll be no easy way to un`Sync` the code in the future.

This should probably use `import fs from "node:fs/promises"`

src/optimizer/test/__snapshots__/expr_simplification.spec.ts.snap
```diff
@@ -0,0 +1,184 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`expression-simplification should fail expression simplification for interpreter-called-when-no-contract 1`] = `
```
verytactical: Very, very wise use of Jest snapshots.

I was thinking to investigate how to do this, but you've already done it. Thanks!

src/optimizer/test/expr_simplification.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+        it("should pass expression simplification for " + r.name, () => {
+            let ctx = openContext(
+                new CompilerContext(),
+                [{ code: r.code, path: "<unknown>", origin: "user" }],
+                [],
+            );
+            ctx = featureEnable(ctx, "external");
+            ctx = resolveDescriptors(ctx);
+            ctx = resolveStatements(ctx);
+            ctx = simplify_expressions(ctx);
```
verytactical: We're using camel case ;)

verytactical: Also file names use kebab-case by Node.js convention.

src/optimizer/util.ts
```diff
@@ -22,22 +23,44 @@ export function extractValue(ast: AstValue): Value {
             return ast.value;
         case "string":
             return ast.value;
+        case "struct_instance":
+            return ast.args.reduce(
```
verytactical: `reduce` is usually not the right tool when you're mapping over an object.

JS has `Object.fromEntries` for this, but I'd rather (perversely) prefer to just

```
const object: Record<string, Foo> = {};
for (const field of fields) {
    ...
    object[key] = value;
}
return object;
```

for improved readability.

src/optimizer/util.ts
```diff
@@ -22,22 +23,44 @@ export function extractValue(ast: AstValue): Value {
...
+                (resObj, fieldWithInit) => {
+                    resObj[idText(fieldWithInit.field)] = extractValue(
+                        fieldWithInit.initializer as AstValue,
```
verytactical: If `initializer` is always an AstValue, please, change its type. If not, I'm afraid you can't do this :)

src/optimizer/util.ts
```diff
@@ -22,22 +23,44 @@ export function extractValue(ast: AstValue): Value {
...
+                    );
+                    return resObj;
+                },
+                { $tactStruct: idText(ast.type) } as StructValue,
```
verytactical: This doesn't do what it was intended to do: it casts arbitrary object to `StructValue` disregarding any inconsistencies in its structure.

Most likely it should've been `.reduce<StructValue>(..., { ... })` that ascribes this object a `StructValue` type.

src/optimizer/util.ts
```diff
@@ -46,20 +69,75 @@ export function makeValueExpression(value: Value): AstValue {
...
     if (typeof value === "boolean") {
         const result = createAstNode({
             kind: "boolean",
             value: value,
-            loc: dummySrcInfo,
+            loc: newSrc,
+        });
+        return result as AstValue;
+    }
+    if (typeof value === "object" && "$tactStruct" in value) {
```
verytactical: Unfortunately you were unlucky to encounter some bad code here that can't be extended anymore in the same way.

`typeof ... === '...'` only happen when the code creates an _unsafe_ union of incompatible types, and `in`... nobody knows what it does in TS, because TS doesn't distinguish between open and closed object types.

This condition can't be used in production code :/

src/optimizer/util.ts
```diff
@@ -46,20 +69,75 @@ export function makeValueExpression(value: Value): AstValue {
...
+        const fields = Object.entries(value)
+            .filter(([name, _]) => name !== "$tactStruct")
+            .map(([name, val]) => {
```
verytactical: And this is an another sign we can't go forward with this `Value` type. There isn't even a mention of `$tactStruct`  in its definition.

src/optimizer/util.ts
```diff
@@ -46,20 +69,75 @@ export function makeValueExpression(value: Value): AstValue {
...
+                return `${name}: ${valueToString(val)}`;
+            })
+            .join(", ");
+        return `${value["$tactStruct"] as string} { ${fields} }`;
+    }
+    throwInternalCompilerError(
```
verytactical: If someone finds `valueToString` and tries to use it for another purpose, we might get a surprise in production.

If it's planned to be implemented in the future, please, create an issue and `// TODO: <issue link>`.

src/pipeline/build.ts
```diff
@@ -86,6 +87,37 @@ export async function build(args: {
...
+            if (config.options?.dumpOptimizedTactCode) {
+                // Dump the code after optimization
+                dump_tact_code(
+                    ctx,
+                    config.output + `/${config.name}_optimized_tact_dump.tact`,
+                );
+            }
+        } catch (e) {
+            logger.error("Tact code optimization failed.");
+            logger.error(e as Error);
```
verytactical: Unfortunately `e` is not an `Error`. It has to be checked with `e instanceof Error` prior to this, and `throwInternalCompilerError`ed otherwise.

Because this is very likely to happen in every `catch (e)`, it'd be best to just fix the type of `logger.error(error: unknown)` and do it there.

src/optimizer/expr_simplification.ts
```diff
@@ -0,0 +1,713 @@
...
+
+    for (const f of getAllStaticFunctions(ctx)) {
+        if (f.ast.kind === "function_def") {
+            const statementsResult = process_statements(
+                f.ast.statements,
+                ctx,
+                interpreter,
+            );
+            const newStatements = statementsResult.stmts;
+            ctx = statementsResult.ctx;
```
verytactical: This is very unsafe. If this line wasn't here, we wouldn't know it's missing.

src/optimizer/expr_simplification.ts
```diff
@@ -0,0 +1,713 @@
...
+     *   declared in the contract and not in some trait that the contract inherits from.
+     *
+     * The above means that we will need to process the properties in TypeDescription first,
+     * and then use those properties to build the AST (carefully ensuring that only fields, constants and methods
+     * that were in the original AST, remain in the new AST).
+     */
+    const newTypes: Map<string, TypeDescription> = new Map();
+
+    for (const t of getAllTypes(ctx)) {
+        let newInitializer: InitDescription | null = null;
```
verytactical: It's better to separate such long loop bodies into separate functions. In this case this would allow to avoid an unnecessary mutable variable.

src/optimizer/expr_simplification.ts
```diff
@@ -0,0 +1,713 @@
...
+                    ...c.ast,
+                    initializer: newInitializer,
+                });
+                const newConstantDescription = { ...c, ast: newConstantCode };
+                newConstants.push(newConstantDescription);
+                newConstantsMap.set(c.name, newConstantDescription);
+            } else {
+                // The rest of kinds do not have explicit Tact expressions, so just copy the current description
+                newConstants.push(c);
+                newConstantsMap.set(c.name, c);
```
verytactical: If a change would be required here, it'd have to be done twice.

`newConstants` and `newConstantsMap` are a part of a single data structure that should have its own interface.

verytactical: Probably `newConstantsMap` could be computed from `newConstants` in a separate step.

src/optimizer/expr_simplification.ts
```diff
@@ -0,0 +1,713 @@
...
+
+        // Process fields
+        const newFields: FieldDescription[] = [];
+
+        // This map will be used to quickly recover the new definitions when
+        // building the AST later
+        const newFieldsMap: Map<string, FieldDescription> = new Map();
+
+        for (const f of t.fields) {
+            if (f.ast.initializer !== null) {
```
verytactical: Bodies of these loops look very similar. Also if there is a new place in AST that needs to be optimized, there are no failsafes to ensure it will be handled in optimizer.

src/optimizer/expr_simplification.ts
```diff
@@ -0,0 +1,713 @@
...
+                });
+                newMethods.set(name, { ...m, ast: newMethodCode });
+            } else {
+                // The rest of kinds do not have explicit Tact expressions, so just copy the current function description
+                newMethods.set(name, m);
+            }
+        }
+
+        // Now, we need to create the new AST, depending on its kind.
+        let newAst: AstTypeDecl;
```
verytactical: Not in a mutable variable, please.

src/optimizer/expr_simplification.ts
```diff
@@ -0,0 +1,713 @@
...
+            case "trait": {
+                const newDeclarations: AstTraitDeclaration[] = [];
+
+                for (const decl of t.ast.declarations) {
+                    switch (decl.kind) {
+                        case "asm_function_def":
+                        case "function_decl":
+                        case "function_def": {
+                            const newCode = newMethods.get(idText(decl.name))!
+                                .ast as AstTraitDeclaration;
```
verytactical: Both `!` and `as` are unsafe. Either the type should be fixed, or there should be a condition with `throwInternalCompilerError` here.

src/optimizer/expr_simplification.ts
```diff
@@ -0,0 +1,713 @@
...
+                            newDeclarations.push(newCode);
+                            break;
+                        }
+                        case "constant_decl":
+                        case "constant_def": {
+                            const newCode = newConstantsMap.get(
+                                idText(decl.name),
+                            )!.ast;
```
verytactical: `!`

src/optimizer/expr_simplification.ts
```diff
@@ -0,0 +1,713 @@
...
+                            newDeclarations.push(newCode);
+                            break;
+                        }
+                        case "field_decl": {
+                            const newCode = newFieldsMap.get(
+                                idText(decl.name),
+                            )!.ast;
```
verytactical: `!`

src/optimizer/expr_simplification.ts
```diff
@@ -0,0 +1,713 @@
...
+                    ...t.ast,
+                    declarations: newDeclarations,
+                });
+
+                break;
+            }
+        }
+
+        newTypes.set(t.name, {
+            ...t,
```
verytactical: Spread operator typechecks incorrectly if used on a union type. All the fields that are not modified by the code should be listed explicitly.

The only valid use-case for spread operator is when the type of an object is `Record<string, ...>`. Otherwise the program might compile, but work incorrectly.

src/optimizer/expr_simplification.ts
```diff
@@ -0,0 +1,713 @@
...
+    for (const stmt of statements) {
+        const result = process_statement(stmt, ctx, interpreter);
+        newStatements.push(result.stmt);
+        ctx = result.ctx;
+    }
+
+    return { stmts: newStatements, ctx: ctx };
+}
+
+function process_statement(
```
verytactical: This is a visitor, it's not unique to optimizer, and likely should go into `ast.ts`.

src/optimizer/expr_simplification.ts
```diff
@@ -0,0 +1,713 @@
...
+): { expr: AstExpression; ctx: CompilerContext } {
+    const value = tryExpressionSimplification(expr, interpreter);
+    let newExpr = expr;
+    if (value !== undefined) {
+        try {
+            newExpr = makeValueExpression(value, expr.loc);
+            // Register the new expression in the context
+            ctx = registerAllSubExpTypes(ctx, newExpr, getExpType(ctx, expr));
+        } catch (_) {
+            // This means that transforming the value into an AST node is
```
verytactical: We could have any kind of exception here, including stack overflow exception.

If there is a specific kind of exception that should be handled here, there has to be an `instanceof` check, and all other exceptions have to be rethrown up the stack.

verytactical (changes requested):

src/interpreter.ts
novusnota: > When do we throw non-fatal errors in general?

When we want to stop the constant evaluation, but continue compiling the rest of the contract. Like, when not being able to compute something at compile-time with ability to fallback to run-time implementation for that something.

verytactical: `stopEvaluation()`?

novusnota: Yeah, but the current pipeline kinda relies on throwing and catching errors and not on errors as values `¯\_(ツ)_/¯`

anton-trunov:
I'm wondering what is going on with this PR and a bit surprised with the lack of progress.

jeshecdom:
It required the refactor of the Value type, but now that it is solved, I will finish it ASAP.

package.json
```diff
@@ -95,7 +95,7 @@
     "@typescript-eslint/parser": "^7.0.2",
     "ajv-cli": "^5.0.0",
     "cross-env": "^7.0.3",
-    "cspell": "^8.8.3",
+    "cspell": "^8.16.0",
```
anton-trunov: irrelevant change

src/config/configSchema.json
```diff
@@ -51,6 +51,16 @@
                 "default": false,
                 "description": "False by default. If set to true, enables generation of a getter the information on the interfaces provided by the contract.\n\nRead more about supported interfaces: https://docs.tact-lang.org/ref/evolution/OTP-001."
               },
+              "skipTactOptimizationPhase": {
```
anton-trunov:
```suggestion
              "skipPartialEval": {
```

src/optimizer/optimization-phase.ts
```diff
@@ -0,0 +1,1124 @@
...
+
+    processStaticConstants(optCtx.ctx, nodeReplacements);
+
+    processTypes(optCtx.ctx, nodeReplacements);
+}
+
+export function dumpTactCode(ast: AstModule, file: string) {
+    const program = prettyPrint(ast);
+
+    void writeFile(file, program);
```
anton-trunov: you cannot use Node's file system here as it will break when Tact runs in the browser, use VFS instead

src/config/configSchema.json
```diff
@@ -51,6 +51,16 @@
...
+                "type": "boolean",
+                "default": false,
+                "description": "False by default. If set to true, skips the Tact code optimization phase."
+              },
+              "dumpCodeBeforeAndAfterTactOptimizationPhase": {
```
anton-trunov: This needs to be split into two options, `dumpAst` and `dumpAfterPartialEval`

yarn.lock
anton-trunov, line 1: undo this change

src/types/resolveExpression.ts
```diff
@@ -54,7 +54,12 @@ export function getExpType(ctx: CompilerContext, exp: AstExpression) {
     return t.description;
 }
 
-function registerExpType(
+export function expHasType(ctx: CompilerContext, exp: AstExpression): boolean {
+    const t = store.get(ctx, exp.id);
+    return t ? true : false;
```
```suggestion
    return Boolean(t);
```

src/pipeline/build.ts
```diff
@@ -98,6 +104,55 @@ export async function build(args: {
...
+        ctx,
+        ast,
+        doOptimizationFlag,
+    );
+
+    // Dump the code before optimization phase
+    if (config.options?.dumpCodeBeforeAndAfterTactOptimizationPhase) {
+        dumpTactCode(
+            optimizationCtx.originalAst,
+            config.output + `/${config.name}-unoptimized-tact-dump.tact`,
```
anton-trunov: use `path.join` instead of `+`

src/pipeline/build.ts
```diff
@@ -98,6 +104,55 @@ export async function build(args: {
...
+            }
+            return { ok: false, error: [e as Error] };
+        }
+    }
+
+    // Dump the code after optimization phase
+    if (config.options?.dumpCodeBeforeAndAfterTactOptimizationPhase) {
+        dumpTactCode(
+            optimizationCtx.modifiedAst,
+            config.output + `/${config.name}-optimized-tact-dump.tact`,
```
anton-trunov: use `path.join` instead of `+`

src/optimizer/test/expr-simplification.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+import { defaultParser } from "../../grammar/grammar";
+import { resolveDescriptors } from "../../types/resolveDescriptors";
+import { getAllExpressionTypes } from "../../types/resolveExpression";
+import { resolveStatements } from "../../types/resolveStatements";
+import { loadCases } from "../../utils/loadCases";
+import { simplifyAllExpressions } from "../expr-simplification";
+import { prepareAstForOptimization } from "../optimization-phase";
+
+describe("expression-simplification", () => {
+    for (const r of loadCases(__dirname + "/success/")) {
```
anton-trunov: use `path.join` instead of `+`

src/optimizer/test/expr-simplification.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+            );
+            ctx = featureEnable(ctx, "external");
+            ctx = resolveDescriptors(ctx, ast);
+            ctx = resolveStatements(ctx, ast);
+            const optCtx = prepareAstForOptimization(ctx, ast, true);
+            simplifyAllExpressions(optCtx);
+            expect(getAllExpressionTypes(optCtx.ctx)).toMatchSnapshot();
+        });
+    }
+    for (const r of loadCases(__dirname + "/failed/")) {
```
anton-trunov: use `path.join` instead of `+`

src/optimizer/optimization-phase.ts
```diff
@@ -0,0 +1,1124 @@
...
+    }
+}
+
+function ensureReceiver(ast: AstMutableNode): AstReceiver {
+    if (ast.kind === "receiver") {
+        return ast;
+    } else {
+        throwInternalCompilerError(`kind ${ast.kind} is not a receiver`);
+    }
+}
```
anton-trunov: all `ensureX` functions need to go to `ast/...`

anton-trunov (changes requested):
The first batch of review comments

src/config/configSchema.json
jeshecdom: Renaming it to skipPartialEval would imply that the optimization phase only includes partial evaluation, but it will also include other steps, like constant propagation in the future (and probably other techniques as well). Is there a way to keep the "skipTactOptimizationPhase" for the entire phase, but to include sub-options for each of the steps in the optimization phase?

anton-trunov: We need to come up with a plan for those optimizations and design a config for those then

jeshecdom: I guess it is enough to have an option for each step inside the optimization phase and avoid the complication of having an option for the entire phase.

jeshecdom: > We need to come up with a plan for those optimizations and design a config for those then

Yeah. All right. I will change the option to skipPartialEval, and once we have a plan, I will carry out the changes in a separate issue/PR.

jeshecdom: I'll open an issue-task regarding this.

src/optimizer/optimization-phase.ts
anton-trunov, line 1: We should not unfreeze the AST. Produce a new AST instead.

anton-trunov (changes requested):

Issue #910, opened 2024-10-03
Document Tact calling convention


anton-trunov:
Related issue: https://github.com/ton-community/ton-docs/issues/782

novusnota:
Partially described on the Assembly functions page, but only for them: https://docs.tact-lang.org/book/assembly-functions/#conventions

We also need to mention the `DROP` of unused parameters in all non-assembly functions, except for those with `inline` attribute AFAIR

PR #1061, opened 2024-11-19
feat(docs): `asm` functions

### LINKED ISSUE #1011: Docs: describe `asm` functions
Depends on #888.

P.S.: Initially I thought to unite those two issues into one and submit it all together, but the resulting PR would be unwieldy, so it's better to split.

### LINKED ISSUE #1092: Wrong function signature for `fromSlice`
The signature in the docs is as follows:

```tact
extends fun fromSlice(self: Struct, cell: Slice): Struct;
```

But `fromSlice` is not called on any instance of `Struct`. `Struct` is better understood as a namespace here.

Incidentally, `cell: Slice` is confusing, it should be `slice: Slice`.

### LINKED ISSUE #1131: "Mutable functions" -> "Mutating functions"
The functions are actually not mutable themselves, they *mutate* things

### LINKED ISSUE #1142: Describe how to use TVM assembly from Tact
As of now, one needs to create a FunC function, and then import the file with it and wrap the function into a `native` one in Tact

Related to tact-lang/tact#1154 and tact-lang/tact#1145 

UPD: Tact 1.5 support `asm`-functions natively, no need to go via FunC

novusnota:
Rewrote the method ID collisions section to remove all logical jumps and make it much more streamlined :)

Also adjusted the structure a little towards the upcoming PR revamping this page. I'll push the draft of it right after we deal with asm functions here.

P.S.: I actually call argument to return position mappings "arrangements" and not "shuffle" as in `grammar.ohm`, because the latter kinda implies randomness, while those are actually deterministic. Hence, "asm arrangments".

## Issue

* Closes #1011.
* Closes https://github.com/tact-lang/tact/issues/1142 (old issue in `tact-docs`)

Also, resolved two teeny tiny issues from `tact-docs` — virtually 3-5 lines of fixes for each, no need in a separate CHANGELOG entry:
* Closes https://github.com/tact-lang/tact/issues/1131
* Closes https://github.com/tact-lang/tact/issues/1092

## Checklist


docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,175 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+
+:::caution
+
+  These are very advanced functions that require experience and vigilance in both definitions and usage. The logical errors in them are extremely hard to spot, the error messages are abysmal, and type checking isn't currently provided by Tact.
+
+  That said, if you know what you're doing, they can offer you the smallest possible gas usage, the best performance and the most control over [TVM][tvm] execution. Remember — with great power comes great responsibility.
+
+:::
+
+Assembly functions (or `asm{:tact}` functions for short) are module-level functions that allow writing [Fift and TVM assembly](https://docs.ton.org/v3/documentation/smart-contracts/fift/fift-and-tvm-assembly) directly in Tact. Unlike all other functions, their bodies consist only of [Fift words][fift] and [TVM instructions][tvm-instructions], and don't use any [Tact statements](/book/statements).
```
anton-trunov: we definitely do not want to give the impression Fift is supported, it's vice versa

we want to have intersection in terms of instruction syntax with Fift, but definitely not in terms of its other primitives, it's too powerful

let's get rid of any Fift mentions and perhaps fix our asm-grammar one more time :)

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,175 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+//                         Notice, that the body contains
+//                         only Fift words or TVM instructions
+```
+
+### Caveats {#asm-caveats}
+
+The difference between Fift words and TVM instructions is that words are processed and computed at [compile-time](/ref/core-comptime), while instructions are converted to opcodes and embedded into the contract as part of its code. They are also visually different: TVM instructions are almost all uppercase.
+
+```tact
+/// Defines ++ word to be used elsewhere
```
anton-trunov: definitely no `++` should be officially supported

docs/src/content/docs/book/functions.mdx
```diff
@@ -144,32 +315,34 @@ contract Treasure {
...
+
+Since method IDs are $19$-bit signed integers and some of them are reserved, only the inclusive ranges from $-2^{18}$ to $-5$ and from $2^{14}$ to $2^{18} - 1$ are free to be used by users. To avoid collisions, it's recommended to specify method IDs only in these ranges, avoiding the method IDs of Tact-specific getters mentioned above.
 
-User-specified method IDs are 19-bit signed integers, so you can use integers from $-2^{18}$ to $-5$ and from $2^{14}$ to $2^{18} - 1$.
+[slice]: /book/cells#slices
 
-Also, a few method IDs are reserved for the usage by the getters the Tact compiler can insert during compilation, those are 113617, 115390, 121275.
+[tvm]: https://docs.ton.org/learn/tvm-instructions/tvm-overview
+[tvm-instructions]: https://docs.ton.org/v3/documentation/tvm/instructions
+[fift]: https://docs.ton.org/v3/documentation/smart-contracts/fift/overview
```
anton-trunov:
```suggestion
```

docs/src/content/docs/book/import.mdx
```diff
@@ -50,6 +50,14 @@ import "./dns.fc";
...
 ```
 
+## Import Fift code
+
+Tact does not allow you to import Fift code using the `import{:tact}` keyword. However, you can write almost any Fift code in [`asm{:tact}` functions](/book/functions#asm). The only limitations compared to writing `.fif` files come from trying to redefine the right brace `}`, using it out of place or otherwise messing with it.
+
+Note, that the `Fift.fif` and `Asm.fif` libraries are imported and accessible by default, so you must not include those yourself.
+
+Read more about `asm{:tact}` functions in their dedicated section: [Assembly functions](/book/functions#asm).
+
```
anton-trunov:
```suggestion
```

anton-trunov (changes requested):

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,175 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+
+/// COMPILATION ERROR!
+/// You could've thought that you're providing the value of `boc` to the function
+/// and receiving it back, but actually there's a builtin Fift word `boc` which is
+/// being used instead.
+/// The `boc` would not be recognized as a parameter
+/// even if such word didn't exist in Fift!
+asm fun bocchiThe(boc: Cell): Cell { boc }
+
+/// Loads a signed `l`-bit integer from Slice `s`,
```
jeshecdom: Please use a different letter instead of letter `l`, because it looks like number one. The first time I read it as "one-bit integer", until I realized it was a parameter name.

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,175 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+/// and returns it with the remainder of `s`
+asm fun sliceLoadInt(s: Slice, l: Int): IntSlice { LDIX }
+//                   ↑         ↑
+//                   |         Placed on the stack last
```
jeshecdom: "Placed on" is a bit ambiguous here: Does it mean that parameter `l` will occur as second element in the stack as in this figure: `s, l, ....` (the first element is the top of the stack). Or does it mean that `l` is **pushed last** into the stack, so that the figure will be: `l, s, ....`. It is better to state everything in terms of the words "pushed into" instead of "placed on", since it is standard to use the push instruction when talking about stacks (also, the paragraphs before the example already talk about pushing elements into the stack).
 
However, talking about pushes forces the reader to keep an image in his head of how the stack is modified by the pushes. If we do not want to rely on that, we could say "`l` is placed at the top of the stack" and "`s` is placed as second element from the top" (or use the word "occurs" instead of "placed"). 

Honestly, I prefer to state everything in terms of pushes, because asm functions are supposed to be used by people that already know that the TVM is a stack, and are familiar with the concept of stack.

IMPORTANT: In all the comments that follow, I will assume the "push" interpretation: the declaration order in a function `test(a: Int, b: Int)` means that `a` is pushed first, and `b` is pushed last. Disregard everything if this is not the correct interpretation.

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,175 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+struct IntSlice { a: Int; b: Slice }
+```
+
+### Arrangements {#asm-arrangements}
+
+Sometimes it's useful to change the order of arguments pushed to the stack or the order of return values. You can do that with `asm{:tact}` arrangements in the following manner:
+
+```tact
+// Changing the order of arguments to match the STDICT signature
+asm(c self) extends fun asmStoreDict(self: Builder, c: Cell?): Builder { STDICT }
```
jeshecdom: A little bit of explanation is needed here. 

In the arrangement `asm(c self)`, the order of the variables describe the order in which they are going to be pushed into the stack: `c` first, `self` last, so that `self` will be at the top of the stack. In this example, the TVM instruction `STDICT` requires that its Builder argument occurs at the top of the stack. If we remove the arrangement `asm(c self)` and just leave `asm`, the order in which function `asmStoreDict` declared the arguments would place the cell argument `c` at the top of the stack. The arrangement `asm(c self)` fixes this by instructing the compiler to push Builder `self` last into the stack, so that it occurs at the top of the stack. 

Am I describing this correctly?

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,175 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+
+// Changing the order of return values of LDVARUINT16,
+// capturing only the 2nd one as the return value of the whole function
+asm(-> 1 0) extends mutates fun asmLoadCoins(self: Slice): Int { LDVARUINT16 }
```
jeshecdom: I keep forgetting the meaning of the syntax `-> 1 0`. I tried to read the FunC documentation but I still do not get what it does (it says that 0 is the deepest stack element among the returned values). 

Let me try to understand this. If `-> 0 1` means NO re-arrangement of return values, this means that popping the elements from the stack will coincide with reading the numbers from the right: 1 is popped first, 0 is popped last (because it is the deepest element). Therefore, reading `-> 1 0` from the right means to get the deepest element first and as second the element that was at the top of the stack.

So, if the returned type of the function is not a struct (as in the example, which is an `Int`), which popped element is the one that the function returns? The last one? So, in this case would be the 1-indexed element, right? i.e., the element immediately at the right of the arrow `-> [this one is returned] [....all these are dropped]`.

Is my reasoning correct?

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,175 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+
+// Changing the order of argument and return values
+asm(self len -> 1 0) extends fun asmLoadInt(self: Slice, l: Int): SliceInt { LDIX }
```
jeshecdom: Identifier `len` in the arrangement does not occur in the argument list of the function. Also, the arrangement is not really rearranging the function's arguments: they are in the same order. So the comment above the function: "Changing the order of argument and return values" is misleading.

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,175 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+    let sliceInt = s.asmLoadInt(10); // Slice remainder and 27
+}
+```
+
+### Attributes {#asm-attributes}
+
+The following attributes can be specified:
+
+* `inline{:tact}` — does nothing, since assembly functions cannot be inlined yet.
+* [`extends{:tact}`](#extension-function) — makes it the [extension function](#extension-function).
```
jeshecdom: "makes it the extension" -----> "makes it an extension".

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,175 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+* [`mutates{:tact}`](#mutation-functions) (along with [`extends{:tact}`](#extension-function)) — makes it the [extension mutation function](#mutation-functions).
```
jeshecdom: "makes it the extension mutation" -----> "makes it an extension mutation".

docs/src/content/docs/book/functions.mdx
novusnota: Will be removed with the whole Fift section, other caveats are to be adjusted for TVM instructions and questions from @jeshecdom

docs/src/content/docs/book/functions.mdx
novusnota: > Am I describing this correctly?

Yes

docs/src/content/docs/book/functions.mdx
novusnota: `-> 0 1` — the element denoted by `0` is second to top of the stack, the element denoted by `1` is on top of the stack. In general, the rightmost is the top of the stack, the leftmost is the bottom of the stack. Same as with params.

Now, after this mapping, we should think of the results we get as tuples — and then use or not use Structs to capture first `n` entries of the tuple, discarding the rest.

I'll try to address this and other notes in the text

docs/src/content/docs/book/functions.mdx
novusnota: Yup, "push" interpretation should be the preferred language there

docs/src/content/docs/book/functions.mdx
novusnota: Removed. Now, should we also restrict the grammar?

UPD: Yes, will be done in another PR for https://github.com/tact-lang/tact/issues/837

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,167 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+asm fun sliceLoadInt(s: Slice, len: Int): IntSlice { LDIX }
+//                   ↑         ↑
+//                   |         Pushed last, sits on top of the stack
+//                   Pushed first, sits on the bottom of the stack
+
+/// Maps onto values placed by LDIX on the stack
+struct IntSlice { a: Int; b: Slice }
+//                ↑       ↑
+//                |       Pushed last, sits on top of the stack
+//                Pushed first, sits on the bottom of the stack
```
jeshecdom: Sorry, why are the fields of a struct declaration pushed into the stack? I thought that such behavior was for asm functions. Or is this describing how the LDIX instruction pushes its results into the stack? If that is the case, I think it should be emphasized that they are results. Also, I would add a clarification note at the end. The following summarizes my suggestions:
```
struct IntSlice { a: Int; b: Slice }
//                ↑       ↑
//                |       Result pushed last by LDIX, sits on top of the stack
//                Result pushed first by LDIX, sits on the bottom of the stack
// 
// Hence: when popping the results from the stack, 
// struct fields are filled from right to left.
```
What do you think?

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,167 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+```
+
+The return values are provided bottom-up from the stack and the unused values are discarded.
+
+```tact
+// Same function as before, but now we don't use the `IntSlice` Struct
+// and instead only take one value from the stack (going bottom-up)
```
jeshecdom: "take one value from the stack (going bottom-up)".... which value is taken? The bottom one or the top one?. What about changing the line as follows:

"and instead only take the bottom result sitting in the stack"

I mention the above because a user may think that writing the function like this:

`asm fun sliceLoadInt(s: Slice, len: Int): Slice { LDIX }`

would return the `Slice`, not the `Int`, when in reality the function always returns the bottom result sitting in the stack (which is the `Int` in this case).

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,167 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+// `c` will be pushed first and get on the bottom of the stack,
+// while `self` will be pushed last and get on top of the stack
+asm(c self) extends fun asmStoreDict(self: Builder, c: Cell?): Builder { STDICT }
+
+// Changing the order of return values of LDVARUINT16,
+// capturing only the last one as the return value of the whole function
+asm(-> 1 0) extends mutates fun asmLoadCoins(self: Slice): Int { LDVARUINT16 }
+//     ---
+//     Notice, that return values are best thought as tuples with indexed access into them
+//     and not as bottom-up representation of stack values
```
jeshecdom: Probably we need to clarify this in a more detailed example (tutorial-like), because the connection with the stack is very vague and confusing. I can do such detailed example in a separate page, and then I can add a link in here to such page. I can work on that example as a separate PR, since now I am working on writing docs. What do you think? Do you suggest I should use the same TVM instructions you were using in these examples or should I use another TVM instruction that produces at least 3 or more results to exemplify the use of structs?

docs/src/content/docs/book/functions.mdx
novusnota: Agree with note on `LDIX`. Or, to generalize, "by instructions in the body".

Disagree with the last note — while it clarifies things here, we now implicitly ask people to think right-to-left when it comes to result values and the `-> 0 1`, `-> 1 0` and others get really confusing, IMO. Especially in combination the notes below. Thinking of something better...

docs/src/content/docs/book/functions.mdx
novusnota: I think that makes sense. If that's a tutorial-like example (for the Cookbook, I presume), then I'd use some other instruction with 3+ results to exemplify the use of Structs, exactly :)

UPD: Well, since my original answer I discovered that unused values are discarded only when the primitive type is specified as the return type. And now the idea of a tutorial makes much less sense to me — it's better that we improve descriptions here and now. But I may be totally wrong

docs/src/content/docs/book/functions.mdx
novusnota: Also, just before the Struct I've placed a note saying "/// Maps onto values placed by LDIX on the stack" :)

novusnota: Hold on, discovered a flaw in the logic, will rewrite a couple of things — see the edits

jeshecdom:
I agree that this is confusing, and I think that the cause is that we are mixing two mental models: a stack (low-level) and tuples (high-level).

We should pick only one and stick with it the entire explanation. It seems that describing everything in terms of tuples is more intuitive, but then we should not mention the stack (or mention how tuples are pushed and popped from a stack in a separate subsection, and only in that section mention the stack).

So, for example, I would start the examples saying something about the TVM instructions, something like this:

"
Even though TVM instructions work with a stack, TVM instructions can be seen, intuitively, as maps from tuples to tuples. To see how TVM instructions map tuples to tuples in the TVM stack, see [link: here]. Thinking in terms of tuples makes the explanation of asm functions much clearer, but for those who want to see an explanation using the stack directly, see [link: here].
"

Then, explaining the meaning of a declaration like:
```
asm(len self -> 1 0) fun testFun(self: Slice, len: Int): Result { TVM_INSTRUCTION }

struct Result {
   res1: Int;
   res2: Bool;
}
```

amounts to saying simply:

"
`testFun` passes the argument tuple `(len, self)` to instruction `TVM_INSTRUCTION`. Suppose `(r0, r1)` is the tuple result of `TVM_INSTRUCTION`, then `testFun` reorders the result according to the map `-> 1 0`, i.e., the 1-th index element (`r1`) has now index 0, and the 0-th index element (`r0`) has now index 1, producing the tuple `(r1, r0)`. Finally, `testFun` assigns the tuple `(r1, r0)` into the `Result` struct one field at a time, producing `Result {res1: r1, res2: r0}`.
"

novusnota:
@jeshecdom interesting note. The cases with multiple instructions should be covered too. And tuples on TON are denoted with square brackets `[]`, so it's better to use those. Also, it's best for readability to remove parentheses as much as possible, so things like "and the 0-th index element (`r0`)" will be "and the 0-th index element `r0`" — no need to add indirection and visual pauses with parens :)

@anton-trunov wdyt about https://github.com/tact-lang/tact/pull/1061#issuecomment-2489235137?

anton-trunov:
> TVM instructions can be seen, intuitively, as maps from tuples to tuples.

this is incorrect in a very specific technical sense: a tuple is a TVM data structure that occupy precisely one TVM stack position but can contain multiple other TVM primitives, including tuples

the term you probably intended to use is *tensor*

anton-trunov:
> testFun passes the argument tuple (len, self)

I find it confusing

anton-trunov:
@novusnota just adapt the corresponding calling convention description from tvm.pdf

anton-trunov:
@novusnota you also need to check how structures that are returned from a function are actually encoded

anton-trunov:
To *fully* finish this section, #910 needs to be resolved too

jeshecdom:
> > TVM instructions can be seen, intuitively, as maps from tuples to tuples.
> 
> this is incorrect in a very specific technical sense: a tuple is a TVM data structure that occupy precisely one TVM stack position but can contain multiple other TVM primitives, including tuples
> 
> the term you probably intended to use is _tensor_

I meant mathematical tuple, but now I see that this would introduce much more confusion because of the technical terms in TVM. So, the explanation should stick with the stack and use the technical terms in TVM.

jeshecdom:
> @jeshecdom interesting note. The cases with multiple instructions should be covered too. And tuples on TON are denoted with square brackets `[]`, so it's better to use those. Also, it's best for readability to remove parentheses as much as possible, so things like "and the 0-th index element (`r0`)" will be "and the 0-th index element `r0`" — no need to add indirection and visual pauses with parens :)
> 
> @anton-trunov wdyt about [#1061 (comment)](https://github.com/tact-lang/tact/pull/1061#issuecomment-2489235137)?

Yeah, we should stick with the stack explanation and use the technical terms in TVM, because I see everyone is confused now :). But the way, a question:

In a function like this:
```
asm fun testFun(a: Int, b: Int): Result { 
 INS_1
 INS_2 
 .....
 INS_n   // Let us suppose that after INS_n finishes, 
         // there are 5 results in the stack
}
```

Does Tact know that after executing those instructions, there will be exactly 5 results in the stack?
What happens if struct `Result` has more than 5 fields? Will Tact pop more than 5 elements from the stack until it fills the struct, having as consequence the popping of elements that are not part of the intended result?

anton-trunov:
> Does Tact know that after executing those instructions, there will be exactly 5 results in the stack?

not in the current implementation

> Will Tact pop more than 5 elements from the stack until it fills the struct

nope (the consequence of the previous answer)

This should be documented, of course, but an even more important question is "are returned structs actually represented as tensors (multiple TVM values)?"

anton-trunov:
and, of course, the symmetrical question for input function parameters (including passing structs)

anton-trunov:
in any case, each such point should be accompanied by a concrete example of an `asm`-function

novusnota:
> just adapt the corresponding calling convention description from tvm.pdf

Sure, the `-> 0 1` is much better explained in terms of s0..s255 stack registers.

> Does Tact know that after executing those instructions, there will be exactly 5 results in the stack?

Nope, at the moment it's all handled by FunC, which it turn just passes it to Fift, which does all the work. Neither Tact nor FunC check anything until it's too late and user hits exit code 5, 7, or whatever else.

Also, there could be more things in the stack, only the topmost 5 are of interest if we know that after all the instructions we need 5 values.

> This should be documented, of course, but an even more important question is "are returned structs actually represented as tensors (multiple TVM values)?"
> and, of course, the symmetrical question for input function parameters (including passing structs)
> in any case, each such point should be accompanied by a concrete example of an asm-function

👍

anton-trunov:
@jeshecdom when we have the grammar and AST for our embedded assembly language then we will be able to typecheck asm-functions and warn the user their stack discipline makes sense

jeshecdom:
> > just adapt the corresponding calling convention description from tvm.pdf
> 
> Sure, the `-> 0 1` is much better explained in terms of s0..s255 stack registers.
> 

We could do the following with specific examples (previously, we should have explained how function arguments are pushed into the stack):

- First, explain how the result tensor is popped from the stack. Suppose after popping, you get some tensor `(r0, r1, ..., rn)`.
- Explain how the notation `-> n m l` is just a re-arrangement of the result tensor.
- Explain how the re-arranged tensor is mapped into the result type of the function (this includes explaining how the tensor is mapped into a struct or a primitive type, or whatever).

novusnota:
Eh, Structs are represented as tensors `(...)`, but that's the doing of Tact+FunC. If we were to target TVM directly, we would've dealt with stack entries ourselves, all without tensors.

So I'm a bit hesitant on explaining things in tensors, and instead let's just properly describe stack and stack registers.

jeshecdom:
> Eh, Structs are represented as tensors `(...)`, but that's the doing of Tact+FunC. If we were to target TVM directly, we would've dealt with stack entries ourselves, all without tensors.
> 
> So I'm a bit hesitant on explaining things in tensors, and instead let's just properly describe stack and stack registers.

Sounds good.

anton-trunov:
> So I'm a bit hesitant on explaining things in tensors, and instead let's just properly describe stack and stack registers.

sounds good to me too, since tvm.pdf does not even mention tensors, looks like it's a term coined by the FunC community

novusnota:
1. Added description of the current Tact-flavored assembly from https://github.com/tact-lang/tact/pull/1064
1. Described everything from the stack point of view, including its "registers"
1. Refined the overall top-to-bottom reading flow

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,243 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+}
+```
+
+:::caution
+
+  The `i s()` syntax for referring to stack registers beyond the $0 - 15$ range is deprecated and recognized as an error in Tact 1.6. Whenever you see `[ii] s()` in the [TVM instructions list][tvm-instructions], use one of `s0`, `s1`, ..., `s255` instead.
+
+:::
+
+### Stack calling conventions {#asm-calling}
```
jeshecdom: The stack calling conventions section reads really nice. Just a couple of questions. What happens if one of the asm function arguments is a struct? What about an argument which is a struct with nested structs? And what happens if the return type is a struct with nested structs? like in this declaration:
```
struct A {
  a1: Int;
  a2: Int;
}

struct B {
  b1: Int;
  b2: A;
}

asm fun test(s: B, ...): B 
{ ....... }
```

docs/src/content/docs/book/functions.mdx
```diff
@@ -114,6 +116,243 @@ native storeUint(s: Builder, value: Int, bits: Int): Builder;
...
+
+Often times it's useful to change the order of arguments pushed to the stack or the order of return values without referring to stack registers in the body. You can do that with `asm{:tact}` arrangements in the following manner:
+
+```tact
+// Changing the order of arguments to match the STDICT signature:
+// `c` will be pushed first and get on the bottom of the stack,
+// while `self` will be pushed last and get on top of the stack
+asm(c self) extends fun asmStoreDict(self: Builder, c: Cell?): Builder { STDICT }
+
+// Changing the order of return values of LDVARUINT16,
```
jeshecdom: It is still not clear what the notation `-> 1 0` means regarding what happens to the results of `LDVARUINT16` **in** the stack itself. The explanation states that 1 represents the value of stack register 1, etc. but it does not explain the significance of writing them in the order `-> 1 0`. Probably what needs to be said is that the notation `-> 1 0` describes how the contents of the stack will be rearranged, when reading `-> 1 0` left-to-right: the contents of register `s1` will be placed at the top of the stack, and the contents of register `s0` will be placed second-to-top. 

One alternative way of explaining could be in terms of removing from the stack: `-> 1 0` means that `s1` is removed first, followed by `s0`.  Hence, the function returns the `Builder` in `s0` because it was the stack content removed last.

jeshecdom: Now I am having second thoughts on using "removing" because it becomes confusing with what happens with the rest of the stack. For example, suppose that after executing some asm function with declaration `-> 2 1 0`, we have the 5 element stack (top is leftmost):
```
a b c d e
```
Then, `-> 0 1 2` means "remove `s0`, then `s1`, then `s2`", so that the stack after removing `s0` is:
```
b c d e
```
But then, `s0` contains now `b`, when previously `b` was in `s1`.

So, probably a better word instead of "removing" would be "read from":

`-> 1 0` means that `s1` is read from the stack first, followed by `s0`. Hence, the function returns the `Builder` in `s0` because it was the stack content read last.

novusnota: Thing is, as I've just checked in tests, `-> 0 1 2` is not about taking or not taking any results, but merely about positioning items for the whatever result type we've specified. Like, if the return type is `Int`, one can only specify `-> 0` and nothing else, even though `-> 0` in this case is the same as not writing anything at all. And when the Structs, long Structs (more than 15 entries) or even nested Structs are involved, this is getting complicated.

Thus, my description of `s0` matching `0`, `s1` matching `1` is actually incorrect and has to be rewritten. And I've got to check the cases with long or nested Structs here as well, same as for the "stack calling conventions" bit.

jeshecdom: I see. So, this declaration is incorrect (because it returns only one element):
```
asm(self len -> 1 0) extends fun asmLoadInt(self: Slice, len: Int): Slice { LDIX }
```
but this is correct:
```
asm(self len) extends fun asmLoadInt(self: Slice, len: Int): Int { LDIX }
```
even though it will discard the `Slice` result and keep only the `Int`. Or is this last one also incorrect?

Mmmm.... very confusing indeed. So, when using the notation `-> m n p` it is not possible to discard values in the result type. I think this is acceptable. It is better to explicitly state all the results than to rely on understanding implicit discards.

novusnota: First one is incorrect. Second one could've been correct if we had our own backend or if we'd alter FunC generation, but since I tested that it's also incorrect — nothing can be discarded in result type.

It worked for me in previous tests mainly because FunC doesn't perform any checks, and because all asm function bodies are embedded in Fift code.

I had some DROP instructions very deep later on in other asm functions, which unexpectedly (for me) cleared the stack for this one. And I noticed that a little too late.

In the end, this really proves the point of those cautionary paragraphs at the top of the assembly functions description. This stuff is quite messy :)

jeshecdom: I understand and thank you for your effort! 

So, let's adapt the explanation so that no discards happen in the result type. 

Now, regarding nested structs, structs in arguments and structs with more than 15 fields, if you think that the explanation would become too complex to fit it in the page or that the explanation would become so convoluted because of those exceptional cases, probably it would be better to explain those in a separate page, with a link to that page.

novusnota: > First one is incorrect. Second one could've been correct if we had our own backend or if we'd alter FunC generation, but since I tested that it's also incorrect — nothing can be discarded in result type.

Now that I've tested more things I can say that "discarding" is possible, but must never be used in its current state, since all the unused stack entries are still there and can conflict with the stuff below.

Also, return values are always taken as a right-to-left mapping going from the topmost stack elements to the bottommost. I.e. if the primitive type is specified as a return value, then the topmost value of the resulting stack will take its place. And if the return type is some Struct, then it's last field will be the top value of the stack, the second-to-last field the second-to-top value of the stack and so on. Once again, "discarding" currently can be applied here too, which would just leave all unmapped stack entries on the stack and potentially conflict with subsequent expessions down the line.

Anyways, here's an example of integer "discarding":

```tact
struct TestNest { a: Int; b: TestBig; c: Int }
struct TestBig { f1: Int; f2: Int; f3: Int; f4: Int; f5: Int }

asm fun test(structure: TestNest): Int {
    DROP DROP
}

fun oneThing() {
    dump(
        test(
            TestNest{
                a: 1,
                b: TestBig{ f1: 2, f2: 3, f3: 4, f4: 5, f5: 6 },
                c: 100,
            }
        )
    ); // → 5, since the 100 and 6 were removed from the top of the stack by two DROPs
    // but 4, 3, 2, 1 are still there on stack and you won't notice them until
    // you add more code and realize that it consumes those unmapped stack values
    // instead of the intended arguments. That can be very evil if you
    // expected the values of the same type as those left on the stack :)
}
```

Oh, and a note on Structs — they are recursively flattened both for arguments and return values even if they're deeply nested or long (up to 16 fields without counting nested ones). That's why in the example above the result is 5 and not 1 as one would expect.

P.S.: Adding `mutates` attribute makes the asm function consume one more value from the stack after all the return values. I.e. the `asm(-> 1 0) extends mutates fun loadTestRef(self: Slice): Cell { LDREF }` will actually take the Cell first as the topmost element and then implicitly will take the Slice and assign it back to `self`. Whereas the non-mutates variant of the same asm function will leave that Slice on the stack untouched, which would pollute the stack for any subsequent expression in the function body down the line.

novusnota: Btw, going beyond 16 fields in Structs fails the compilation of FunC. @Gusarich, wasn't that issue resolved long time ago? Or, perhaps, the way we transpile asm functions differs from the rest of the functions that you've fixed?

Actually, now that I think about it, fixing that limit here would complicate things even further and asm functions are already quite involved. So we can just leave the limit of 16 fields of Structs or overall params per asm function as is. At least for now — that can be revisited later iff needed.

jeshecdom: Very dangerous indeed. In the future, once we have our embedded assembly language we will be able to add checks to asm functions, as Anton mentioned in a previous comment. For the moment, we just need to add to the documentation the extra stuff you mentioned (with examples would be better):
- Flattening of structs in arguments and results.
- Limit to 16 arguments and 16 results (counting flattened structs).
- Effect of mutating functions on the stack.

I think the limit of 16 arguments and results is not that bad. In principle, asm functions are used to program very specific, small procedures with simple logic. Is there anyone that would program almost the entire contract logic using asm instructions? I highly doubt it. I mean, the point of using Tact is precisely not to write everything using assembly language. But I agree that in the future, the limit of 16 could be changed.

novusnota: Packing notes from the weekend to finish this PR ASAP.

UPD: Final corrections, such as wrong description of `dumpStack()` or stack shenanigans between FunC and Tact
UPD2: Experimenting with better stack notations for code examples.
UPD3: Got to re-write final bits on return types and Structs, but just a little.

novusnota:
image

All done here, @anton-trunov @jeshecdom.

I'll move to finilizing #1064 and then to other things

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,678 @@
...
+}
+```
+
+:::note[Useful links:]
+
+  [List of TVM instructions in TON Docs][tvm-instructions]
+
+:::
+
+## Tact assembly {#tact}
```
anton-trunov: Let's remove this section for now, until we finalize the asm syntax

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,678 @@
...
+/// Simply returns back the value of `x`
+asm fun identity(x: Int): Int { }
+
+/// COMPILATION ERROR!
+/// The `BOC` is not recognized as a parameter,
+/// but instead is interpreted as a non-existent TVM instruction
+asm fun bocchiThe(BOC: Cell): Cell { BOC }
+```
+
+The parameters of arbitrary [Struct][struct] types are distributed over their fields, recursively flattened as the arguments are pushed onto the stack. In particular, the value of the first field of the [Struct][struct] is pushed first, the second is pushed second, and so on, so that the value of the first field is at the bottom of the stack and the value of the last is at the top. If there are nested structures inside those [Structs][struct], they're flattened in the same manner.
```
anton-trunov, line 103: Add a warning saying this is not guaranteed and subject to change in future Tact releases

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,678 @@
...
+
+When present, an assembly function's return type attempts to grab relevant values from the resulting stack after the function execution and any [result arrangements](#arrangements). If the return type is not present, however, the assembly function does not take any values from the stack.
+
+```tact
+// Pushes `x` onto the stack, increments it there,
+// but does not capture the result, leaving it on the stack
+asm fun push(x: Int) { INC }
+```
+
+Specifying a [primitive type][p], such as an [`Int{:tact}`][int] or a [`Cell{:tact}`][cell], will make the assembly function pop the top value from the stack and produce it as a result. If the run-time type of the popped value doesn't match the specified return type, an exception with [exit code 7](/book/exit-codes#7) will be thrown: `Type check error`.
```
anton-trunov: confusing description: "pop the top value from the stack and produce it as a result"

afaiu, no popping happens, also unclear that "produce" means here

better say something like nothing happens to `x` below and it just sits on top of the stack and will be used as the return value

should we also make it clear that the asm function below by itself won't produce exit code 7, but only when it's called and the return value is used as a cell

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,678 @@
...
+
+```tact
+// CAUSES RUN-TIME ERROR!
+// Pushes `x` onto the stack, then tries to capture it as a Cell,
+// causing an exit code 7: Type check error
+asm fun push(x: Int): Cell { }
+```
+
+Just like in [parameters](#conventions-parameters), arbitrary [Struct][struct] return types are distributed across their fields and recursively flattened in exactly the same order. The only differences are that they now capture or pop values from the stack instead of pushing them onto the stack, and they do so in a right-to-left fashion — the last field of the [Struct][struct] pops the topmost value from the stack, the second-to-last pops the second to the top, and so on, so that the last field contains the value from the top of the stack and the first field contains the value from the bottom.
```
anton-trunov: popping here is confusing ("pops" = "represents"?)

also, are we sure about flattening here?

anton-trunov (changes requested):
Partially reviewed

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,678 @@
...
+
+To see both the stack depth and the values on it, there's a function in the Core library of Tact: [`dumpStack(){:tact}`](/ref/core-debug#dumpstack). It's great for keeping track of the stack while debugging, although it's computationally expensive and only prints values, not returns them, so use it sparingly and only when testing.
+
+Read more about debugging Tact contracts on the dedicated page: [Debugging](/book/debug).
+
+## Attributes {#attributes}
+
+The following attributes can be specified:
+
+* `inline{:tact}` — does nothing, since assembly functions cannot be inlined yet.
```
anton-trunov: these are inlined by the FunC compiler regardless

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,678 @@
...
+    //                         |     sits right above the bottom of the stack
+    //                         Pushed first, sits on the bottom of the stack
+}
+```
+
+### Returns {#conventions-returns}
+
+When present, return type of an assembly function attempts to capture relevant values from the resulting stack after the function execution and possible stack [arrangements](#arrangements). When not present, however, assembly function does not take any values from the stack.
+
+When present, an assembly function's return type attempts to grab relevant values from the resulting stack after the function execution and any [result arrangements](#arrangements). If the return type is not present, however, the assembly function does not take any values from the stack.
```
jeshecdom, line 147: The paragraph at line 187 is almost the same as the paragraph at line 189. I prefer the paragraph at line 189.

By "grab relevant values" I guess it is meant that it will read the values from the stack and remove them from the stack, or does it mean that it will only read the values from the stack but it will not remove them from the stack?

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,678 @@
...
+```
+
+Just like in [parameters](#conventions-parameters), arbitrary [Struct][struct] return types are distributed across their fields and recursively flattened in exactly the same order. The only differences are that they now capture or pop values from the stack instead of pushing them onto the stack, and they do so in a right-to-left fashion — the last field of the [Struct][struct] pops the topmost value from the stack, the second-to-last pops the second to the top, and so on, so that the last field contains the value from the top of the stack and the first field contains the value from the bottom.
+
+```tact
+// Struct with two fields of type Int
+struct MinMax { minVal: Int; maxVal: Int }
+
+// Pushes `a` and `b` onto the stack,
+// then captures two values back via the `MinMax` Struct
```
jeshecdom, line 171: "captures" here is meant to mean "pops"? Similarly for previous examples.

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,678 @@
...
+```
+
+If the run-time type of some popped value doesn't match some specified field type of the [Struct][struct] or the nested [Structs][struct], if any, an exception with [exit code 7](/book/exit-codes#7) will be thrown: `Type check error`. Moreover, attempts to capture more values than there were on the stack throw an exception with [exit code 2](/book/exit-codes#2): `Stack underflow`.
+
+```tact
+// Struct with way too many fields for initial stack to handle
+struct Handler { f1: Int; f2: Int; f3: Int; f4: Int; f5: Int; f6: Int; f7: Int }
+
+// CAUSES RUN-TIME ERROR!
+// Tries to capture 7 values from the stack and map them onto the fields of `Handler`,
```
jeshecdom, line 182: "capture" again, and similarly in all the following examples.

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,678 @@
...
+```tact
+// Struct with two fields of type Int
+struct MinMax { minVal: Int; maxVal: Int }
+
+// Pushes `a` and `b` onto the stack,
+// then captures two values back via the `MinMax` Struct
+asm fun minmax(a: Int, b: Int): MinMax { MINMAX }
+```
+
+If the run-time type of some popped value doesn't match some specified field type of the [Struct][struct] or the nested [Structs][struct], if any, an exception with [exit code 7](/book/exit-codes#7) will be thrown: `Type check error`. Moreover, attempts to capture more values than there were on the stack throw an exception with [exit code 2](/book/exit-codes#2): `Stack underflow`.
```
jeshecdom: Probably add an example for the first sentence in paragraph 217, where the return type is incorrect:
```
asm fun sum(a: Int, b: Int): Bool { ADD }

fun test() {
   let b = sum(2,3) && true;  // Run-time error
}
```

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,678 @@
...
+struct Handler { f1: Int; f2: Int; f3: Int; f4: Int; f5: Int; f6: Int; f7: Int }
+
+// CAUSES RUN-TIME ERROR!
+// Tries to capture 7 values from the stack and map them onto the fields of `Handler`,
+// but there's just isn't that many values on the initial stack after TVM initialization,
+// which causes an exit code 2 to be thrown: Stack underflow
+asm fun overHandler(): Handler { }
+```
+
+As parameters and return values of assembly functions, [Structs][struct] can only have up to $16$ fields on each nested level. That is, you could go over $16$ fields in total if, for example, some of them were the nested structures, thus increasing the nesting level. Note that the outer parameters of the function also count as a nested level, so specifying a [Struct][struct] with more than $16$ fields as parameters or with more than $16$ parameters will cause compilation errors, especially in assembly functions.
```
jeshecdom: Probably more examples are needed here, because the explanation for "nesting levels" is a bit confusing. So, the following is OK because I do not exceed 16 fields (and function params) at each nested level?
```
struct A { field_1: Int ....... field_16: Int };

struct B { nested: A, F_1: Int, ..... F_15: Int};

asm fun test(n: B, param_1: Int, ..... param_15: Int) {
.........
}
```
But I will get a compilation error if I do either of the following actions: 
- Add an extra field to struct `A`, because there would be more than 16 fields in struct `A`.
- Add an extra field to struct `B`, because there would be more than 16 fields in struct `B`.
- Add an extra parameter to function `test`, because there would be more than 16 parameters in `test` (function parameters act as the outermost nested level).

I wonder if we could avoid the concept of nested levels in the explanation and say:
<<As parameters and return values of assembly functions, structs can only have up to $16$ fields. Each nested struct in those fields can also have up to 16 fields. This restriction also applies to asm function parameters, which can only have up to 16 parameters.>>

Also, I did not get what "especially in assembly functions" means.

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,678 @@
...
+// Seventeen fields
+struct S17 { f1:Int; f2:Int; f3:Int; f4:Int; f5:Int; f6:Int; f7:Int; f8:Int; f9:Int; f10:Int; f11:Int; f12:Int; f13:Int; f14:Int; f15: Int; f16: Int; f17: Int }
+
+// COMPILATION ERROR!
+asm fun chuckles(s: S17) { }
+```
+
+## Stack registers {#stack-registers}
+
+The so-called _stack registers_ are conventional way of referring to the values at the top of the stack. In total, there are $256$ stack registers, i.e. values held on the stack at any given time. You can refer to any of them using any of `s0`, `s1`, ..., `s255`.
```
jeshecdom: are conventional way ----> are a conventional way
or:
are conventional way ----> are the conventional way

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,678 @@
...
+  More about stack limits: [Limitations](#limitations).
+
+:::
+
+## Arrangements {#arrangements}
+
+Often times it's useful to change the order of arguments pushed to the stack or the order of return values without referring to stack registers in the body. You can do this with `asm{:tact}` arrangements — with them, the evaluation flow of the assembly function can be thought of in these $5$ steps:
+
+1. Function takes arguments in the order specified by the parameters.
+2. If an argument arrangement is present, arguments are reordered before being pushed to the stack.
```
jeshecdom, line 248: The examples that follow never make an explicit connection between the concepts "argument arrangement" and "result arrangement" described in these steps. The reader is forced to infer the connection in the examples.

In the examples, the connection must be made explicit:
```
// The argument arrangement in the declaration:
// asm(c self)
// changes the order of arguments to match the STDICT signature:
// `c` will be pushed first and get on the bottom of the stack,
// while `self` will be pushed last and get on top of the stack

asm(c self) extends fun asmStoreDict(self: Builder, c: Cell?): Builder { STDICT }

// The return arrangement in the declaration:
// asm(-> 1 0)
// changes the order of return values of LDVARUINT16 instruction,
// since originally it would place the modified Slice on top of the stack
asm(-> 1 0) extends fun asmLoadCoins(self: Slice): SliceInt { LDVARUINT16 }
......
......

// We can add both an argument and return arrangements, 
// to change the order of pushed arguments into the stack and 
// the order of returned values as well
asm(s len -> 1 0) fun asmLoadInt(len: Int, s: Slice): SliceInt { LDIX }

File: docs/src/content/docs/book/assembly-functions.mdx (hunk 18 above)
Comment (novusnota) [reply docs/src/content/docs/book/assembly-functions.mdx:103]: Do we not want that behavior for Structs? This behavior seems rather convenient.

To do something else, we'd need revisit the limit on number of parameters and, perhaps, tweak Structs into Tuples underneath, which will break compatibility with FunC code. And I think we'd need to preserve the ability to import it even when we'll have the new backend — there's a lot of FunC code and it's a nice escape hatch for doing all kinds of hacks on the side (like invoking Fift in FunC's asm functions or working with continuations a bit easier).

Currently, tuples are a totally different thing barely related to structures, even emulating them via the `Struct?` hack doesn't work in all contexts — [local bindings don't type check](https://github.com/tact-lang/tact/issues/1127#issuecomment-2549296989).

File: docs/src/content/docs/book/assembly-functions.mdx (hunk 20 above)
Comment (novusnota) [reply docs/src/content/docs/book/assembly-functions.mdx]: > also, are we sure about flattening here?

Wdym? It's the current behavior. Do we not want to capture multiple values from the stack at once, individually, and not as a tuple? That'd be detrimental to power-users with zero benefit to regular users, IMHO

File: docs/src/content/docs/book/assembly-functions.mdx (hunk 18 above)
Comment (novusnota) [reply docs/src/content/docs/book/assembly-functions.mdx:103]: > Add a warning saying this is not guaranteed and subject to change in future Tact releases

Added a note right after that paragraph

File: docs/src/content/docs/book/assembly-functions.mdx (hunk 20 above)
Comment (anton-trunov) [reply docs/src/content/docs/book/assembly-functions.mdx]: ok, the flattening question has been resolved; let's figure out what to do with the "pops" terminology here

Comment (novusnota) [reply docs/src/content/docs/book/assembly-functions.mdx]: I've removed that :)

File: docs/src/content/docs/book/assembly-functions.mdx [hunk 29]
```diff
@@ -0,0 +1,636 @@
...
+
+```tact
+// Changing the order of arguments to match the STDICT signature:
+// `c` will be pushed first and get on the bottom of the stack,
+// while `self` will be pushed last and get on top of the stack
+asm(c self) extends fun asmStoreDict(self: Builder, c: Cell?): Builder { STDICT }
+
+// Changing the order of return values of LDVARUINT16 instruction,
+// since originally it would place the modified Slice on top of the stack
+asm(-> 1 0) extends fun asmLoadCoins(self: Slice): SliceInt { LDVARUINT16 }
```
jeshecdom, line 267: In the 4th step above, it says "resulting values are reordered on the stack". The example should make clear that the syntax `-> 1 0` means that the reordering will happen left-to-right: the contents of `s1` will be returned first, followed by the contents of `s0`.

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,636 @@
...
+    1 -1 SETCONTARGS // ← this blows up
+}
+
+fun exitCode3() {
+    // Overflows the inner stack of a continuation
+    stackOverflow();
+}
+```
+
+Although there are only $256$ [stack registers](#stack-registers), the stack itself can have more than $256$ values on it in total. The deeper values won't be immediately accessible by any [TVM instructions][tvm-instructions], but they would be on the stack nonetheless.
```
jeshecdom, line 352: If an attempt to push more than 256 values into the stack causes a stack overflow (as described in paragraph at line 313), how is it possible that the stack could have more than 256 values? Could it be initialized in some special way that does not check for overflow?

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,636 @@
...
+    // Drops way more elements from the stack
+    // than there were before, causing an underflow
+    repeat (100) { drop() }
+}
+```
+
+Attempts to push more than $256$ values onto the stack at once throw an exception with [exit code 3](/book/exit-codes#3): `Stack overflow`. For example, this might happen when you specify a lot of nested [Structs][struct] as parameters to the assembly function. This upper limit applies to the [TVM][tvm] stack itself, corresponding to the [return continuation in register `c0`](https://docs.ton.org/v3/documentation/tvm/tvm-overview#control-registers), but different [continuations](https://docs.ton.org/v3/documentation/tvm/tvm-overview#tvm-is-a-stack-machine) may have a different upper limit of values on their inner stacks.
+
+```tact
+asm fun stackOverflow() {
```
jeshecdom, line 337: I did not understand what this example is doing =)

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,636 @@
...
+
+There, `LDREF` instruction produces two stack entries: a [`Cell{:tact}`][cell] and a modified [`Slice{:tact}`][slice] in that order, with the [`Slice{:tact}`][slice] pushed on top of the stack. Then, the arrangement `-> 1 0{:tact}` inverses those values, making the [`Cell{:tact}`][cell] sit on top of the stack.
+
+Finally, the [`mutates{:tact}`](/book/functions#mutation-functions) attribute makes the function consume the deepest value on the stack, i.e. [`Slice{:tact}`][slice], and assign it to `self{:tact}`, while returning the [`Cell{:tact}`][cell] value to the caller.
+
+Overall, [`mutates{:tact}`](/book/functions#mutation-functions) attribute can be useful in some cases, but you must stay vigilant when using it with assembly functions.
+
+:::caution
+
+  If the type of the deepest value on the stack won't match the type of `self{:tact}`, then an exception with [exit code 7](/book/exit-codes#7) will be thrown: `Type check error`.
```
jeshecdom, line 401: But the exception will only occur only if the function is called and we attempt to use the returned `self` in an incorrect way, right?

For example, in the following, would a call to `test` throw a type check error?
```
// In the result, Slice is at the top of the stack, so that 
// self gets assigned the Cell result.
asm extends mutates fun loadRef(self: Slice): Slice { LDREF }   

fun test(sl: Slice) {
   let a = loadRef(sl);
}
```

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -0,0 +1,636 @@
...
+asm extends mutates fun skipBits(self: Slice, l: Int) {
+    SDSKIPFIRST
+}
+```
+
+## Interesting examples {#examples}
+
+On the [TVM instructions][tvm-instructions] page, you may have noticed that the "signatures" of instructions are written in a special form called _stack notation_, which describes the state of the stack before and after the given instruction is executed.
+
+For example, `x y - z` describes an instruction that grabs two values `x` and `y` from the stack, with `y` at the top of the stack and `x` second to the top, and then pushes the result `z` onto the stack. Notice that other values deeper down the stack are not accessed.
```
jeshecdom, line 459: Is "grabs" here the same as "takes" or "pops"?

docs/src/content/docs/book/assembly-functions.mdx
novusnota: From https://docs.ton.org/v3/guidelines/smart-contracts/testing/writing-test-examples#compute-phase--exit-code--3:

> Any code before execution becomes a continuation. This is a special data type that contains a slice with code, a stack, registers, and other data necessary for the execution of the code. Such a continuation can be run later if necessary, passing the necessary parameters for the initial state of the stack.

Tact doesn't have a direct way of working with continuations in its type system, but an overflow is hard to showcase without them. Moreover, that example is listed on the exit codes page for quite some time now.

docs/src/content/docs/book/assembly-functions.mdx
novusnota: One cannot call `loadRef` that way in Tact, we'll issue a compilation error

docs/src/content/docs/book/assembly-functions.mdx
novusnota: Yes. @anton-trunov didn't like "pops", so I wrote "grabs", "takes" and "captures" throughout the page.

In general, there are no "pops" because there's no such explicit action — there are no instructions placed for that, the values are just mapped/captured/taken and used, without popping them from stack in the conventional sense of that verb.

The DROPs and PUSHINT or similar are explicit, but the pops — not. Some POP instructions exists merely as an alias to DROP, other do behave a bit close, but ultimately nothing gets off the stack somewhere else since there's no somewhere else. All other derivatives, such as POPCTR or similar are, essentially, DROPs with extra steps / checks.

docs/src/content/docs/book/assembly-functions.mdx
novusnota: There's an important "at once"! If values are pushed in lesser quantities, things work fine.

```tact
asm fun push42() { 42 PUSHINT }
asm fun depth(): Int { DEPTH }
asm fun drop() { DROP }

fun showcase() {
    dump(depth()); // either 3 or 2, depending on the call site: receiver or getter
                   // but there's a section that tells not to depend upon
                   // and don't use those initial stack values, so ignore them
                   
    repeat (255) { push42() } // stack now has +255 42's
    dump(depth()); // 3 + 255 or 2 + 255 (note above)
    
    // cleanup
    repeat (255) { drop() }
}

File: docs/src/content/docs/book/assembly-functions.mdx (hunk 25 above)
Comment (novusnota) [reply docs/src/content/docs/book/assembly-functions.mdx]: I've wrote an example for the case of primitive type, thinking that adding a very similar one for Structs is useless. If I'm wrong and it's needed, press "unresolve" :)

File: docs/src/content/docs/book/assembly-functions.mdx (hunk 24 above)
Comment (novusnota) [reply docs/src/content/docs/book/assembly-functions.mdx:182]: https://github.com/tact-lang/tact/pull/1061#discussion_r1890986747

File: docs/src/content/docs/book/assembly-functions.mdx (hunk 23 above)
Comment (novusnota) [reply docs/src/content/docs/book/assembly-functions.mdx:171]: Yes https://github.com/tact-lang/tact/pull/1061#discussion_r1890986747

File: docs/src/content/docs/book/assembly-functions.mdx (hunk 22 above)
Comment (novusnota) [reply docs/src/content/docs/book/assembly-functions.mdx:147]: Reads and removes. Previously, I wrote pops, but https://github.com/tact-lang/tact/pull/1061#discussion_r1890986747

File: docs/src/content/docs/book/assembly-functions.mdx (hunk 30 above)
Comment (jeshecdom) [reply docs/src/content/docs/book/assembly-functions.mdx:352]: I see. Thank you. 👍

File: docs/src/content/docs/book/assembly-functions.mdx (hunk 33 above)
Comment (novusnota) [reply docs/src/content/docs/book/assembly-functions.mdx:459]: That said, I'll add a note about this somewhere in the page. Something along the lines of (without the continuations bit):

There's no popping of values from the TVM stack to somewhere else. There could be nested stacks of continuations, but that's a slightly different story.

The main stack of the TVM is _the_ environment, _the_ entity that Tact operates within.

File: docs/src/content/docs/book/assembly-functions.mdx (hunk 28 above)
Comment (novusnota) [reply docs/src/content/docs/book/assembly-functions.mdx:248]: Yeah, I'll split the code examples and move text descriptions out of the comments

File: docs/src/content/docs/book/assembly-functions.mdx (hunk 26 above)
Comment (novusnota) [reply docs/src/content/docs/book/assembly-functions.mdx]: > So, the following is OK because I do not exceed 16 fields (and function params) at each nested level? ...code block...

Yes

> But I will get a compilation error if I do either of the following actions:
> 
> Add an extra field to struct A, because there would be more than 16 fields in struct A.
> Add an extra field to struct B, because there would be more than 16 fields in struct B.
> Add an extra parameter to function test, because there would be more than 16 parameters in test (function parameters act as the outermost nested level).

Yes

> I wonder if we could avoid the concept of nested levels in the explanation and say:
> <<As parameters and return values of assembly functions, structs can only have up to 
> 16
>  fields. Each nested struct in those fields can also have up to 16 fields. This restriction also applies to asm function parameters, which can only have up to 16 parameters.>>

Hmm, maybe

> Also, I did not get what "especially in assembly functions" means.

"especially" there probably has to be replaced by "only" — the intent was to say that there are no such limitation in non-asm functions, but only because of #590. And, on that note, the Structs are encoded different in different places of Tact, which isn't what we want to document (probably), at least until we sort it out with #1127, #458 and related.

And even in the same place, using or not using optionals will yield different results:

```tact
struct Two { a: Int; b: Int }

// In FunC, expects a tensor, (a, b)
asm fun two1(structure: Two) { }

// Notice the ?
// In FunC, expects a tuple, [a b] (or just `tuple` type)
asm fun two2(structure: Two?) { }
```

anton-trunov (approved):
LGTM, let's merge this PR and request more edits with separate doc issues. @jeshecdom if you find anything that's worth further clarification please open new issues with `docs.tact-lang.org` tag

PR #1062, opened 2024-11-20
feat(docs): new functions from TVM upgrades 2023-07 and 2024-04

### LINKED ISSUE #909: Document stdlib functions for the new TVM instructions from 2023.07 and 2024.04 upgrades
https://github.com/tact-lang/tact/pull/331

novusnota:
## Issue

Closes #909.

## Checklist


novusnota:
I'll finish this first thing in the morning, but now I'll go sleep a little

UPD: It took a while, but things look good now.

docs/src/content/docs/ref/core-advanced.mdx
anton-trunov, line 1: I would say my only problem with this PR is that it lacks practically-oriented examples with the new functions

anton-trunov (changes requested):
-

docs/src/content/docs/ref/core-advanced.mdx
novusnota: Well, for `getXFee()` functions, where `X` is either `Compute`, `Storage`, `Forward`, `SimpleCompute` or `SimpleForward`, it's tough to provide any real-world examples without, essentially, rewriting all calculation flow descriptions of #1068 and the related page in TON Docs: https://docs.ton.org/v3/guidelines/smart-contracts/fee-calculation.

I may be wrong here, but I think my assessment is correct and thus the examples have been kept to a minimum — just describing the syntax plus some usage hints, like calculating the `flat_price` or using `context().readForwardFee()`.

docs/src/content/docs/ref/core-advanced.mdx
```diff
@@ -29,6 +171,110 @@ Usage example:
...
+let fee: Int = getForwardFee(1_000, 1_000, false);
+//                           -----  -----  -----
+//                           ↑      ↑      ↑
+//                           |      |      Both source and destination
+//                           |      |      isn't on the masterchain,
+//                           |      |      but on the basechain
+//                           |      Number of bits in a message
+//                           Number of cells in a message
+```
+
```
Shvandre, line 210: Maybe we should add a mention of the get_data_size family of functions here? (And add them to the tact stdlib accordingly)

docs/src/content/docs/ref/core-advanced.mdx
```diff
@@ -29,6 +171,110 @@ Usage example:
...
+If both the source and the destination addresses are in the [basechain][basechain], then specify `isMasterchain` as `false{:tact}`. Otherwise, specify `true{:tact}`.
+
+Note, that the values of `cells` and `bits` are taken modulo their maximum values plus $1$. That is, specifying values higher than those listed in [account state limits (`max_msg_cells` and `max_msg_bits`)](/book/exit-codes#50) will have the same result as with specifying the exact limits. In addition, make sure you take [deduplication of cells with the same hash][deduplication] into account.
+
+However, regardless of the values of `cells` and `bits`, this function always adds the minimum price based on the value of [`lump_price`][param-24-25]. See the example for [`getSimpleForwardFee(){:tact}`](#getsimpleforwardfee) to derive it.
```
Shvandre: Also one should note that root cell and its bits are not counted in fwd_fee. Fees for that cell are covered by lump_price

docs/src/content/docs/ref/core-advanced.mdx
```diff
@@ -29,6 +171,110 @@ Usage example:
...
+
+<Badge text="Available since Tact 1.5" variant="tip" size="medium"/><p/>
+
+```tact
+fun getOriginalFwdFee(fwdFee: Int, isMasterchain: Bool): Int;
+```
+
+Calculates and returns the so-called _original_ [forward fee][forward-fee] in [nanoToncoins](/book/integers#nanotoncoin) [`Int{:tact}`][int] for an outgoing message based on the `fwdFee` obtained from the incoming message. If both the source and the destination addresses are in the [basechain][basechain], then specify `isMasterchain` as `false{:tact}`. Otherwise, specify `true{:tact}`.
+
+This function is useful when the outgoing message depends heavily on the structure of the incoming message, so much so that you cannot fully predict the fee using [`getForwardFee(){:tact}`](#getforwardfee) alone.
```
Shvandre: The official documentation says the same thing, but as far as I know the point is that such an instruction costs only 26 gas, as does GETFORWARDFEE, but GETFORWARDFEE requires a number of cells, and bits, for accurate counting, and CDATASIZE is not cheap. So if you need an approximate value rather than a nanotone precision, and the message contains almost the same as the incoming message, GETORIGINALFORWARDFEE is a good way to go

docs/src/content/docs/ref/core-advanced.mdx
novusnota: Do you mean the [`compute_data_size()`](https://github.com/ton-blockchain/ton/blob/7151ff26279fef6dcfa1f47fc0c5b63677ae2458/crypto/smartcont/stdlib.fc#L189C17-L189C34) (i.e. [`CDATASIZE`](https://docs.ton.org/v3/documentation/tvm/instructions#F941)) and alike?

Shvandre: Yes

novusnota:
If there are no more suggestions, let's merge this now and add real-world examples of fee calculation in #1068 in the future (as I [suggested here](https://github.com/tact-lang/tact/pull/1062#discussion_r1855476112)).

Wdyt, folks @anton-trunov @Gusarich @Shvandre?

anton-trunov:
> If there are no more suggestions, let's merge this now and add real-world examples of fee calculation in https://github.com/tact-lang/tact/issues/1068 in the future (as I https://github.com/tact-lang/tact/pull/1062#discussion_r1855476112).

sure

anton-trunov (approved):

PR #1064, opened 2024-11-21
WIP feat: new `asm` parser

### LINKED ISSUE #837: Better parser for `asm` block bodies

### LINKED ISSUE #1030: Malformed asm functions pass Tact compiler but fail during FunC compilation stage

### Tact source code

```tact
import "@stdlib/deploy";

asm fun myCode(): Cell { "MYCODE" }

contract Sample {
    mycode: Cell;
    init() {
        self.mycode = myCode();
    }
}
```

```shell
yarn run v1.22.22
$ tact --config ./tact.config.json
💼 Compiling project sample ...
   > Sample: tact compiler
   > Sample: func compiler
Func compilation error: /home/b4cksl4sh/TON-Studio/tact-template/sources/output/sample_Sample.code.fc:13:47: error: string extends past end of line
  cell $global_myCode() impure asm """"MYCODE"""";
error Command failed with exit code 30.
```

### What happened?

_No response_

### What did you expect?

As asm functions are declared in funC in "", it might be the common mistake. Maybe tact compiler should check that? Or change the format of asm functions, because static analizers are mad at them.

### Steps to reproduce

_No response_

### How do you run Tact?

tact-template

### Anything else?

_No response_

novusnota:
I don't know if this should be rewritten before or after the non-Ohm parser. That said, I'll start describing examples of TVM instructions and arguments that they expect, such that we can come up with a number of primitives to express those arguments.

The things people are used to are:
* `42 -42` — decimal number literals, with in-between underscores `_` allowed for readabilty (underscores `_` aren't allowed in Fift, but may be allowed in Tact)
* `x{babecafe_}` — hexadecimal bitstrings (up to 1023 bits) with optinal padding via the `_` at the end.
* `b{010101}` — binary bitstrings (up to 1023 bits) without padding.
* "string literals" — primarily for use in debug-related instructions, so we might remove those
* `s0`, `s1`, ..., `s15` — stack registers
* `c0`, `c1`, ..., `c15` — control registers
* `MYCODE ADDRSHIFT#MOD IF: XCHG3_l 2SWAP -ROT` — TVM instructions themselves. By the way, the ones ending with `_l` are useless and just legacy; most of those starting with a number have an alias where they end with that number (`2DROP` or `DROP2`, etc.); and `IF:` and related control flow-ish ones are used with continuations. Don't confuse the latter with `IF:<{`, which are purely Fift-guided and aren't TVM instructions per se.

The new-ish syntax proposals:
* `c{...}`, a 1-to-1 replacement of Fift's `B{...} B>boc`, which are hex-encoded BoCs with no limit on bitsize, except for the account state limits of course. Quite important for PUSHREF and similar instructions.
* Extra stack registers: instead of deprecated `i s()` syntax, users now can use any of `s0`, `s1`, ..., up to `s255`

The important bit is to parse deprecated things or recognize instuctions inside strings "" to provide clear error messages for users, with simple migration from the old syntax of Fift to new one of Tact.

---

This is another attempt at making the `asm` parser — now we're deliberately limiting ourselves to TVM instructions. In the current iteration it's still possible to write code in a very limited subset of Fift, but only if we won't restrict the list of supported TVM instructions in `grammar.ts`, i.e. during semantical actions on the lexed tokens.

TODO:
- [x] check that I'm going in the right direction before proceeding further (@anton-trunov)
- [x] `grammar.ohm`
- [x] `ast.ts`
- [x] `grammar.ts`
- [ ] `prettyPrinter.ts` → `writeFunction.ts` (to actually have asm instructions written)
- [ ] `hash.ts`
- [ ] `compare.ts`

## Issue

Towards #837.
Towards #1030 (once the `grammar.ts` is written, there will be a small check for such gotchas).

## Checklist


anton-trunov:
Let's start with concrete examples of asm-functions we would like to write and then try to develop some grammar for it. As of now, we should start with instruction sequences and comments (let's use the Tact/Fift syntax for comments).

novusnota:
> we should start with instruction sequences and comments

i.e. no `WHILE:<{ … }>DO<{ … }>` loops and no similar constructs for now, right?

anton-trunov:
> i.e. no WHILE:<{ … }>DO<{ … }> loops and no similar constructs for now, right?

definitely not using this syntax :)

anton-trunov:
we might allow asm-blocks to avoid introducing high-level constructions into the assembly language, so the user might able to insert those into loops, etc.

novusnota:
Hmm, then, something like this we've [showed earlier](https://t.me/tact_kitchen/34) (and have in our tests) won't work anymore:

```tact
asm fun isIntAnInt(x: Int): Bool {
    <{
        TRY:<{
            0 PUSHINT ADD DROP -1 PUSHINT
        }>CATCH<{
            2DROP 0 PUSHINT
        }>
    }>CONT 1 1 CALLXARGS
}
```

Proceed with removing that? This kinda goes against [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

UPD: Edited the original post, saying that the full range of Fift+TVM syntax there was highly experimental and subject to change in next releases. Ok, now we can proceed.

anton-trunov:
we haven't documented it, so there is no backwards compatibility violation

anton-trunov:
and Tact Kitchen posts are about raw immature things, sneak peek

src/prettyPrinter.ts
```diff
@@ -480,9 +480,55 @@ export const ppAstAsmFunctionDef: Printer<A.AstAsmFunctionDef> =
             c.row(
                 `asm${ppAsmShuffle(node.shuffle)} ${ppAstFunctionSignature(node)} `,
             ),
-            ppAsmInstructionsBlock(node.instructions)(c),
+            ppAsmExpressionsBlock(node.expressions)(c),
         ]);
 
+// NOTE: ok, I might need some help here...
+export const ppAstAsmExpressionList: Printer<A.AstAsmExpressionList> =
```
verytactical, line 487: `c.braced(node.exprs.map(ppAstAsmExpression))`

verytactical: Depending on the order PRs get merged, you might not have this problem here at all :)

src/grammar/grammar.ohm
```diff
@@ -99,21 +99,40 @@ Tact {
...
+    // for people coming from FunC/Fift background,
+    // gently guiding them towards simpler s0...s255 syntax
+    asmStackReg256 (indexed stack register) = digit digit? digit? whiteSpace+ "s()"
+    asmNumber      (number)                 = "-"? digit+ ~(letterAscii | "_" | "#" | ":")
+
+    // To ensure correct parse:
+    // 1. Instructions cannot contain braces
+    // 2. Cannot be braces themselves
+    // 3. Cannot contain lowercase letters, except for the very last position
+    asmInstruction (TVM instruction) = ~(("{" | "}") ~asmWord) asmWord
```
verytactical, line 133: Unfortunately all newline characters get parsed as `spaces` and are completely erased from AST. There is no way to pretty-print this at all.

I can imagine only one way to make this work without adding `;` or some other explicit way to terminate lines: put this into a separate grammar with its own `spaces` rule.

src/prettyPrinter.ts
novusnota: Thanks, but I've already removed the asm expression lists from the grammar, such that `{ ... }` are prohibited. At least until we may or may not make sense of them ourselves and without Fift's aid.

Didn't remove it there in pretty printer yet, but will.

src/grammar/grammar.ohm
novusnota: Hmm, I'm thinking of the Prettier's approach with trying to fit as much as allowed in 80, 120 or other column limits, and put all excesses on new lines. Or, to simply put a newline character after each instruction — to keep all prior primitives required for it on the same line.

The second option sounds best from legibility perspective as well:

```tact
asm fun showcase(a: Int, b: Int, c: Int): Int {
    s0 s1 s2 XCHG3
    s1 s2 s0 XCHG3
    DROP2
}
```

novusnota: Oh, and it would be nice to also check if the instructions are written on the same line with the opening and closing braces `{ ... }`, i.e. when the start row of the function body equals its end row. Because if that's the case, we should respect the author's intent and inline everything there: add spaces after instructions, not newlines.

novusnota:
Closing this in favor of: #1258. We should discuss the resulting assembly syntax a bit more — for example, besides https://github.com/tact-lang/tact/pull/1064#issue-2677607203, we might need the `<{ ... }>` (creates a Builder at comptime) to allow expressing anything beyond the top-level declarations (`0 DECLPROC name_name_name`) directly in TVM instructions and not as large hex-encoded bitstrings.

For example, when working on assembly functions page in the docs, I've replaced this:

`<{ }>CONT`

with

`x{} SLICE BLESS`,

where the former is a comptime-resolved expression by Fift, and the latter is exclusively processed by TVM — `SLICE` is an alias to `PUSHSLICE` instruction with one argument, and `BLESS` converts the Slice from the stack to the continuation.

Now, you could imagine that the `x{...}` could've been much larger and it could even be replaced with `B{...} B>boc` to overcome the limits of 128 bytes of the former, but it would be nigh-impossible to read and understand which instructions are encoded with them, so some macros like `<{ ... }>` (with `Builder { ... }` as an alternative syntax, for example) are needed.

Issue #255, opened 2024-04-14
Add resulting wrappers/bindings for other languages


novusnota:
At the moment, after the full compilation (regular one, see #254) only the TypeScript `.ts` wrapper is produced.

It would be nice to additionally support wrappers for:
* [Dart](https://github.com/novusnota/tonutils-dart)
* [.NET](https://github.com/continuation-team/TonSdk.NET)
* [Golang](https://github.com/xssnick/tonutils-go)
* etc.

This should probably be done via some extension mechanism or external templates to reduce the maintanance work on the Tact's side — those wrappers/bindings should be maintained by the authors of the respective language libraries for TON.

PR #1106, opened 2024-12-02
fix(ts-wrappers): fix TypeScript wrappers generation for messages with single quote

### LINKED ISSUE #972: Invalid string escaping in generated typescript code

### Tact source code

```tact
contract Test{
 receive("test'"){
    self.reply("OK");
 }
}
```

_No response_

### What happened?

Tact compiler generated incorrect typescript code:
```typescript
    async send(provider: ContractProvider, via: Sender, args: { value: bigint, bounce?: boolean| null | undefined }, message: 'cancel'') {
     //blablabla     
    }


```

### What did you expect?

I expected, that the generated typescript strings would contain escaped contents

### Steps to reproduce

_No response_

### How do you run Tact?

Blueprint

### Anything else?

_No response_

i582:
Hey, this particular bug is not that important, of course, but it's still a great option to dig into the project :)

I haven't found any test cases for TS wrappers. Does Tact plan to create (or have) one?

Fixes #972

### DIFF
--- a/src/bindings/writeTypescript.ts
+++ b/src/bindings/writeTypescript.ts
@@ -352,12 +352,9 @@ export function writeTypescript(
                                 r.message.text !== null &&
                                 r.message.text !== undefined
                             ) {
-                                // since we are using single quotes, escape them in the message name
-                                const escaped = r.message.text.replace(
-                                    /'/g,
-                                    "\\'",
+                                receivers.push(
+                                    `"${JSON.stringify(r.message.text).slice(1, -1)}"`,
                                 );
-                                receivers.push(`'${escaped}'`);
                             } else {
                                 receivers.push(`string`);
                             }
@@ -424,12 +421,9 @@ export function writeTypescript(
                                     });
                                     w.append(`}`);
                                 } else {
-                                    // since we are using single quotes, escape them in the message name
-                                    const escaped = msg.text.replace(
-                                        /'/g,
-                                        "\\'",
+                                    w.append(
+                                        `if (message === "${JSON.stringify(msg.text).slice(1, -1)}") {`,
                                     );
-                                    w.append(`if (message === '${escaped}') {`);
                                     w.inIndent(() => {
                                         w.append(
                                             `body = beginCell().storeUint(0, 32).storeStringTail(message).endCell();`,
--- a/src/bindings/writeTypescript.ts
+++ b/src/bindings/writeTypescript.ts
@@ -352,9 +352,7 @@ export function writeTypescript(
                                 r.message.text !== null &&
                                 r.message.text !== undefined
                             ) {
-                                receivers.push(
-                                    `"${JSON.stringify(r.message.text).slice(1, -1)}"`,
-                                );
+                                receivers.push(JSON.stringify(r.message.text));
                             } else {
                                 receivers.push(`string`);
                             }
@@ -422,7 +420,7 @@ export function writeTypescript(
                                     w.append(`}`);
                                 } else {
                                     w.append(
-                                        `if (message === "${JSON.stringify(msg.text).slice(1, -1)}") {`,
+                                        `if (message === ${JSON.stringify(msg.text)}) {`,
                                     );
                                     w.inIndent(() => {
                                         w.append(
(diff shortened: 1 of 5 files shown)


novusnota:
> Hey, this particular bug is not that important, of course, but it's still a great option to dig into the project :)

Thank you kindly for the contribution!

> I haven't found any test cases for TS wrappers. Does Tact plan to create (or have) one?

That would be nice, but there are no immediate plans (ok, now there are: #1107). We have a [good-to-have plans](https://github.com/tact-lang/tact/issues/255) to make the wrapper generation system a bit more flexible to allow making wrappers for multiple languages, but those are a bit far-fetched now that the team is focusing more efforts on many other areas of the compiler.

@i582 Could you please add an entry to the CHANGELOG.md?

i582:
@novusnota done

novusnota (dismissed):
LGTM, but let's wait for the input from @verytactical, @anton-trunov and others :)

src/bindings/writeTypescript.ts
```diff
@@ -352,7 +352,12 @@ export function writeTypescript(
                                 r.message.text !== null &&
                                 r.message.text !== undefined
                             ) {
-                                receivers.push(`'${r.message.text}'`);
+                                // since we are using single quotes, escape them in the message name
+                                const escaped = r.message.text.replace(
```
verytactical: Could there also be a `\` in `r.message.text`?

Probably `JSON.stringify(r.message.text)` is what we need here.

i582: AFAIK, `JSON.stringify` doesn't escape `'` in string 🤔

novusnota: > Could there also be a \ in r.message.text?

Only a double one, like `\\`, https://github.com/tact-lang/tact/blob/229ce26cf95023ceff3d754d2622501d70b83cb6/src/grammar/grammar.ohm#L340

i582: also:
```
TactParseError: test/main.tact:15:17: Parse error: expected "\""

Line 15, col 17:
  14 | 
> 15 |     receive("get\'") {
                       ^
  16 | 
```

i582: But maybe we can just use:
```js
`string`
```

verytactical: > doesn't escape ' in string

That wouldn't be a problem if it was a `"`-string, would it? :D

> Only a double one, like `\\`

Then it probably shouldn't matter. It's just that serializing JS-compatible string is easily done with `JSON.stringify`, and there isn't really any way it can go wrong (JS-syntax-wise). If the goal is to generate TS in a way that when it runs runtime has the same exact string, `stringify` is the best option.

i582: > That wouldn't be a problem if it was a `"`-string, would it? :D

Yeah, but then it's better to use '`' and escape only it in the string (less problems with any \\ things)

i582: > Then it probably shouldn't matter. It's just that serializing JS-compatible string is easily done with `JSON.stringify`, and there isn't really any way it can go wrong (JS-syntax-wise). If the goal is to generate TS in a way that when it runs runtime has the same exact string, `stringify` is the best option.

Now I understand, this seems like the best solution of all

anton-trunov:
@i582 Hi, thanks for the contribution. The original PR template had some pointers on our PR merging policy and it requires adding tests to make sure this bug won't creep in later after some refactoring, for instance.

i582:
> it requires adding tests to make sure this bug won't creep in later after some refactoring, for instance.

Yeah, I saw that, that's why I asked about tests. However, since there are no tests for wrappers yet (as I understand), what tests could I add, @anton-trunov?

anton-trunov:
@i582 you can use the example from the linked issue, e.g. by including an apostrophe into one of the string receiver's in the end-to-end tests (see `e2e-emulated` folder)

i582:
Thank you, @anton-trunov! I added test for different messages

verytactical (approved):
LGTM

anton-trunov:
@i582 Thanks a bunch!

PR #1111, opened 2024-12-03
feat: clarify error message for bounced types from which accessed a field that does not fit in 224 bytes

### LINKED ISSUE #1021: Improve error reporting for `bounced` messages with missing fields
A typical error message for a large bounced message looks like this:

```
Type bounced<"Withdraw"> does not have a field named "amount"
```

The error message should include the info why some field is missing. This should prevent issues like #1020, #940, #54.

i582:
Previously, the following code would produce a confusing error:

```tact
message Withdraw {
    data: Int as uint128;
    amount: Int as uint128;
}

contract Fund {
    balance: Int as uint256 = 0;

    bounced(msg: bounced<Withdraw>) {
        self.balance += msg.data;
        self.balance += msg.amount; // Type bounced<"Withdraw"> does not have a field named "amount"
    }
}
```

Now, it gives following error:

> Maximum size of the bounced message is 224 bytes, but the "amount" field of type "Withdraw" cannot fit into it due to the size of previous fields or its own size, so it cannot be accessed. Make the type of the fields before this one smaller, or reduce the type of this field so that it fits into 224 bytes

If the type specifies only one field, it will look like this:

> Maximum size of the bounced message is 224 bytes, but the "amount" field of type "Withdraw" cannot fit into it because its too big, so it cannot be accessed. Reduce the type of this field so that it fits into 224 bytes

It's way too big, but since we don't have a more complex UI for errors (like in Rust), a long clear message with all the details is better than a short but not very clear message.

## Issue

Closes #1021.

## Checklist


src/types/resolveExpression.ts
```diff
@@ -450,6 +450,26 @@ function resolveFieldAccess(
             }
         }
 
+        // If it is bounced<T>, check all fields of T
+        if (src.kind === "ref_bounced") {
+            const field = srcT.fields.find((v) => eqNames(v.name, exp.field));
+
+            // If it has field, it means that this field doesn't fit into bounced message
+            if (field !== undefined) {
+                if (srcT.fields.length == 1) {
```
```suggestion
                if (srcT.fields.length === 1) {
```

src/types/resolveExpression.ts
anton-trunov, line 1: I would argue 

```ts
if (src.kind === "ref_bounced") {
        fields = fields.slice(0, srcT.partialFieldCount);
}
```

needs to be changed: let's not remove the fields here and just check that for bounced messages if the field's index is beyond `srcT.partialFieldCount` then we should make the message more informative.

anton-trunov (changes requested):
Looks good, just some minor remarks

src/types/resolveExpression.ts
i582: Makes sense, I thought about it, but decided to make the change as small as possible

i582: Done

src/types/resolveExpression.ts
i582: Done

i582: I'm not sure if we need to include info about indexes, it would make the message more confusing to me

i582:
Kinda strange failure on CI:
```
Tact compilation failed
Error: stdlib/std/base.tact:23:13: Invalid type "Cell" for binary operator ">"
Line 23, col 13:
  22 |         // Lock storage if needed
> 23 |         if (self.storageReserve > 0) { // Optimized in compile-time
                   ^~~~~~~~~~~~~~~~~~~~~~~
  24 |             let ctx: Context = context();
```

anton-trunov (approved):

PR #1223, opened 2024-12-20
Draft: fix: correct serialization for addresses

### LINKED ISSUE #1198: Correct serialization of `Address?`
Currently, optional types are described as nullable ones (e.g. `boolean | null`), which is fine for all types except `Address`.
`@ton/core` handles addresses in such way that `null` means `addr_none`, and it was implemented in Tact in a similar way due to the original developer being the same person who made the library.
We want to rework addresses (#1119) for consistency and predictability, but `null`s prevent us from doing so without refactoring either the ton-core library itself or generated TS wrappers for Tact contracts. Current maintainer of the library suggested the second option.

More precisely, we can make some type wrapper for `Maybe` that will be used instead of the `| null` thing to support `null` as a value itself. This will potentially resolve the problem and allow us to describe `Maybe MsgAddress` type which will look like `Maybe<Address | ExternalAddress | null>`.

What do you think? @anton-trunov @verytactical

verytactical:
## Issue

Blocked by https://github.com/ton-org/ton-core/pull/60

Closes #1198.

## Checklist

### DIFF
--- a/src/bindings/typescript/serializers.ts
+++ b/src/bindings/typescript/serializers.ts
@@ -243,30 +243,38 @@ const boolSerializer: Serializer<{ optional: boolean }> = {
 const addressSerializer: Serializer<{ optional: boolean }> = {
     tsType(v) {
         if (v.optional) {
-            return "Address | null";
+            return "'none' | Address | ExternalAddress | null | undefined";
         } else {
-            return "Address";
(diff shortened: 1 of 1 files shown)


anton-trunov:
Closing this for now because of inactivity

PR #1248, opened 2024-12-24
Parser/printer fuzz tests for expressions

### LINKED ISSUE #642: AST-based randomized testing of Tact parser/pretty-printer
### Summary

We generate random Tact ASTs and test the following property:

- pretty-printed AST when parsed should provide the original AST.

One more property would be nice to test:

- the pretty-printer produces text with the minimum possible parentheses.

### LINKED ISSUE #1469: AST-based randomized testing of parser/pretty-printer for expressions
### Some pointers for testing Tact expressions

Goal: implement automated tests to ensure the correctness of the [Tact](https://docs.tact-lang.org) pretty-printer for expressions using the property-based randomized testing approach, also known as [quickchecking](https://en.wikipedia.org/wiki/QuickCheck).

The idea here is to randomly generate abstract syntax trees (AST) that represent Tact language constructions after parsing.

The fully blown implementation would start with generating abstract Tact AST modules, essentially generating a Tact file with multiple structures/messages, top-level functions, contracts and traits. But this is a pretty large task, so we just would like to do it for a small subset of Tact, namely expressions.

As a property based testing framework we could recommend using [fast-check](https://github.com/dubzzz/fast-check).

The approach here is as follows:

- generate a random AST for expression,
- pretty-print it, i.e. convert it to plain text representation,
- parse it, turning again into an AST,
- compare the originally generated AST and the parsed one: those must be the same, otherwise there is a bug (most likely in the pretty-printer),
- if the ASTs are the same, repeat the process.

More formally, for every generated Tact expression AST ast you need to check the following property:

```ts
parseExpression(prettyPrint(ast)) = ast
```

Note that you cannot use the builtin TS equality `===`, because Tact ASTs contain the location information `loc` and AST identifiers `id`, so you will most likely need to implement your own equality comparison operation which would ignore that information.

Note that often times the generated counter-example is too large for a human to understand where the bug is, so if you implement a shrinker which tries to minimize the counter-example, that gives bonus points, but it's not strictly necessary for this test assignment.

#### Useful links

- The Tact AST for expressions: https://github.com/tact-lang/tact/blob/fbabf31673e439d66d4133f5e3588558769964f6/src/grammar/ast.ts#L380
- The pretty-printer entry point: https://github.com/tact-lang/tact/blob/fbabf31673e439d66d4133f5e3588558769964f6/src/prettyPrinter.ts#L892
- The Tact expression parser: https://github.com/tact-lang/tact/blob/fbabf31673e439d66d4133f5e3588558769964f6/src/grammar/grammar.ts#L1515
- The description of the Tact repository: https://github.com/tact-lang/tact/blob/main/CONTRIBUTING.md

xpyctumo:
## Issue

Closes #1469 

## Checklist

## Plan
- [x] **AstExpression**
  - [x] AstOpBinary
  - [x] AstOpUnary
  - [x] AstConditional
  - [x] AstMethodCall
  - [x] AstFieldAccess
  - [x] AstStaticCall
  - [x] AstStructInstance
  - [x] AstNumber
  - [x] AstBoolean
  - [x] AstId
  - [x] AstInitOf
  - [x] AstString
- [x] **AstLiteral**
  - [x] AstNumber
  - [x] AstBoolean
  - [x] AstNull
  - [ ] AstSimplifiedString
  - [ ] AstAddress
  - [ ] AstCell
  - [ ] AstSlice
  - [ ] AstCommentValue
  - [ ] AstStructValue


src/test/prettyPrint/expressions.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+    const maxShrinks = 15;
+
+    const generateAstNumber = () =>
+        fc.record({
+            kind: fc.constant("number"),
+            base: fc.constantFrom(2, 8, 10, 16),
+            value: fc.bigInt().filter((n) => n > 0),
+            id: fc.constant(0),
+            loc: fc.constant(dummySrcInfo),
+        }) as fc.Arbitrary<AstNumber>;
```
reverofevil, line 25: Please don't use `as`.

anton-trunov:
when I see something like this, I immediately think of adding shrinking (fast-check should support it)
```
(+0b100100110001001100110111100100000101101111000011101011010101101101011110010000001000101100100000011000111001001110101001111011111100001101111110011000101010010001100000100100011010000100110011001111001000111001011101000100011011100111001110 ? 0xf39c6286c40970dcb2a321a1c21e4549de1a2d60bcc7ff84622639b35024a7f || 0o3235042541620525012122214562100130425412434124566713521630100545344175751456215504064 - -0o7777777777777777777777777777777777777777777777777777777777777777777777777777777777743 : 8 - (16227299991510006730850008471880976967479204885141379564359356953639308462371 == 0o5735742350164242162734544666035326212173453606274734532135112142107426251434140545063)) ? 0b10111100010010010001011110101011101010001000111101110001100011100001101000001011001111111101110100100011001000001001001111011110100011001100011010100111110010110110010011110110010001101000100111000101011110110110110010100100010011010011 + 0b11001101101101011111001101010011101111110011111110101001101100111000110011110110011010111001101011001111001111110001100110000010001001101011100010111101000010100000010001011100101110010100100101100101101110110110111100111100110111000101011111011001100010 + 26 + (0x157e944ffa07c17e8b222faf8abfab932cd54c213784eca386a32755e232289a ? 0o2667602156054031504566507661155302000265110317032415711565650770337656725250250767327 << 0o5520154115374334170135502317743632312724505264461136361671447006177506075556352660004 : 0o6624600205764760715715350617562272506065501073701250447411343437104466044353352053407) : 0b1
```

even shrinking integer literals would already help a lot here

xpyctumo:
> when I see something like this, I immediately think of adding shrinking (fast-check should support it)
> 
> ```
> (+0b100100110001001100110111100100000101101111000011101011010101101101011110010000001000101100100000011000111001001110101001111011111100001101111110011000101010010001100000100100011010000100110011001111001000111001011101000100011011100111001110 ? 0xf39c6286c40970dcb2a321a1c21e4549de1a2d60bcc7ff84622639b35024a7f || 0o3235042541620525012122214562100130425412434124566713521630100545344175751456215504064 - -0o7777777777777777777777777777777777777777777777777777777777777777777777777777777777743 : 8 - (16227299991510006730850008471880976967479204885141379564359356953639308462371 == 0o5735742350164242162734544666035326212173453606274734532135112142107426251434140545063)) ? 0b10111100010010010001011110101011101010001000111101110001100011100001101000001011001111111101110100100011001000001001001111011110100011001100011010100111110010110110010011110110010001101000100111000101011110110110110010100100010011010011 + 0b11001101101101011111001101010011101111110011111110101001101100111000110011110110011010111001101011001111001111110001100110000010001001101011100010111101000010100000010001011100101110010100100101100101101110110110111100111100110111000101011111011001100010 + 26 + (0x157e944ffa07c17e8b222faf8abfab932cd54c213784eca386a32755e232289a ? 0o2667602156054031504566507661155302000265110317032415711565650770337656725250250767327 << 0o5520154115374334170135502317743632312724505264461136361671447006177506075556352660004 : 0o6624600205764760715715350617562272506065501073701250447411343437104466044353352053407) : 0b1
> ```
> 
> even shrinking integer literals would already help a lot here

I'm thinking about that right now. It might become more user- friendly values
```js
value: fc.bigInt({ min: -1_000_000_000n, max: 1_000_000_000n })
```

src/test/prettyPrint/expressions.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+                "||",
+                "%",
+                "<<",
+                ">>",
+                "&",
+                "|",
+                "^",
+            ),
+            left: expression,
+            right: expression,
```
anton-trunov: These two need to be different expressions in general

src/test/prettyPrint/expressions.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+    // Max depth of the expression tree
+    const maxShrinks = 15;
+
+    const generateAstNumber = () =>
+        fc.record({
+            kind: fc.constant("number"),
+            base: fc.constantFrom(2, 8, 10, 16),
+            value: fc.bigInt().filter((n) => n > 0),
+            id: fc.constant(0),
+            loc: fc.constant(dummySrcInfo),
```
anton-trunov: I'd add a wrapper to automatically add these `id` and `loc` to the generated AST nodes

src/test/prettyPrint/expressions.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+        }) as fc.Arbitrary<AstOpBinary>;
+
+    const generateAstConditional = (expression: fc.Arbitrary<AstExpression>) =>
+        fc.record({
+            kind: fc.constant("conditional"),
+            condition: expression,
+            thenBranch: expression,
+            elseBranch: expression,
```
anton-trunov: These need to be 3 different expressions

src/test/prettyPrint/expressions.spec.ts
xpyctumo: All generators wrapped by `dummyAstNode` now

src/test/prettyPrint/expressions.spec.ts
xpyctumo: Fixed, every expression is new-generated

src/test/prettyPrint/expressions.spec.ts
xpyctumo: Also fixed

anton-trunov:
@xpyctumo I took the liberty to merge the new parser into your branch and pushed a few fixes for the expressions generator too. Looks like the new parser fixed the issue with duplicated unary operators!

anton-trunov:
Btw, these fuzzing-like tests should go into their own folder and only gets executed on a special request, separately from `yarn tests`.

src/test/prettyPrint/expressions.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    randomAstBoolean,
+    randomAstNumber,
+    randomAstString,
+} from "../utils/expression/randomAst";
+
+describe("Pretty Print Expressions", () => {
+    const maxDepth = 3;
+    const expression = () => randomAstExpression(maxDepth);
+
+    const cases: [string, fc.Arbitrary<AstExpression>][] = [
```
anton-trunov: looks like you don't need `cases` here: you just need to generate a lot of random expressions, using `randomAstExpression`: it will handle all the cases internally

xpyctumo: Do you want to generate one thing?

I left it out specifically so that you can selectively do tests on each subset

anton-trunov: yep, expressions are a fine level of granularity, given that you also need to generate functions with statements, contracts etc

xpyctumo: I removed the list of cases, but increased the number of runs

src/test/prettyPrint/expressions.spec.ts
anton-trunov, line 1: let's move and rename this file: `src/ast/fuzz.spec.ts`

src/test/utils/expression/randomAst.ts
anton-trunov, line 1: let's move and rename this file: `src/ast/random.infra.ts` and then exclude all files like `*.infra.ts` from bundling into the `dist` folder

src/test/utils/expression/randomAst.ts
```diff
@@ -0,0 +1,260 @@
...
+    AstMethodCall,
+    AstNull,
+    AstNumber,
+    AstOpBinary,
+    AstOpUnary,
+    AstStaticCall,
+    AstString,
+    AstStructFieldInitializer,
+    AstStructInstance,
+} from "../../../ast/ast";
```
anton-trunov: Long imports, see #1407

anton-trunov (changes requested):

src/test/prettyPrint/expressions.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+import { randomAstExpression } from "../utils/expression/randomAst";
+
+describe("Pretty Print Expressions", () => {
+    const maxDepth = 3;
+
+    it(`should parse Ast`, () => {
+        fc.assert(
+            fc.property(randomAstExpression(maxDepth), (generatedAst) => {
+                const prettyBefore = prettyPrint(generatedAst);
+                const parser = getParser(getAstFactory(), "new");
```
anton-trunov: move `const parser = getParser(getAstFactory(), "new");` out of the loop, here we don't care about `id` sharing

src/test/utils/expression/randomAst.ts
```diff
@@ -0,0 +1,260 @@
...
+    );
+}
+function randomAstOpBinary(
+    leftExpression: fc.Arbitrary<AstExpression>,
+    rightExpression: fc.Arbitrary<AstExpression>,
+): fc.Arbitrary<AstOpBinary> {
+    return dummyAstNode(
+        fc.record({
+            kind: fc.constant("op_binary"),
+            op: fc.constantFrom(
```
anton-trunov: Reuse the corresponding array defined in AST when #1465 is solved

src/test/utils/expression/randomAst.ts
```diff
@@ -0,0 +1,260 @@
...
+    );
+}
+
+function randomAstOpUnary(
+    operand: fc.Arbitrary<AstExpression>,
+): fc.Arbitrary<AstOpUnary> {
+    return dummyAstNode(
+        fc.record({
+            kind: fc.constant("op_unary"),
+            op: fc.constantFrom("+", "-", "!", "!!", "~"),
```
anton-trunov: Reuse the corresponding array defined in AST when https://github.com/tact-lang/tact/issues/1465 is solved

src/test/utils/expression/randomAst.ts
```diff
@@ -0,0 +1,260 @@
...
+function randomAstNumber(): fc.Arbitrary<AstNumber> {
+    const values = [
+        ...Array.from({ length: 10 }, (_, i) => [0n, BigInt(i)]).flat(),
+        ...Array.from({ length: 256 }, (_, i) => 1n ** BigInt(i)),
+    ];
+
+    return dummyAstNode(
+        fc.record({
+            kind: fc.constant("number"),
+            base: fc.constantFrom(2, 8, 10, 16),
```
anton-trunov: Reuse the corresponding array defined in AST when https://github.com/tact-lang/tact/issues/1465 is solved

anton-trunov:
Let's move that code around, add the rest of literals and merge this PR so we can move on to generating statements

src/test/prettyPrint/expressions.spec.ts
xpyctumo: Ready 🚀

src/test/utils/expression/randomAst.ts
xpyctumo: Good point, fixed

src/test/utils/expression/randomAst.ts
xpyctumo: Moved, however I may have misunderstood about the build.

I added in `package.json`
```
  “files": [
    ...
    “!**/*/*.infra.ts”
  ],
```
Right?

src/test/prettyPrint/expressions.spec.ts
xpyctumo: 👍

src/test/utils/expression/randomAst.ts
anton-trunov: "!**/*.infra.ts",

src/ast/random.infra.ts
```diff
@@ -0,0 +1,244 @@
...
+            elseBranch: elseBranchExpression,
+        }),
+    );
+}
+
+function randomAstId(): fc.Arbitrary<A.AstId> {
+    return dummyAstNode(
+        fc.record({
+            kind: fc.constant("id"),
+            text: fc.stringMatching(/^[A-Za-z_][A-Za-z0-9_]*$/),
```
anton-trunov: this should exclude the keywords

src/ast/random.infra.ts
```diff
@@ -0,0 +1,244 @@
...
+            // Rules for text value are in src/grammar/grammar.ohm
+        }),
+    );
+}
+
+function randomAstCapitalizedId(): fc.Arbitrary<A.AstId> {
+    return dummyAstNode(
+        fc.record({
+            kind: fc.constant("id"),
+            text: fc.stringMatching(/^[A-Z]+$/),
```
anton-trunov: this should generate capitalized identifiers, not all caps, e.g. `Foo` should also be generated but now this is not the case

src/ast/random.infra.ts
```diff
@@ -0,0 +1,244 @@
...
+    maxDepth: number,
+): fc.Arbitrary<A.AstExpression> {
+    // No weighted items
+    const baseExpressions = [
+        randomAstNumber(),
+        randomAstBoolean(),
+        randomAstId(),
+        randomAstNull(),
+        randomAstString(),
+    ];
```
anton-trunov: These should be literals, so you need a separate generator for just the literals

anton-trunov (changes requested):
Looks like you need to fix imports after moving files around

src/test/utils/expression/randomAst.ts
xpyctumo: > "!**/*.infra.ts",

That's exactly what I commited. My comment above contains an error in the path

src/ast/random.infra.ts
xpyctumo: I see 

```js
export type AstExpression =
    | AstOpBinary
    | AstOpUnary
    | AstConditional
    | AstMethodCall
    | AstFieldAccess
    | AstStaticCall
    | AstStructInstance
    | AstId
    | AstInitOf
    | AstString
    | AstLiteral;   // <----------

export type AstLiteral =
    | AstNumber
    | AstBoolean
    | AstNull
    | AstSimplifiedString
    | AstAddress
    | AstCell
    | AstSlice
    | AstCommentValue
    | AstStructValue;
```
You suggest make  `baseExpressions` which will contains only `AstLiteral`?

src/ast/random.infra.ts
xpyctumo: I have added a list of reserve words that will be excluded from generation

src/ast/random.infra.ts
xpyctumo: Fixed

src/ast/random.infra.ts
```diff
@@ -0,0 +1,287 @@
...
+        }),
+    );
+}
+
+function randomAstId(): fc.Arbitrary<A.AstId> {
+    return dummyAstNode(
+        fc.record({
+            kind: fc.constant("id"),
+            text: fc
+                .stringMatching(/^[A-Za-z_][A-Za-z0-9_]*$/)
```
anton-trunov, line 161: https://docs.tact-lang.org/book/expressions/#identifiers

All identifiers **starting** with `__gen` and `__tact` are not allowed and are instead reserved for internal use by the compiler.

It's unlikely that we generate identifiers like that, but still needs to be included, since we don't need flaky tests

src/ast/random.infra.ts
```diff
@@ -0,0 +1,287 @@
...
+                .filter((i) => !reservedWords.includes(i)),
+        }),
+    );
+}
+
+function randomAstCapitalizedId(): fc.Arbitrary<A.AstId> {
+    return dummyAstNode(
+        fc.record({
+            kind: fc.constant("id"),
+            text: fc.stringMatching(/^[A-Z][a-z]*$/),
```
anton-trunov:
```suggestion
            text: fc.stringMatching(/^[A-Z][A-Za-z0-9_]*$/),
```

src/ast/random.infra.ts
```diff
@@ -0,0 +1,287 @@
...
+}
+
+function randomAstInitOf(
+    expression: fc.Arbitrary<A.AstExpression>,
+): fc.Arbitrary<A.AstInitOf> {
+    return dummyAstNode(
+        fc.record({
+            kind: fc.constant("init_of"),
+            contract: randomAstId(),
+            args: fc.array(expression, { maxLength: 1 }),
```
anton-trunov: why `maxLength: 1` here? it needs to be a random list of expressions

src/ast/random.infra.ts
anton-trunov: Your generators should just follow the type structure.

You define `randomAstLiteral() : fc.Arbitrary<A.AstLiteral>` and then `randomAstExpression` just calls it for the base case

anton-trunov (changes requested):

src/ast/random.infra.ts
xpyctumo: Thanks for the reminder, I'll add them to `reservedWords`

anton-trunov: `__gen` and `__tact` are not full words, those are prefixes, so anything like `__tact1` needs to be excluded

xpyctumo: Yeah, i've used `.startsWith()` to detect these prefixes

src/ast/random.infra.ts
xpyctumo: I have added the remaining generators that are described in `AstLiteral`.

xpyctumo:
```js
function randomAddress(): fc.Arbitrary<Address> {
    return fc.constant(
        address("EQCD39VS5jcptHL8vMjEXrzGaRcCVYto7HUn4bpAOg8xqB2N"), // TODO: use random address
    );
}

function randomAstAddress(): fc.Arbitrary<A.AstAddress> {
    return dummyAstNode(
        fc.record({
            kind: fc.constant("address"),
            value: randomAddress(),
        }),
    );
}
```

Generating random address via this we get AST like this
```json
{
  "kind": "address",
  "value": {
    "workChain": 0,
    "hash": {
      "type": "Buffer",
      "data": [
        131, 223, 213,  82, 230,  55,  41, 180, 114, 252, 188, 200, 196,  94,
        188, 198, 105,  23,   2,  85, 139, 104, 236, 117,  39, 225, 186,  64,
         58,  15,  49, 168
      ]
    }
  },
  "id": 0,
  "loc": {}
}
```
Which pretty printer converts into 
```
addr("0:83dfd552e63729b472fcbcc8c45ebcc6691702558b68ec7527e1ba403a0f31a8")
```
The parser then casts this to the following AST

```json
{
  "id": 3,
  "kind": "static_call",
  "function": {"id": 1, "kind": "id", "text": "addr", "loc": {}},
  "args": [
    {
      "id": 2,
      "kind": "string",
      "value": "0:83dfd552e63729b472fcbcc8c45ebcc6691702558b68ec7527e1ba403a0f31a8",
      "loc": {}
    }
  ],
  "loc": {}
}
```

Which is not the expected result.

src/ast/random.infra.ts
xpyctumo: I used this when I was looking for where there might be a leak entering recursion.
Removed cause this is no need more

jest.setup.js
```diff
@@ -1,8 +1,51 @@
...
+            }
+        }
+        return newObj;
     }
-};
+    return obj;
+}
+
+fc.configureGlobal({
+    reporter: (log) => {
```
verytactical, line 30: As far as I understand this changes jest output for all tests, not only for property-based tests.

Could you please add some rationale on why this code is required and what it does? I'm not really used to this kind of low-level jest configuration.

jest.setup.js
```diff
@@ -1,8 +1,51 @@
...
+                    value: (val) =>
+                        typeof val === "bigint" ? val.toString() : val,
+                },
+            });
+
+            const errorMessage = `
+      Property failed after ${log.numRuns} tests
+      Seed: ${log.seed}
+      Path: ${log.counterexamplePath}
+      Counterexample: ${JSON.stringify(sanitizedCounterexample, null, 0)}
```
verytactical, line 44: I don't know if "no spaces" is the best way to display a counterexample :)

jest.setup.js
```diff
@@ -1,8 +1,51 @@
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const coverage = require("@tact-lang/coverage");
+const fc = require("fast-check");
 
-module.exports = async () => {
-    if (process.env.COVERAGE === "true") {
-        coverage.beginCoverage();
+function sanitizeObject(
```
verytactical, line 3: This code is a tad bit too complex for a `.js` file. Consider changing file extension to `.ts`. I tested it recently, and it should work just fine.

verytactical: Also there might be an `id` or `loc` field on some AST object that is not an AST node, so this code isn't really safe to use.

Please add an issue [here](https://github.com/tact-lang/tact/issues/1401) if this code is really needed, so that we can make a proper type-directed sanitizer.

src/ast/fuzz.spec.ts
```diff
@@ -0,0 +1,21 @@
+import fc from "fast-check";
+import { eqExpressions, getAstFactory } from "./ast";
+import { prettyPrint } from "../prettyPrinter";
```
verytactical: Pretty-printer is now in a different place. Apparently git rebase went wrong here.

jest.setup.js
xpyctumo: It may big if you will print big expressions. 

So, i removed spaces and you can copy-paste this json into any json viewer to see

jest.setup.js
xpyctumo: You're right, it affects the reporter of all tests.

I'm using this to have a low-level effect on the reporter and control the incoming log, because right now it doesn't know how to output AST as JSON

src/ast/fuzz.spec.ts
xpyctumo: Fixed

jest.setup.js
verytactical: Doesn't sound like good developer experience to me, to be honest :)

Maybe we could solve it this way:
- If it's short, show it directly (with proper JSON formatting).
- If it's long, `fs.writeFile` JSON with formatting to `/output/<8 random hex digits>.json`, and print full path to that file.

jest.setup.js
verytactical: Is there a way to avoid changing output for all the other tests? It's very likely that `counterexamplePath` would not be present, for example.

jest.setup.js
novusnota: Perhaps, the second approach with the usage of [`fs.mkdtempSync()`](https://nodejs.org/docs/latest-v22.x/api/fs.html#fsmkdtempsyncprefix-options) (or async) is best — we'll get `/tmp/<randomized>` (or similar for non-Linux OSes) and then we can put `.json` there. Then, user can do whatever they want with it: view it with, say `jq`, or just copy-paste to a buffer like `cat /tmp/<randomized>/that.json | xsel -bi` (or `pbcopy` for macOS)

i582:
One of the example I get:
```
(+"").a(a{})
```
is not allowed because we need a capital initial letter for the types

```
Property failed after 5 tests
Seed: 1
Path: 4:0:0:0:0:1:0:0:0:1:1:1
```

src/ast/random.infra.ts
```diff
@@ -0,0 +1,400 @@
...
+    );
+}
+
+function randomAstStructValue(
+    subLiteral: fc.Arbitrary<A.AstLiteral>,
+): fc.Arbitrary<A.AstStructValue> {
+    return dummyAstNode(
+        fc.record({
+            kind: fc.constant("struct_value"),
+            type: randomAstId(),
```
i582: This should be `randomAstCapitalizedId()`

i582:
I think we definitely need diff when the trees differ, like here:

Screenshot 2025-01-28 at 21 49 31

src/ast/random.infra.ts
```diff
@@ -0,0 +1,400 @@
...
+            value: randomSlice(),
+        }),
+    );
+}
+
+function randomAstCommentValue(): fc.Arbitrary<A.AstCommentValue> {
+    return dummyAstNode(
+        fc.record({
+            kind: fc.constant("comment_value"),
+            value: fc.string(),
```
i582: Why do we generate `comment_value` as a string?

i582:
I think supporting Slice might be too complicated at the moment since we use `Slice` type in AST node. We can skip it, fix all remaining issues and finally merge it

i582:
One more example with wrong output:

```
0b1._("".a(), S_{oaFb3yD_: null})

0b1._("".a(), S_{oaFb3yD_: null})
```

Screenshot 2025-01-28 at 22 03 57

i582:
It's kinda bad code, but it can be a good starting point for nice and clear error output:

```typescript
import {diffJson} from "diff";

fc.property(randomAstExpression(maxDepth), (generatedAst) => {
    const prettyBefore = prettyPrint(generatedAst);

    const parsedAst = parser.parseExpression(prettyBefore);
    const prettyAfter = prettyPrint(parsedAst);

    expect(prettyBefore).toBe(prettyAfter)

    const actual = eqExpressions(generatedAst, parsedAst);
    if (!actual) {
        const replacer = (key: string, value: unknown): unknown => {
            if (key === "id") return undefined;
            if (typeof value === "bigint") return value.toString();
            return value;
        };

        const left = JSON.stringify(sortObjectKeys(generatedAst), replacer, 4);
        const right = JSON.stringify(sortObjectKeys(parsedAst), replacer, 4);

        const differences = diffJson(left, right);

        const ConsoleColors = {
            added: "\x1b[32m",
            removed: "\x1b[31m",
            reset: "\x1b[0m",
        };

        differences.forEach(part => {
            const color = part.added
                ? ConsoleColors.added
                : part.removed
                    ? ConsoleColors.removed
                    : ConsoleColors.reset;

            process.stdout.write(color + part.value + ConsoleColors.reset);
        });

        process.stdout.write(`\n\n${prettyBefore}\n\n`);
        process.stdout.write(`${prettyAfter}\n\n`);
    }

    expect(actual).toBe(true);
}),
```

src/ast/random.infra.ts
xpyctumo: i've been following type 

```js
export type AstCommentValue = {
    kind: "comment_value";
    value: string; 
    id: number;
    loc: SrcInfo;
};
```

jest.setup.js
xpyctumo: #1630 here we go

src/ast/random.infra.ts
xpyctumo: Updated, thanks!

xpyctumo:
> I think supporting Slice might be too complicated at the moment since we use `Slice` type in AST node. We can skip it, fix all remaining issues and finally merge it

I decided to ignore the `Address`, `Cell`, and `Slice` generators until we solve their problem.

xpyctumo:
Following generators are **must be** implemented. I leave them here to debug.

```js

function randomAddress(): fc.Arbitrary<Address> {
    return fc.constant(
        address("EQCD39VS5jcptHL8vMjEXrzGaRcCVYto7HUn4bpAOg8xqB2N"), // TODO: use random address
    );
}

function randomCell(): fc.Arbitrary<Cell> {
    return fc.constant(beginCell().endCell()); // TODO: use random random here
}

function randomAstAddress(): fc.Arbitrary<A.AstAddress> {
    return dummyAstNode(
        fc.record({
            kind: fc.constant("address"),
            value: randomAddress(),
        }),
    );
}

function randomAstCell(): fc.Arbitrary<A.AstCell> {
    return dummyAstNode(
        fc.record({
            kind: fc.constant("cell"),
            value: randomCell(),
        }),
    );
}

function randomSlice(): fc.Arbitrary<Slice> {
    return fc.constant(beginCell().endCell().beginParse());
}

function randomAstSlice(): fc.Arbitrary<A.AstSlice> {
    return dummyAstNode(
        fc.record({
            kind: fc.constant("slice"),
            value: randomSlice(),
        }),
    );
}
```

anton-trunov (changes requested):
We need
- to introduce a command to run a sampler: `yarn random-ast 42` would produce 42 random expressions and pretty print those in the terminal -- this is needed to understand what kind of expressions we actually produce 
- to document this type of tests in CONTRIBUTING.md

src/ast/random.infra.ts
xpyctumo: That's removed. Since we need the test of the optimizer separately
Also this affects `AstSimplifiedString`

jest.setup.js
xpyctumo: I suggest for the time being to unfold AST for now while his length < 1000 characters

I'm not sure we need to write the broken AST samples to disk

jest.setup.js
xpyctumo: I've had other tests fail, however, I haven't found any problems

jest.setup.js
xpyctumo: > This code is a tad bit too complex for a .js file. Consider changing file extension to .ts. I tested it recently, and it should work just fine.

```
/Users/admin/projects/tact/jest.setup.ts
  0:0  error  Parsing error: "parserOptions.project" has been provided for @typescript-eslint/parser.
The file was not found in any of the provided project(s): jest.setup.ts
```

ESLint does not allow i think.

jest.globalSetup.js
```diff
@@ -0,0 +1,8 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const coverage = require("@tact-lang/coverage");
+
+module.exports = async () => {
+    if (process.env.COVERAGE === "true") {
+        coverage.beginCoverage();
```
verytactical, line 6: Why do we need this?

xpyctumo: I moved the file, those aren't my lines

verytactical (dismissed):

PR #1250, opened 2024-12-24
docs: more styleguide

verytactical:
PR to keep other styleguide changes as long as I keep remembering them


novusnota (commented):
It would be cool if you could encode some of the rules from the list as our ESLint rules

anton-trunov (approved):

PR #1271, opened 2024-12-29
feat: `fromCell` method for maps

### LINKED ISSUE #1229: `fromCell` is missing for maps
Maps are essentially cells under-the-hood. There is the `asCell` no-operation type converter which allows, for instance, to send maps in messages, but there is no corresponding `fromCell` type converter that would allow to use the transmitted map as a map on the other end.

Gusarich:
## Issue

Closes #1229.

## Checklist

### DIFF
--- a/docs/src/content/docs/book/maps.mdx
+++ b/docs/src/content/docs/book/maps.mdx
@@ -360,6 +360,21 @@ contract Example {
 }
 ```
 
+### Convert from a `Cell`, `.fromCell()` {#fromcell}
+
+<Badge text="Available since Tact 1.6 (not released yet)" variant="tip" size="medium"/><p/>
+
+To convert a [`Cell{:tact}`][cell] type back to a map, use the `.fromCell(){:tact}` [method](/book/functions#extension-function).
+
+```tact
+// Suppose we have a Cell
+let cell: Cell = ...;
+
+// And we want to initialize a map variable from it
+let fizz: map<Int, Int> = emptyMap();
+fizz.fromCell(cell);
+```
+
 ### Traverse over entries {#traverse}
 
 To iterate over map entries there is a [`foreach{:tact}`](/book/statements#foreach-loop) loop statement:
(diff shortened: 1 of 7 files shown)


docs/src/content/docs/book/maps.mdx
```diff
@@ -360,6 +360,21 @@ contract Example {
...
+
+To convert a [`Cell{:tact}`][cell] type back to a map, use the `.fromCell(){:tact}` [method](/book/functions#extension-function).
+
+```tact
+// Suppose we have a Cell
+let cell: Cell = ...;
+
+// And we want to initialize a map variable from it
+let fizz: map<Int, Int> = emptyMap();
+fizz.fromCell(cell);
```
anton-trunov, line 375: hmmm, this should be consistent with what we have for structs:

```tact
struct Fizz { foo: Int }
let fizzCell = ...
let parsedFizz: Fizz = Fizz.fromCell(fizzCell);
```

So I would expect something like:

```tact
let fizz = map<Int, Int>.fromCell(cell);
```

This makes parsing harder but it seems it's too late to do it differently

anton-trunov (changes requested):
1. Let's change the syntax here to make it consistent with `fromCell` for structs/messages.

2. We need negative tests showing what happens when the cell does not represent a valid map. We need to validate the cell and it seems to me we could use TVM's `LDDICTS` instruction here and check the remainder is an empty slice. Or is there a better approach? What do you think?

Gusarich:
> 2\. We need negative tests showing what happens when the cell does not represent a valid map. We need to validate the cell and it seems to me we could use TVM's `LDDICTS` instruction here and check the remainder is an empty slice. Or is there a better approach? What do you think?

@anton-trunov `LDDICTS` doesn't check anything, and there are no TVM instructions for checking the whole dictionary for "correctness" actually. Implementation would have to iterate over the whole tree, which is very gas-consuming.

anton-trunov:
@Gusarich The TVM docs say this for `LDDICTS`: 

> Loads (parses) a (Slice-represented) dictionary s' from Slice s, and returns the remainder of s as s''

So, if a cell is map, then the remainder should be an empty slice, if I understand this correctly.

Gusarich:
> So, if a cell is map, then the remainder should be an empty slice, if I understand this correctly.

@anton-trunov `LDDICTS` doesn't even check the header of the map, it just checks the one single bit indicating whether it's empty or not when it's represented as `Maybe X` (similarly to what `.load_dict()` does.

https://github.com/ton-blockchain/ton/blob/ea0dc161639ef2640876d6de06f7224ac5873847/crypto/vm/dictops.cpp#L121-L146

anton-trunov:
Let's close this for now, we need to brainstorm how this feature should look like in the end

PR #1277, opened 2025-01-01
refactor: `Value` type

### LINKED ISSUE #1190: Refactor `Value` type
`Value` type needs to be refactored so that it is part of `AstExpression` in preparation for partial evaluation, code optimization (specifically to avoid multiple transformations between `Value -> AstExpression -> Value .....`), and inclusion of other values -cells, slices, addresses, simplified strings- into expressions. Additionally, it is currently very hard to work with `Value` in conditionals, specially the handling of `StructValue`.

The proposal is to include `AstLiteral` into `AstExpression`:

```
AstExpression = .... | AstLiteral

AstLiteral = AstNumber | AstStructValue | AstCommentValue | AstSimplifiedString | AstCell ......
```

jeshecdom:
## Issue

Closes #1190.

## Checklist


src/grammar/ast.ts
```diff
@@ -381,7 +383,8 @@ export type AstExpression =
     | AstExpressionPrimary
     | AstOpBinary
     | AstOpUnary
-    | AstConditional;
+    | AstConditional
+    | AstLiteral; // Should I place it here or in AstExpressionPrimary?
```
verytactical: should remove this comment

verytactical: or rather replace with a link to https://github.com/tact-lang/tact/issues/1290

src/grammar/ast.ts
```diff
@@ -876,51 +970,61 @@ function eqParameters(
...
+        ast,
+        () => true,
+        () => false,
+    );
+}
+
+export function checkLiteral<T>(
+    ast: AstExpression,
+    t: (node: AstLiteral) => T,
+    f: (node: Exclude<AstExpression, AstLiteral>) => T,
```
anton-trunov, line 1095: Let's document this either locally or as a recommended pattern to use across the codebase

src/grammar/ast.ts
```diff
@@ -929,6 +1033,9 @@ export function isValue(ast: AstExpression): boolean {
         case "conditional":
         case "field_access":
         case "static_call":
-            return false;
+            return f(ast);
+
+        default:
+            throwInternalCompilerError("Unrecognized expression kind");
```
anton-trunov, line 1122: Is this reachable code?

src/grammar/ast.ts
```diff
@@ -876,51 +970,61 @@ function eqParameters(
...
-export function isValue(ast: AstExpression): boolean {
+export function isLiteral(ast: AstExpression): ast is AstLiteral {
+    return checkLiteral(
+        ast,
+        () => true,
+        () => false,
+    );
+}
+
+export function checkLiteral<T>(
```
verytactical, line 1092: we should add a comment about why this is even needed

something in the lines of

```
Without this intermediate function guard type in `isLiteral` is unsafe to use. Usually guard types must not be used in production code, but using `checkLiteral` instead would significantly reduce readability.
```

src/optimizer/util.ts
```diff
@@ -92,10 +49,126 @@ export const getAstUtil = ({ createNode }: FactoryAst) => {
         return result as AstExpression;
     }
 
+    function makeNumberLiteral(n: bigint, loc: SrcInfo): AstNumber {
```
anton-trunov, line 52: Should these `makeX` functions reside here or in `ast.ts`?

src/prettyPrinter.ts
```diff
@@ -182,6 +186,16 @@ export const ppAstBoolean = ({ value }: A.AstBoolean) => value.toString();
 export const ppAstId = ({ text }: A.AstId) => text;
 export const ppAstNull = (_expr: A.AstNull) => "null";
 export const ppAstString = ({ value }: A.AstString) => `"${value}"`;
+export const ppAstCommentValue = ({ value }: A.AstCommentValue) =>
+    JSON.stringify(value);
+export const ppAstSimplifiedString = ({ value }: A.AstSimplifiedString) =>
+    JSON.stringify(value);
```
verytactical, line 192: I think we should document the difference between AstString and AstSimplifiedString somewhere in ast.ts. Otherwise this code seems quite unexpected.

src/types/resolveExpression.ts
```diff
@@ -765,6 +765,16 @@ export function resolveExpression(
...
+        case "address":
+        case "cell":
+        case "slice":
+        case "simplified_string":
+        case "comment_value":
+        case "struct_value":
+            throwInternalCompilerError(
+                `Expression kind ${exp.kind} should not happen here`,
+            );
+            break;
```
anton-trunov: what if we wanted to typecheck an optimized AST? any optimization needs to preserve the type of the original expression, so this fact can be used in tests

src/types/types.ts
```diff
@@ -63,47 +65,32 @@ export type TypeRef =
...
+            return "null";
+        case "struct_value": {
+            const assocList = val.args.map(
+                (field) =>
+                    `${idText(field.field)}: ${showValue(field.initializer)}`,
+            );
+            return `{${assocList.join(",")}}`;
+        }
+        default:
+            throwInternalCompilerError("Invalid value");
```
anton-trunov, line 93: is this reachable code?

anton-trunov (commented):
I only reviewed the PR superficially, feel free to argue against the suggestions

src/grammar/ast.ts
jeshecdom: It is unreachable. But it is recommended to always include a default with an internal compiler error just in case new kinds get added in the future and someone forgets to add the case in the switch. In this particular case, the default case is redundant since the compiler would complain if someone forgets to add the case because the return type of the function does not include "undefined". However, if the function is of void return type or it includes "undefined", the compiler would remain silent. So, it does not hurt to always include the default case, but I am open to suggestions.

src/optimizer/util.ts
jeshecdom: Mmmm... not quite sure where they should be. Probably they could be moved to the AstFactory methods. However, I recall @verytactical mentioned to me that these functions will probably disappear in the future, because they are kind of already included somewhere, but those already present functions need some adapting to be used in the context where I am using the `makeLiteral` functions, and I am not sure how much effort would be needed to carry out the adaptation. What do you think @verytactical ?

src/types/resolveExpression.ts
jeshecdom: Agreed. Added code for those cases now.

anton-trunov (approved):

Issue #1200, opened 2024-12-17
Docs: Addresses explanation


Shvandre:
Currently, address types in TON (like addr_std, addr_var) are only described in block.tlb, which may not be clear for ordinary users.

Problem:

- Ordinary users may not understand the distinction between addr_std and addr_var.
- The concept of addr_ext is unclear and undocumented.
- There is no explanation of Anycast, which also might be useful.

novusnota:
Related to #925 and, partially, #1110.
Depends on #1198 and #1119.

> There is no explanation of Anycast, which also might be useful.

Btw, Anycast is mostly ignored, isn't it?

Shvandre:
> Btw, Anycast is mostly ignored, isn't it?

Anycast is ignored in 100% cases for now. (And may never be implemented). But actually for now only way to understand what is anycast is to read whitepaper

Shvandre:
Probably Anycast explanation may be skipped now, as TON Core is about to fully rewrite that functionality

Shvandre:
And also we should somewhere mention, that out Address type in tact is addr_std (at least for now). 

I am sorry if it is already stated somewhere and I just didn't find.

Issue #925, opened 2024-10-07
Add `Address` to list of basic types


anton-trunov:
The page should explain the concept of addresses in TON and also combine all the functions that either consume or produce addresses.

This page should mention `parseStdAddress` and `newAddress` are inverse functions for each other.

Somewhere here:

<img width="268" alt="image" src="https://github.com/user-attachments/assets/236e3d47-9c6f-4d0a-86a9-fe1a42ab1c9d">

anton-trunov:
Storage requirements should be explained for addresses to avoid issues like #940. We should mention it in that page.

novusnota:
Potentially depends on https://github.com/tact-lang/tact/pull/1223

anton-trunov:
> Potentially depends on https://github.com/tact-lang/tact/pull/1223

Let's resolve the issue as if that PR does not exist

novusnota:
> > Potentially depends on [#1223](https://github.com/tact-lang/tact/pull/1223)
> 
> Let's resolve the issue as if that PR does not exist

Ok, I'll try to leave the place for it and the potential future fix of https://github.com/tact-lang/tact/issues/1119

novusnota:
Note to self: add info about bounceable and non-bounceable "user-friendly" address representations. Especially that it's highly recommended to use, give and share non-bounceable addresses for wallet contracts (and why) and bounceable addresses for any other contract (and why).

https://docs.ton.org/v3/documentation/smart-contracts/addresses#bounceable-vs-non-bounceable-addresses

novusnota:
Note to self: make sure all the references of `chain` / workchain ID (including basechain and masterchain) as well as `hash` / account ID lead to this new page and are well-explained on it. Currently, all such links lead to TON Docs.

novusnota:
This!: https://docs.ton.org/v3/guidelines/dapps/cookbook#what-flags-are-there-in-user-friendly-addresses

PR #1279, opened 2025-01-03
tests: Equality on addresses, slices, cells.

jeshecdom:
## Issue

Towards #539.

## Checklist


src/test/e2e-emulated/contracts/semantics.tact
```diff
@@ -1886,4 +1911,216 @@ contract SemanticsTester {
...
+
+
+    /*************** Addresses ********************/
+
+    get fun testAddressEquality(): Bool {
+        
+        // Addresses are immutable.
+
+        let addr1 = myAddress();
+        let addr2 = contractAddress(initOf SemanticsTester());
```
anton-trunov, line 1923: Let's also add this equality to the docs, if it's not there yet

src/test/e2e-emulated/contracts/semantics.tact
```diff
@@ -1886,4 +1911,216 @@ contract SemanticsTester {
...
+
+        // Check that slice6 is now empty
+        result &&= slice6.empty();
+
+        return result;
+    }
+
+    /*************** Cells ********************/
+
+    get fun testCellEquality(): Bool {
```
anton-trunov: Let's add tests for cells obtained in a different way, not just from structs

anton-trunov (changes requested):
Cool stuff, just some minor comments

src/test/e2e-emulated/contracts/semantics.tact
novusnota: We can add something to `initOf()` for the time being. Later, this moment can be expanded upon in #925 and #1200

jeshecdom: I am not quite sure of what I need to add to the docs. Do you mean the equivalence between obtaining the contract address with `myAddress()` and indirectly through `initOf`?

novusnota: @jeshecdom yes, something like that. If you don't mind, I'll add it myself to this PR in a couple of hours :)

UPD: Actually, this will happen within this hour — I'm already at the laptop
UPD2: By the way, the `contractAddress(initOf ...())` approach is about 2.5k gas more expensive than just `myAddress()`, which only uses a single instruction.

novusnota: Done! Also improved the similar descriptions on the operators page

anton-trunov (approved):

Issue #573, opened 2024-07-14
Add documentation comments to standard library of Tact, `stdlib/`


novusnota:
They will help various tools, especially editor plugins and language servers, to show additional information in hover requests and make go to definitions look just a little bit better. This change will significantly reduce the maintenance burden on their side, while allowing core and community contributors to the Tact compiler to keep them up to date with every PR that changes or adds something to Tact's standard library, `stdlib/`.

 As a bonus, such documentation comments might even serve as an offline reference to some extent, for anyone wandering around in `node_modules/` :)

anton-trunov:
We should also auto-generate docs for `docs.tact-lang.org` from the doc comments in stdlib

novusnota:
> We should also auto-generate docs for `docs.tact-lang.org` from the doc comments in stdlib

Or vice versa, which seems to be easier to maintain

anton-trunov:
> Or vice versa, which seems to be easier to maintain

The source of truth should be the Tact source code, not vice versa.

novusnota:
> > Or vice versa, which seems to be easier to maintain
> 
> The source of truth should be the Tact source code, not vice versa.

While I usually agree with this point, in this case there's no difference in consistency — both code and docs have to be maintained equally well and written in consonant manner. Which is to say that source of truth can be in either of places, and the docs are simpler to make one from my today's POV :)

P.S.: Note to self: the code-first approach will require a lot of trouble with adding temp `[...]:` links in order to not break the docs.

UPD: Oh, and I also forgot to mention all the vocabulary (already), grammar (soon) and link (already) checks that docs have and doc comments don't.

anton-trunov:
Done

PR #1296, opened 2025-01-13
doc(stdlib): doc comments for `BaseTrait`

i582:
This PR adds more documentation, as an example of the level I would like to see for the entire standard library!

The new documentation for `BaseTrait` shows how we document code in general. 

- We use `///` for documentation comments
- Any code examples are wrapped in three apostrophes

I think we need an RFC in which we will describe all such rules that everyone in the team agrees with. Everything described above is negotiable and I will be glad to hear your thoughts on how to further change and improve the documentation rules!

Towards #573


anton-trunov:
what issue is it solving?

anton-trunov:
I would argue `cells.tact` should either be left intact, or split into two files `builder.tact` (serialization) and `slice.tact` (deserialization).

i582:
> what issue is it solving?

#573
added to description

i582:
> I would argue `cells.tact` should either be left intact, or split into two files `builder.tact` (serialization) and `slice.tact` (deserialization).

Yeah, if this file layout is ok, I'll rename this file to `cell.tact`.

There are now also functions like `asSlice` for `Builder` and `Slice` that should be placed somewhere (together in a file like `conversions.tact` or split into `cell.tact` and `builder.tact`).

anton-trunov:
> or split into cell.tact and builder.tact

this is exactly what I'm arguing against:
- it's either everything in one file called `cell.tact` (or `cells.tact` as we have now),
- or `builder.tact` and `slice.tact`: you simply cannot do anything with cells as those are opaque (except passing around as stack values)

anton-trunov (changes requested):
Let's keep the changes for `base.tact` and revert the rest of them until we have a clear plan

stdlib/std/base.tact
```diff
@@ -1,16 +1,64 @@
...
+    /// ```
+    /// contract AllYourStorageBelongsToUs {
+    ///     // This would change the behavior of `self.forward()` function,
+    ///     // causing it to try reserving this amount of nanoToncoins before
+    ///     // forwarding a message with `SendRemainingBalance` mode
+    ///     override const storageReserve: Int = ton("0.1");
+    /// }
+    /// ```
+    ///
+    /// See [`self.forward`] for more details.
```
novusnota:
```suggestion
    /// See: https://docs.tact-lang.org/ref/core-base/#self-forward
```

stdlib/std/base.tact
```diff
@@ -1,16 +1,64 @@
+/// Describes the base logic that is available in all contracts and traits by default.
+///
+/// See: https://docs.tact-lang.org/ref/core-base/
 trait BaseTrait {
+    /// The amount of nanoToncoins to reserve before forwarding a message with 
+    /// `SendRemainingBalance` mode. Default is 0 (no reserve).
+    /// ```
```
novusnota:
```suggestion
    /// `SendRemainingBalance` mode. Default is 0 (no reserve).
    ///
    /// ```tact
```

stdlib/std/base.tact
```diff
@@ -1,16 +1,64 @@
...
+    ///     override const storageReserve: Int = ton("0.1");
+    /// }
+    /// ```
+    ///
+    /// See [`self.forward`] for more details.
     virtual const storageReserve: Int = 0;
     
+    /// Sends a bounceable message back to the sender of the current message.
+    /// An alias function to calling the [`self.forward`] function with the following arguments:
+    /// ```
```
novusnota:
```suggestion
    /// Storage function. An alias to calling the `self.forward()` function with the following arguments:
    ///
    /// ```tact
```

stdlib/std/base.tact
```diff
@@ -1,16 +1,64 @@
...
+    ///              |         body: Cell?
+    ///              to: Address
+    /// ```
     virtual inline fun reply(body: Cell?) {
         self.forward(sender(), body, true, null);
     }
 
+    /// Sends a non-bounceable message back to the sender of the current message.
+    /// An alias function to calling the [`self.forward`] function with the following arguments:
+    /// ```
```
novusnota:
```suggestion
    /// Storage function. An alias to calling the `self.forward()` function with the following arguments:
    ///
    /// ```tact
```

stdlib/std/base.tact
```diff
@@ -1,16 +1,64 @@
...
+    /// also tries to reserve the [`self.storageReserve`] amount of nanoToncoins from the remaining
+    /// balance before making the send in the [`SendRemainingBalance`] (128) mode.
+    ///
+    /// In case reservation attempt fails and in the default case without the attempt, the message 
+    /// is sent with the [`SendRemainingValue`] (64) mode instead.
+    ///
+    /// > Note that [`self.forward`] never sends additional nanoToncoins on top of what’s available on the balance.
+    /// > To be able to send more nanoToncoins with a single message, use the the [`send`] function.
+    ///
+    /// See: https://docs.tact-lang.org/ref/core-base/#self-forward for examples.
```
novusnota:
```suggestion
    /// Storage function. Queues the message (bounceable or non-bounceable) to be sent to the specified address `to`. 
    /// Optionally, you may provide a `body` of the message and the `init` package with `initOf`.
    /// 
    /// When `self.storageReserve()` constant is overwritten to be greater than zero, before sending a message it
    /// also tries to reserve the `self.storageReserve()` amount of nanoToncoins from the remaining
    /// balance before making the send in the `SendRemainingBalance` (128) mode.
    ///
    /// In case reservation attempt fails and in the default case without the attempt, the message 
    /// is sent with the `SendRemainingValue` (64) mode instead.
    ///
    /// > Note that `self.forward()` never sends additional nanoToncoins on top of what’s available on the balance.
    /// > To be able to send more nanoToncoins with a single message, use the the `send()` function.
    ///
    /// See: https://docs.tact-lang.org/ref/core-base/#self-forward
```

stdlib/std/base.tact
```diff
@@ -1,16 +1,64 @@
+/// Describes the base logic that is available in all contracts and traits by default.
```
novusnota:
```suggestion
/// This trait is implicitly inherited by every other contract and trait. It contains a number of the most useful internal functions for any kind of contract, and a constant `self.storageReserve` aimed at advanced users of Tact.
```

novusnota (changes requested):
Nice job overall, I've only edited descriptions to match the rest of `///` comments and their original source — the docs.

stdlib/std/base.tact
i582: Done

stdlib/std/base.tact
i582: Fixed

stdlib/std/base.tact
i582: Fixed

stdlib/std/base.tact
i582: Fixed

stdlib/std/base.tact
i582: Fixed

stdlib/std/base.tact
i582: Fixed

CHANGELOG.md
```diff
@@ -82,6 +82,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Document that identifiers cannot start with `__gen` or `__tact`, and cannot contain Unicode characters apart from the small subset `a-zA-Z0-9_`: PR [#1312](https://github.com/tact-lang/tact/pull/1312)
 - Added signatures for map methods, such as `.get()`, `.exists()`, `.set()`, `.replace()`, `.replaceGet()`, `.del()`, `.isEmpty()`, `.deepEquals()`, `.asCell()`: PR [#1352](https://github.com/tact-lang/tact/pull/1352)
 - Added a compilation-related page with the description of the compilation report: PR [#1309](https://github.com/tact-lang/tact/pull/1309)
+- Documented `BaseTrait` and methods: PR [#1296](https://github.com/tact-lang/tact/pull/1296)
```
anton-trunov: Let's make it clear this is not a docs.tact-lang.org change, but a doc comment in stdlib

src/stdlib/stdlib/std/base.tact
```diff
@@ -1,16 +1,74 @@
...
+    ///              |         |     |     init: StateInit?
+    ///              |         |     bounce: Bool
+    ///              |         body: Cell?
+    ///              to: Address
+    /// ```
     virtual inline fun notify(body: Cell?) {
         self.forward(sender(), body, false, null);
     }
 
+    /// Storage function. Queues the message (bounceable or non-bounceable) to be sent to the specified
```
anton-trunov: What does it mean, "storage function?"

src/stdlib/stdlib/std/base.tact
```diff
@@ -26,12 +84,28 @@ trait BaseTrait {
...
                 nativeReserve(self.storageReserve, ReserveExact);
-                send(SendParameters{bounce: bounce, to: to, value: 0, mode: SendRemainingBalance | SendIgnoreErrors, body: body, code: code, data: data });
+                send(SendParameters{
+                    bounce: bounce,
+                    to: to,
+                    value: 0,
+                    mode: SendRemainingBalance | SendIgnoreErrors,
+                    body: body,
+                    code: code,
+                    data: data,
```
anton-trunov: Let's use struct field punning here: `... bounce, to, value: 0, ...`

src/stdlib/stdlib/std/base.tact
```diff
@@ -26,12 +84,28 @@ trait BaseTrait {
...
 
         // Just send with remaining balance
-        send(SendParameters{bounce: bounce, to: to, value: 0, mode: SendRemainingValue | SendIgnoreErrors, body: body, code: code, data: data });
+        send(SendParameters{
+            bounce: bounce,
+            to: to,
+            value: 0,
+            mode: SendRemainingValue | SendIgnoreErrors,
+            body: body,
+            code: code,
```
anton-trunov: field punning

anton-trunov (changes requested):

CHANGELOG.md
i582: Done

src/stdlib/stdlib/std/base.tact
i582: @novusnota's suggestion, I deleted this part for now because I also don't understand why we need this information

src/stdlib/stdlib/std/base.tact
i582: Done

src/stdlib/stdlib/std/base.tact
i582: Done

anton-trunov (approved):
LGTM, but let's give @novusnota the chance to comment

src/stdlib/stdlib/std/base.tact
novusnota: It meant a function as a part of the contract and called with `self.` prefix. The idea is similar to saying "Extension function" (`type.smth()`) or "Global function" (just `smth()`).

Perhaps, the naming should be "Internal function" or "Internal contract function" or just "Contract function".

anton-trunov: it should be "internal contract function":  https://docs.tact-lang.org/book/contracts/#internal-functions

i582: I don't think we need to specify this in a comment because it is clear from the code

novusnota: > I don't think we need to specify this in a comment because it is clear from the code

Fair, but the same redundant-ish info is provided for extension and global functions, just so the user has multiple ways to quickly get this info — from the first line of the doc and from the brief look at the code

novusnota: "For every sentence, there is a user that skipped only that sentence" — [Eric Wastl](https://youtu.be/uZ8DcbhojOw?t=979) :)

anton-trunov: you are welcome to improve it later, for now I'd prefer to move on

PR #1301, opened 2025-01-13
fix(types): forbid "remainder" field at the middle of a contract storage

### LINKED ISSUE #765: Only allow `as remaining` annotation for last fields of contract storage
A follow-up on #688.

Since, `as remaining` does not advance the storage slice when parsing the contract storage, so, for instance, the following example:

```
import "@stdlib/deploy";

contract SerializationExample with Deployable {
    someCell: Cell as remaining;
    someInt: Int as uint32;

    init() {
        self.someCell = beginCell().storeUint(42, 32).endCell();
        self.someInt = 56;
    }

    get fun foo(): Int {
        return self.someInt;
    }
}
```

results in `42` when the `foo` getter is called, while `56` is expected. Hence this should be forbidden.

While `as remaining` for the last field is probably never used, I can see a potential use case: the user might want to deploy a contract with a data cell whose *prefix* as formatted according to the contract storage description, while the last field with `as remaining` annotation would be any extra data that is placed after the prefix: for instance this might be used as a "forward payload" for which we are not aware of its internal structure.

i582:
## Issue

Closes #765.

~~This PR actually contains a tiny refactoring, now we have the `checkAggregateTypes` function with only one loop over all types, instead of two as before (it just looked a bit weird after the initial changes I made). However, I don't think we need 2 PRs here (implementation and refactoring :)~~ False way

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `VarInt16`, `VarInt32`, `VarUint16`, `VarUint32` integer serialization types: PR [#1186](https://github.com/tact-lang/tact/pull/1186)
 - `unboc`: a standalone CLI utility to expose Tact's TVM disassembler: PR [#1259](https://github.com/tact-lang/tact/pull/1259)
 - Added alternative parser: PR [#1258](https://github.com/tact-lang/tact/pull/1258)
-- 'The "remainder" field can only be the last field' inspection for contracts: PR [#1301](https://github.com/tact-lang/tact/pull/1301)
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `VarInt16`, `VarInt32`, `VarUint16`, `VarUint32` integer serialization types: PR [#1186](https://github.com/tact-lang/tact/pull/1186)
 - `unboc`: a standalone CLI utility to expose Tact's TVM disassembler: PR [#1259](https://github.com/tact-lang/tact/pull/1259)
 - Added alternative parser: PR [#1258](https://github.com/tact-lang/tact/pull/1258)
+- 'The "remainder" field can only be the last field' inspection for contracts: PR [#1301](https://github.com/tact-lang/tact/pull/1301)
```
anton-trunov: It's a bug fix, not a feature: please check the counter-example I added to the linked issue

anton-trunov (changes requested):

anton-trunov (approved):

PR #1303, opened 2025-01-13
feat(types): allow serialization specifiers for trait fields

### LINKED ISSUE #655: Allow serialization specifiers for trait variables
The following Tact module does not compile:
```
trait Trait {
    id: Int as uint32;
}
```

Compilation fails with

```
Trait field cannot have serialization specifier
```

i582:
## Issue

Closes #655.

## Checklist


src/types/resolveDescriptors.ts
```diff
@@ -1552,6 +1546,17 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
                         `Trait "${tr.name}" requires field "${f.name}" of type "${printTypeRef(f.type)}"`,
                         t.ast.loc,
                     );
+                } else if (f.as !== null && f.as !== ex.as) {
+                    const expected = `${printTypeRef(f.type)} as ${f.as}`;
+                    let actual = printTypeRef(ex.type);
+                    if (ex.as !== null) {
+                        actual += ` as ${ex.as}`;
+                    }
```
anton-trunov: let's use a conditional expression here and remove mutation, please check our code style guide: https://github.com/tact-lang/tact/blob/main/STYLEGUIDE.md

src/types/test/trait-without-as-typed-field-inherited-with-it.tact
anton-trunov, line 1: This test must fail: `f: Int` means exactly `f: Int as int257`

(I have a feeling this might result in some existing contracts breaking, but this needs to be fixed, since trait functions might manually parse `f: Int` as `int257`)

anton-trunov (changes requested):
Let's also add some diamond-like inheritance tests

src/types/resolveDescriptors.ts
i582: fixed

src/types/test/trait-without-as-typed-field-inherited-with-it.tact
i582: Fixed, we now require exact `as` type match (except `int257` case)

src/types/resolveDescriptors.ts
```diff
@@ -1552,6 +1550,27 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
...
                         t.ast.loc,
                     );
+                } else if (
+                    (f.as !== null || ex.as !== null) &&
+                    f.as !== ex.as
+                ) {
+                    if (
+                        (ex.as === "int257" && f.as === null) ||
+                        (ex.as === null && f.as === "int257")
+                    ) {
```
anton-trunov: let's merge all these conditions: this part is too complex and having `f.as !== null` and `f.as === null` makes me think it can be simplified

anton-trunov (changes requested):

src/types/resolveDescriptors.ts
i582: yeah, the new version is much more simple

src/types/resolveDescriptors.ts
```diff
@@ -1552,6 +1550,18 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
                         `Trait "${tr.name}" requires field "${f.name}" of type "${printTypeRef(f.type)}"`,
                         t.ast.loc,
                     );
+                } else if (
+                    f.as !== ex.as &&
+                    f.as !== "int257" &&
+                    ex.as !== "int257"
```
anton-trunov: I think something like

```tact
trait Trait {
    i: Int as int257;
}

contract Foo with Trait {
    i: Int as coins = 0;
}
```

won't be reported

i582: Indeed, I simplified the conditions too much, the new ones should work fine

anton-trunov (approved):

PR #1334, opened 2025-01-14
feat: support block statements

### LINKED ISSUE #362: Block statements are missing from ASTs
Block statements are a [part](https://github.com/tact-lang/tact/blob/9915bd1da208520cd5e823068aeb7cac003d40db/src/grammar/grammar.ohm#L91) of the Tact grammar but those are not processed when converting Ohm.js's CSTs to ASTs.

Hence this code snippet

```
fun foo() {
    {
        // I'm a block statement
        let x: Int = 42;
    }
}
```

results in the following error message:

```
Missing semantic action for 'StatementBlock' in operation 'resolve_statement'. 
Action stack (most recent call last): 
  resolve_program > Program 
  resolve_program_item > default action for 'ProgramItem' 
  resolve_program_item > StaticFunction 
  resolve_declaration > default action for 'Function' 
  resolve_declaration > Function_withVoid 
  resolve_statement > default action for 'Statement' 
  resolve_statement > StatementBlock 
```

i582:
This PR adds full support for standalone `{}` blocks:

```tact
let a = 100;
{
   let b = 200;
   dump(a + b);
}
{
   let b = 300;
   dump(a + b);
}
dump(a);
```

Each block represents its own scope, so the user can reuse the same variable names within subsequent block statements (but not nested!).

## Issue

Closes #362.

## Checklist

      Looks like we already have documentation for blocks: https://docs.tact-lang.org/book/statements/#block


src/types/stmts/var-scope-subsequent-block-statements.tact
anton-trunov, line 1: Would be nice to add a test like this to e2e tests and interpreter tests

anton-trunov (changes requested):
Nice stuff!

Let's add a few more tests:
- negative test showing that after a block statement a variable defined inside the block is no longer accessible;
- positive and negative return reachability analysis tests (`return-analysis-*.tact` files)
- interpreter / const-eval tests with functions with block statements

src/optimizer/interpreter.ts
```diff
@@ -1807,4 +1811,8 @@ export class Interpreter {
             } while (condition.value);
         });
     }
+
+    public interpretBlockStatement(ast: AstStatementBlock) {
+        ast.statements.forEach(this.interpretStatement, this);
```
jeshecdom: The method implementation should be changed to this:
```
public interpretBlockStatement(ast: AstStatementBlock) {
     this.envStack.executeInNewEnvironment(() => {
                ast.statements.forEach(this.interpretStatement, this);
            });
}
```
The reason is that variables declared inside the block statement should not be accessible after the block finishes. The method `executeInNewEnvironment` adds a fresh environment on top the current environment stack and executes `ast.statements.forEach(this.interpretStatement, this)` inside this fresh environment. 
Once `ast.statements.forEach(this.interpretStatement, this)` finishes execution, method `executeInNewEnvironment` pops the fresh environment from the stack, effectively discarding any variable created inside the block (but variables declared outside the block and modified inside the block remain modified after the block).

src/test/contracts/renamer-expected/case-block-statements.tact
```diff
@@ -0,0 +1,12 @@
+contract contract_0 {
```
jeshecdom, line 1: This test and the one in `src/test/contracts/case-block-statements.tact` are identical (by ignoring names for contracts and functions). Probably because this PR is still a draft, but just wanted to point it out.

src/optimizer/interpreter.ts
i582: Thank you for the explanation! Fixed

src/test/contracts/renamer-expected/case-block-statements.tact
i582: I'm not sure if I did it right, but all tests in `src/test/contracts/` have the test file with the same name in the `renamer-expected/` directory, differing only in contract/function/etc. names 🤔 

@anton-trunov is it the correct way?

src/types/stmts/var-scope-subsequent-block-statements.tact
i582: added

i582:
Thank you guys for your help! This feature requires changes in many places, your help really helped me understand all (or most 😄) of the necessary places.

I added use after block test, reachability tests, and e2e tests for both runtime and interpreter (via functions that initialize the constant)

src/test/contracts/renamer-expected/case-block-statements.tact
anton-trunov: Yeah, that should be correct, but I'm not entirely sure (for the life of me cannot remember that)

anton-trunov (approved):

Issue #315, opened 2024-05-01
Add cram-like tests for Tact CLI


anton-trunov:
Would be nice to have [cram](https://pypi.org/project/cram/)-like tests. It would be nice to stay within the JS/TS/Node.js ecosystem. On the other hand, something like [oclif testing](https://oclif.io/docs/testing) looks like too much JS. It should be based on snapshots, i.e. tests should look a lot like shell interaction histories, resembling e.g. [this report](https://github.com/tact-lang/tact/pull/287#issuecomment-2088399034). Btw, the cram-like framework we are going to pick should definitely support updating those snapshots.

novusnota:
Here's the cram fork we might use, which continues developing its snapshot-based testing approach: https://github.com/prysk/prysk

anton-trunov:
> Here's the cram fork we might use, which continues developing its snapshot-based testing approach: https://github.com/prysk/prysk

As mentioned in the original post, I'd very much prefer not to mix the Node.js and Python ecosystem within one project.

novusnota:
What if we only use it in CI? Although I agree with your point.

anton-trunov:
We might end up doing that, but only as the last resort, if there are no alternatives with our ecosystem

novusnota:
Jest has support for snapshot testing, which is probably the most compatible solution to our current repo layout: https://jestjs.io/docs/snapshot-testing

Although we'll have to wrap CLI calls a bit (in `child_process` calls or something like that)

anton-trunov:
Jest does have snapshot tests and we use it in Tact. But I also used cram-like tests a lot with OCaml (https://dune.readthedocs.io/en/stable/reference/cram.html) and it's so much better for CLI apps testing than what Jest or oclif is offering.

PR #1375, opened 2025-01-17
fix: show stacktrace of a compiler error only in verbose mode

### LINKED ISSUE #1360: Don't show stacktrace from compiler after compiler error by default
For example:

```
Error: sources/contract.tact:10:5: Unsupported format "uint322"
Line 10, col 5:
   9 |     owner: Address;
> 10 |     counter: Int as uint322;
           ^~~~~~~~~~~~~~~~~~~~~~~
  11 | 

    at throwCompilationError (/Users/petrmakhnev/tact-template/node_modules/@tact-lang/compiler/dist/errors.js:81:11)
    at resolveABIType (/Users/petrmakhnev/tact-template/node_modules/@tact-lang/compiler/dist/types/resolveABITypeRef.js:49:56)
    at buildFieldDescription (/Users/petrmakhnev/tact-template/node_modules/@tact-lang/compiler/dist/types/resolveDescriptors.js:308:61)
    at resolveDescriptors (/Users/petrmakhnev/tact-template/node_modules/@tact-lang/compiler/dist/types/resolveDescriptors.js:347:38)
    at precompile (/Users/petrmakhnev/tact-template/node_modules/@tact-lang/compiler/dist/pipeline/precompile.js:18:55)
    at build (/Users/petrmakhnev/tact-template/node_modules/@tact-lang/compiler/dist/pipeline/build.js:62:43)
    at Object.run (/Users/petrmakhnev/tact-template/node_modules/@tact-lang/compiler/dist/node.js:101:47)
```

This stacktrace looks scary and doesn't add any useful information to the user, let's show it only with the `--dev` flag enabled

#183

i582:
## Issue

Closes #1360.

## Checklist


anton-trunov:
We also need to add some tests for this flag.
For now we have some CLI tests in CI: https://github.com/tact-lang/tact/blob/4dbbf403beec34b3b068f71c26d776b722c62689/.github/workflows/tact.yml#L110, until we have some proper cram-like tests: #315.

anton-trunov:
We also need to be able to active the verbose mode via [blueprint](https://github.com/ton-org/blueprint) which basically ignores `tact.config.json`: https://github.com/ton-org/blueprint/issues/138

.github/workflows/tact.yml
```diff
@@ -119,6 +119,27 @@ jobs:
...
+        run: |
+          set -o pipefail
+
+          output=$(tact src/test/compilation-failed/contracts/const-eval-ascii-empty.tact 2>&1 || true)
+          echo "$output" | (! grep -E ".*at \w+ \(.*")
+
+          output=$(tact src/test/compilation-failed/contracts/contract-duplicate-bounced-opcode.tact 2>&1 || true)
+          echo "$output" | (! grep -E ".*at \w+ \(.*")
+
+      - name: CLI Test | Check compilation of broken contract contains stacktrace with --verbose flag
```
novusnota:
```suggestion
      - name: CLI Test | Check compilation of broken contract contains stacktrace with --verbose flag
        if: runner.os != 'Windows'
```

novusnota: Also, since you're adding a `tact.config.json` key, it would be nice to adjust:
1. `src/config/configSchema.json`
2. `docs/src/content/docs/book/config.mdx`

anton-trunov (changes requested):
Let's fix CI for macOS and also resolve comments by @novusnota.

Although, looks like it's too hard to work with the current CLI tests since cross-platform scripts are harder even for Unix-like OSes.

.github/workflows/tact.yml
i582: Thanks for pointing that out! Added

src/stdlib/stdlib.ts
anton-trunov, line 1: I'm wondering why stdlib.ts got changed here

src/config/configSchema.json
```diff
@@ -70,6 +70,11 @@
             "enum": ["fullWithDecompilation", "full", "funcOnly", "checkOnly"],
             "title": "Compilation mode of the project. In Blueprint, it's always set to `full` and cannot be overwritten.",
             "description": "Set to `full` by default, which runs the whole pipeline of the compilation and emits FunC code, BoC, and various utility files, including wrappers for TypeScript.\nIf set to `fullWithDecompilation`, does full compilation and also decompiles produced binary code in the BoC format.\nIf set to `funcOnly`, only outputs intermediate FunC code, preventing further compilation.\nIf set to `checkOnly`, only performs syntax and type checking, preventing further compilation."
+          },
+          "verbose": {
+            "type": "boolean",
```
anton-trunov: What if we created a custom type here for better control of the level of verbosity? For instance, I probably would also cut down the usual output from Tact where it outputs the build stages. This could also be helpful for debugging optimizations or some other complex stuff.

anton-trunov (changes requested):

src/config/configSchema.json
i582: Great idea! Let's make a separate PR (I'll create an issue for discussion about possible values) that will be done after we rewrite the CLI to TypeScript or when we have an agreement on possible values?

src/stdlib/stdlib.ts
i582: Fixed

src/config/configSchema.json
anton-trunov: That would mean breaking CLI's interface. In general, we don't promise to keep CLI, TS API and other things like that backwards compatible, but still let's not do that if we can foresee changes. Breaking compatibility only temporarily before making the next release if fine, though.

anton-trunov: Btw, that means we cannot document this in docs.tact-lang.org, because it is deployed after the PR is merged (and also should have "Available since Tact 1.6" badge)

i582: I see let's use a number as the verbosity level, what do you think?

```
--verbose 1
--verbose 5
```

i582:
```
--verbose LEVEL             Set verbosity level (higher = more details, default: 0)
```

i582:
Let's add actual documentation, like
```
0 - no output
1 - current without stack trace (default)
2 - current
```
after we implement level 0

src/config/configSchema.json
novusnota: > I see let's use a number as the verbosity level, what do you think?
> 
> ```
> --verbose 1
> --verbose 5
> ```

Whatever we use, people would have to look up the `--help`, since it's easy to forget which value corresponds to which verbosity level, so anything's fine.

Thus, number levels with 0 meaning "no output" seems like a good option :)

But the default might not be 0, since it should probably correspond to `-q, --quiet` flag. Alternatively, the `-q, --quiet` can represent -1, and then the default is 0.

i582: True, we will use 1 as default for now

.github/workflows/tact.yml
novusnota: Great!

.github/workflows/tact.yml
```diff
@@ -119,6 +119,36 @@ jobs:
           then false
           fi
 
+      - name: CLI Test | Check compilation of broken contract doesn't contain stacktrace
+        if: runner.os != 'Windows'
+        run: |
+          output=$(tact src/test/compilation-failed/contracts/const-eval-ascii-empty.tact 2>&1 || true)
+          if echo "$output" | grep -E ".*at \w+ \(.*"; then
```
anton-trunov, line 126: Should we include `at Object.run (` here to make the pattern more robust?

src/config/configSchema.json
```diff
@@ -70,6 +70,11 @@
             "enum": ["fullWithDecompilation", "full", "funcOnly", "checkOnly"],
             "title": "Compilation mode of the project. In Blueprint, it's always set to `full` and cannot be overwritten.",
             "description": "Set to `full` by default, which runs the whole pipeline of the compilation and emits FunC code, BoC, and various utility files, including wrappers for TypeScript.\nIf set to `fullWithDecompilation`, does full compilation and also decompiles produced binary code in the BoC format.\nIf set to `funcOnly`, only outputs intermediate FunC code, preventing further compilation.\nIf set to `checkOnly`, only performs syntax and type checking, preventing further compilation."
+          },
+          "verbose": {
+            "type": "number",
+            "default": false,
```
anton-trunov:
```suggestion
            "default": 1,
```

.github/workflows/tact.yml
i582: I don't think we can call this type of testing robust in any case, since we need a proper way to test the CLI :)

src/config/configSchema.json
i582: fixed

.github/workflows/tact.yml
i582: Also for constant evaluation errors there is no `Object.run`

anton-trunov: Ah, nice catch!

anton-trunov (approved):

PR #1393, opened 2025-01-20
test: add tests for asm instructions without whitespace

### LINKED ISSUE #1268: Add grammar tests for whitespace after `asm` instructions before `}`
The new parser merged in #1258 fixed issue #1265, but grammar tests are missing.

xpyctumo:
## Issue

Closes #1268

## Checklist

~~**Also, we need fix old parser to handle this tests**~~
UPD: As a solution - tests that only require to use **new parser** should be placed in `src/grammar/test/next/test`


src/grammar/next/test-failed/asm-instruction-invalid-syntax.tact
```diff
@@ -0,0 +1,3 @@
+asm fun Bar(): Int {-ROT ADD MUL};
```
anton-trunov:
```suggestion
asm fun Bar(): Int {-ROT ADD MUL}
```

src/grammar/next/grammar.spec.ts
```diff
@@ -9,21 +9,32 @@ expect.addSnapshotSerializer({
...
-            const { parse } = getParser(ast, "new");
-            expect(parse(r.code, "<unknown>", "user")).toMatchSnapshot();
-        });
+    const shouldParsePaths = [__dirname + "/../test/", __dirname + "/test/"];
+
+    for (const path of shouldParsePaths) {
+        for (const r of loadCases(path)) {
+            it("should parse " + r.name, () => {
+                const ast = getAstFactory();
+                const { parse } = getParser(ast, "new");
```
anton-trunov, line 18: this should be moved out of the loop and also reused in the new tests for the new parser

anton-trunov: ah, this suggestion is incorrect, since `id`s would be shared

src/grammar/next/test-failed/asm-instruction-invalid-syntax.tact
xpyctumo: Thats an example of failed test.  Just so there's a `test-failed` folder.

xpyctumo: Do you think we don't need to do `test-failed` separately?

xpyctumo:
Ouch. CI on Windows was failed. Quite a surprise

reverofevil:
> Ouch. CI on Windows was failed. Quite a surprise

At the moment it's expected

src/grammar/next/test-failed/asm-instruction-invalid-syntax.tact
reverofevil: Probably we should do a different kind of failing test, because in this case it's not the asm instruction parser that is failing.

Maybe something like `asm fun Bar() Int`.

reverofevil: Oh wait, we already have a `test-failed` folder, this test is not related to whitespaces, and I don't see why would we need this test.

verytactical: We already have `test-failed` folder, and this test is not about whitespace.

xpyctumo: I removed that

verytactical (approved):

PR #1419, opened 2025-01-20
Fixed flag name in help (`--project` to `--projects`)

### LINKED ISSUE #1273: `--project` flag doesn't exist
While `--help` reports there is a flag `--project`

```
-p, --project ...names      Build only the specified project name(s) from the config file
```

in fact it doesn't exist

```
node bin/tact.js --project a b
Unknown flag
--project
```

and is called `--projects`.

Bug prevention measures:
- Generate help message automatically
- Rewrite CLI code to TypeScript

i582:
## Issue

Closes #1273.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Remove duplicate line and column info from error messages: PR [#1362](https://github.com/tact-lang/tact/pull/1362)
 - Support `AstTypedParameter` AST node in pretty printer: PR [#1347](https://github.com/tact-lang/tact/pull/1347)
 - Show stacktrace of a compiler error only in verbose mode: PR [#1375](https://github.com/tact-lang/tact/pull/1375)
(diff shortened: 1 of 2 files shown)


novusnota (changes requested):
Hold on, the `-p, --project` flag is correct since it denotes a name of the project to build and it can be specified multiple times to selectively build multiple projects.

The issue wasn't with the help text, but with the actual CLI flag — it's `--projects` now, but should be just like in the help, i.e. `--project`

i582:
> Hold on, the `-p, --project` flag is correct since it denotes a name of the project to build and it can be specified multiple times to selectively build multiple projects.
> 
> The issue wasn't with the help text, but with the actual CLI flag — it's `--projects` now, but should be just like in the help, i.e. `--project`

What about current users who use the current `--projects` flag? If we change it to `--project`, we break all code that depends on the `--projects` flag

anton-trunov:
@novusnota since we allow specifying multiple projects, it's only logical to call it `--projects` (plural)

anton-trunov (approved):

PR #1425, opened 2025-01-20
chore(deps): updated `typescript-eslint`

### LINKED ISSUE #1430: Update typescript-eslint
When running `eslint` locally, there's an error message

```
=============
WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.
You may find that it works just fine, or you may not.
SUPPORTED TYPESCRIPT VERSIONS: >=4.7.4 <5.6.0
YOUR TYPESCRIPT VERSION: 5.6.3
Please only submit bug reports when using the officially supported version.
=============
```

This error was encountered by @Danil42Russia, @i582.

Do not downgrade `typescript` version. Update `typescript-eslint`.

Danil42Russia:
## Issue

Closes #1430.

There will likely be errors on CI that I couldn't catch locally, I'll fix them. Also maybe some of my solutions to errors are not suitable for you, I am ready to fix them.

## Checklist


src/grammar/next/index.ts
```diff
@@ -375,6 +375,7 @@ const parseParens = ({ child }: $ast.Parens): Handler<A.AstExpression> => {
 };
 
 // has to be an interface because of the way TS handles circular type references
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
 interface Binary extends $ast.Binary<Expression, A.AstBinaryOperation> {}
```
Danil42Russia, line 380: That's the error I couldn't beat. As a temporary solution, I suppressed it

verytactical: I think it's completely fine to ignore it here, and you did the right thing.

This "empty" type is a workaround for a buggy implementation of recursive types in TS, and linter has no way to know about it.

src/utils/tricks.ts
```diff
@@ -106,11 +106,12 @@ type Intersect<T> = (T extends unknown ? (x: T) => 0 : never) extends (
  */
 type Unwrap<T> = T extends infer R ? { [K in keyof R]: R[K] } : never;
 
-type Inputs<I, T extends string> = I extends { [Z in T]: infer K }
-    ? K extends string
-        ? Record<K, (input: I) => unknown>
-        : never
-    : never;
+type Inputs<I, T extends string> =
+    I extends Record<T, infer K>
```
verytactical: This should be checked manually. In TS a type and alias to the same type are not the same thing, because they implement aliases as instantations, in C++ way.

I'd rather prefer to disable that rule, because it's mostly pointless.

tsconfig.base.json
```diff
@@ -6,7 +6,7 @@
     "target": "esnext" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', 'ES2021', or 'ESNEXT'. */,
     "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */,
     "lib": ["ESNext"],
-    // "allowJs": true,                             /* Allow javascript files to be compiled. */
+    "allowJs": true /* Allow javascript files to be compiled. */,
```
verytactical: Why do we need this? Tact is TS-only project, and `allowJs: false` seems to be some kind of a lint.

anton-trunov (changes requested):
Let's resolve the comments by @verytactical. (Doing this request to change the status of this PR to "Changes Requested")

src/grammar/next/index.ts
Danil42Russia: Okay, shall we close this conversation?

tsconfig.base.json
Danil42Russia: Ohhhh... Good comment, I've been thinking for a long time about whether to include `allowJs` or not.
Generally yes, look in the `tsconfig.eslint.json` file, `.js` files are specified there. If you do not enable this parameter, `typescript-eslint/parser` will generate an error.

src/utils/tricks.ts
Danil42Russia: Yes, I can do that, but then the code style may start to vary. You can configure it for yourself: https://typescript-eslint.io/rules/consistent-indexed-object-style/

If you think it should be disabled, I will do it.

verytactical: At the very least disable it in this file. This code is, well, quite tricky :)

tsconfig.base.json
verytactical: Sounds like a problem. `allowJs` means "start checking JS files with TS", i.e. it will start parsing JSDocs inside of them. Worst part is that it will try checking huge generated `.js` file with func-js, and that will likely increase typechecking time by a lot.

EDIT. Oh, also there's a 120KB generated .js file with legacy grammar.

src/utils/tricks.ts
Danil42Russia: fixed

src/cli/arg-parser.ts
```diff
@@ -134,8 +134,7 @@ export const ArgParser = (Errors: ArgParserErrors) => {
         },
     });
 
-    // eslint-disable-next-line @typescript-eslint/ban-types
-    const tokenizer = makeTokenizer<{}>((argv) => {
+    const tokenizer = makeTokenizer<object>((argv) => {
```
reverofevil, line 137: Please don't change types here.

tsconfig.base.json
Danil42Russia: disabled the `javascript` code check

anton-trunov (approved):

i582:
https://github.com/user-attachments/assets/77b65b45-ac0a-4ca8-ae71-00825281544d

PR #1448, opened 2025-01-21
ci: check dead links and broken file references

### LINKED ISSUE #755: Add broken links checker to CI
For instance this one: https://github.com/lycheeverse/lychee-action (saw it in https://github.com/nowarp/misti/pull/110)

novusnota:
With careful separation of internal documentation and public Tact docs. And introduction of a scheduled workflow (once a week) that checks external (https) links and auto-creates issues when such checks fail.

Also, resolved issues with some links. For example, the Certik blogpost gives 403, i.e. automatic link scans are probably detected and prohibited — and thus it had to be added to exceptions. Since the CI args in the public docs case are already rather long, I've decided to make a `.lycheeignore` file and put that blogpost there.

## Issue

Closes #755.

## Checklist


verytactical:
In my opinion, CI shouldn't check external links. Third party services are often going down, and we'll have to fix every link before we'll be able to merge any PRs.

I'd rather keep it to only internal links, and create an [optional](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) or [regular](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule) job for external links.

.github/workflows/external-links.yml
```diff
@@ -0,0 +1,55 @@
...
+          fail: false
+          failIfEmpty: false
+
+      - name: (docs) Create issue from the link-checking report
+        if: steps.lychee.outputs.exit_code != 0
+        uses: peter-evans/create-issue-from-file@v5
+        with:
+          title: "ci: Link checker report (docs)"
+          content-filepath: "./out.md"
+          labels: "kind: bug, kind: docs"
```
anton-trunov: So, if an external service is down at the moment of the check, we will get irrelevant issues? I think just a notification is enough

anton-trunov:
macOS job unexpectedly failed

.github/workflows/external-links.yml
novusnota: Removed

.github/workflows/external-links.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: (docs) Check broken HTTPS links
+        uses: lycheeverse/lychee-action@v2
+        id: lychee
+        with:
+          args: >
+            -n -s https --exclude-path docs/node_modules
+            --base 'https://docs.tact-lang.org'
+            --exclude '.*?\\.(?:jpg|png)'
+            docs/README.md './docs/**/*.mdx'
+          output: "./out.md"
```
anton-trunov: Would we be able to see the contents of `out.md`? and `out_dev.md`?

Perhaps it's better to output to the console?

novusnota: Yeah, I'll replace it with `/dev/stdout`

anton-trunov (approved):

PR #1450, opened 2025-01-21
refactor: CLI

### LINKED ISSUE #1104: Remove dependence on fast-json-patch
Dependabot complains about it: https://github.com/tact-lang/tact/security/dependabot/23.

It's used only by `ajv-cli` which we use in CI to validate JSON schemas. We should switch to something else.

### LINKED ISSUE #1295: Convert entrypoint to TS
Entrypoint (tact.js) is in JS. We can't easily convert it to TS, because we're using `meow` library that has no typings.

Remove `meow` dependency.

Related to https://github.com/tact-lang/tact/issues/1426

### LINKED ISSUE #1396: VFS refactor
# Issues
 
- VFS uses absolute paths
- VFS can use `\` separators on Windows
- `resolve` is not path concatenation, does more operations in background
- Browser does not have `"fs"`, `"path"` or other Node modules, but we can accidentally import them
- `__dirname` will include `\` on Windows
- It's unclear if `../` paths can get out of VFS
- VFS is not async, cannot be used in parallel builds

# Criteria

- No imports of Node builtin modules (`"path"`, `"fs"`, `"process"`) in most of compiler code
- No uses of `__dirname` of `__filename` outside of `docs`, CLI, test utils and `*.build.ts` files

# Steps
- `src/func/funcCompile.spec.ts`: make FunC wrapper part of func stage
- `loadCases(__dirname + "/../test/")` should become an export `positiveCases`
- `src/imports/resolveImports.spec.ts`: pass POSIX paths to `createNodeFileSystem`
- Make `src/imports/resolveImports.spec.ts` use relative paths
- `src/storage/resolveAllocation.spec.ts`: inline `primitive Int;`
- `src/test/compilation-failed/util.ts`: pass `loadConfig` result to node's `run`
- `src/test/util.ts`: delete `trimTrailingCR`, do not export `CONTRACTS_DIR`, use immediately with `loadCases`
- `src/vfs/createNodeFileSystem.spec.ts`: should take only posix paths on input, should always return paths relative to VFS root
- `string` -> `Path` in VFS
- Remove dependency on `path-normalize`

Related to https://github.com/tact-lang/tact/issues/1426

From @byakuren-hijiri: Improve workflow with `{Node,Virtual}FileSystem` API. These functions should have improved documentation, e.g. noting that they [do not support relative paths](https://github.com/nowarp/misti/blob/592b65dcc3f097dbfc4517e0171a3639c8516bb2/src/driver.ts#L18). Additionally, we might consider encapsulating them making them private to make users of the Tact API only interact with absolute paths when working with compiler internals.

### LINKED ISSUE #1487: Remove `meow` from `unboc`
Everything here is done the way it was done with `bin/tact.js` [here](https://github.com/tact-lang/tact/pull/1450).

Move most of the `bin/unboc.js` to `src/cli/unboc.ts`.

Move CLI tests from `.github/workflows/tact.yml` to `src/cli/unboc.e2e.spec.ts`.

Remove `meow` from `package.json`.

verytactical:
## Issue

Closes #1295.
Closes #1487.
Closes #1104.

## Checklist


src/cli/index.ts
```diff
@@ -0,0 +1,311 @@
...
+        await parseArgs(E, A);
+    } else {
+        showHelp();
+    }
+};
+
+const ArgSchema = (P: ArgParser) => {
+    return P.tokenizer
+        .add(P.string('config', 'c', 'CONFIG'))
+        .add(P.string('project', 'p', 'NAME'))
```
i582: Why `project`? If we rename this flag from `projects`, it will break existing code that depends on this flag

i582: Also see: https://github.com/tact-lang/tact/pull/1419#issuecomment-2603072297

verytactical: Because it takes a single project.

anton-trunov: Actually, I got confused by the doc description. We should have kept `--project` since `tact --projects foo bar` is not allowed

i582:
Looks like related to #1295

anton-trunov (dismissed):

anton-trunov (approved):

PR #1476, opened 2025-01-21
refactor: enable absolute paths in TypeScript

### LINKED ISSUE #1325: Support absolute imports in TS
Relative import, such as `from "../../foo/bar"`, are problematic when it comes to refactoring. It requires IDE that supports moving files around with updating references.

Enable absolute paths in TS, as it was done in webide ([here](https://github.com/tact-lang/web-ide/blob/main/tsconfig.json#L18-L20))

i582:
😄
God bless WebStorm 

## Issue

Closes #1325.

## Checklist

### DIFF
--- a/examples/contracts.build.ts
+++ b/examples/contracts.build.ts
@@ -1,10 +1,10 @@
 import fs from "fs";
-import { run } from "../src/node";
+import { run } from "@/node";
 import path from "path";
 import { glob } from "glob";
 import { verify } from "./verify";
-import { Logger } from "../src/context/logger";
-import { __DANGER__disableVersionNumber } from "../src/pipeline/version";
+import { Logger } from "@/context/logger";
+import { __DANGER__disableVersionNumber } from "@/pipeline/version";
 
 const configPath = path.join(__dirname, "tact.config.json");
 const packagesPath = path.resolve(__dirname, "output", "*.pkg");
--- a/examples/verify.ts
+++ b/examples/verify.ts
@@ -1,10 +1,10 @@
 import normalize from "path-normalize";
 import { Cell } from "@ton/core";
-import { Config, Options } from "../src/config/parseConfig";
-import { ILogger, Logger } from "../src/context/logger";
+import { Config, Options } from "@/config/parseConfig";
+import { ILogger, Logger } from "@/context/logger";
 import { PackageFileFormat, run } from "../src";
-import { fileFormat } from "../src/packaging/fileFormat";
-import { getCompilerVersion } from "../src/pipeline/version";
+import { fileFormat } from "@/packaging/fileFormat";
+import { getCompilerVersion } from "@/pipeline/version";
 
 export type VerifyResult =
     | {
--- a/examples/wallet.spec.ts
+++ b/examples/wallet.spec.ts
@@ -2,7 +2,7 @@ import { storeTransfer, Transfer, Wallet } from "./output/wallet_Wallet";
 import { Blockchain, SandboxContract, TreasuryContract } from "@ton/sandbox";
 import { beginCell, toNano } from "@ton/core";
 import { sign } from "@ton/crypto";
-import { testKey } from "../src/utils/testKey";
+import { testKey } from "@/utils/testKey";
 import "@ton/test-utils";
 
 describe("wallet", () => {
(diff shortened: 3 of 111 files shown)


i582:
Looks like we have problems with scripts 🤔 

```
Error: Cannot find module '@/bindings/writeTypescript'
Require stack:
- /Users/petrmakhnev/tact/src/pipeline/build.ts
- /Users/petrmakhnev/tact/src/node.ts
- /Users/petrmakhnev/tact/examples/contracts.build.ts
    at Function.<anonymous> (node:internal/modules/cjs/loader:1246:15)
    at Function.Module._resolveFilename.sharedData.moduleResolveFilenameHook.installedValue [as _resolveFilename] (/opt/homebrew/lib/node_modules/ts-node/node_modules/@cspotcode/source-map-support/source-map-support.js:811:30)
    at Function._load (node:internal/modules/cjs/loader:1072:27)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:216:24)
    at Module.require (node:internal/modules/cjs/loader:1337:12)
    at require (node:internal/modules/helpers:139:16)
    at Object.<anonymous> (/Users/petrmakhnev/tact/src/pipeline/build.ts:3:1)
    at Module._compile (node:internal/modules/cjs/loader:1572:14)
    at Module.m._compile (/opt/homebrew/lib/node_modules/ts-node/src/index.ts:1618:23) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/petrmakhnev/tact/src/pipeline/build.ts',
    '/Users/petrmakhnev/tact/src/node.ts',
    '/Users/petrmakhnev/tact/examples/contracts.build.ts'
  ]
}
```

But looks like we can solve it with:

```
npm install tsconfig-paths --save-dev
ts-node -r tsconfig-paths/register examples/contracts.build.ts
```

Danil42Russia:
@i582, the problem you described is normal for `typescript` and `ts-node`. The compiler itself does not do any transformations, but uses only for imports.
So `ts-node` only looks for modules in `node_modules`, to solve this problem, you made the right decision to start using the `tsconfig-paths` library.
But writing `-r tsconfig-paths/register` everywhere is not the most convenient solution. I can offer mine as my own:
You just need to write in `tsconfig.base.json` 
```json
"ts-node": {
   "require": ["tsconfig-paths/register"]
}
```
After that, you should have everything working without adding flags for `ts-node`.

But don't forget that the alias will remain during the build, and you'll have to pass `tsconfig-paths/register` for `javascript` as well. You can try `typescript-transform-paths` to solve the `javascript` paths problem after build.

verytactical:
Oof. I have _really bad news_.

Basically we'll have to add a build step that does something with it, because not everyone runs compiler with `ts-node`.

image

i582:
> Oof. I have _really bad news_.
> 
> Basically we'll have to add a build step that does something with it, because not everyone runs compiler with `ts-node`.
> 

Yeah, it seems like this task is a bit difficult to do right 🤔

verytactical:
[Here is](https://github.com/tact-lang/tact/pull/1503/files) an _example_.

Please beware I haven't run resulting tact executable even once. We have to `npm pack` a package and check that it works by `yarn install path.to.tgz` into some fresh `npm init -y`ed project.

Most likely it will need more care, because `.d.ts` sometimes have `type Foo = typeof import('...')` types, and I haven't checked for those. Also there might be some `require()` generated by dynamic imports (`async import("...")`), but we don't use those _yet_.

Probably as a last resort we should search for `require("@/` and `from "@/` with a regex after the transform is over, so that we'll at least know if something went wrong with it.

Obviously, you can use this script in reverse to avoid pain with rebasing, just be careful and review the result prior to committing. There are parts of the code like map test generation etc. Maybe we even _need_ this script, because there are some open large PRs, and it would be a shame for everyone to do this manually while rebasing.

verytactical:
Please separate it into at least 2 commits
- changes to configs + manual changes to `@/` whenever it can't be scripted
- scripted changes

anton-trunov:
https://github.com/tact-lang/tact/pull/1476#issuecomment-2608213019

I'm having hard time understanding this comment, so it makes me think we should just close this PR and the corresponding issue. Our build system is already complex enough. The benefits of having absolute imports do not seem to justify the complexity of solving this issue.

PR #1478, opened 2025-01-21
fix: support use of constant in other constant and for struct field default value before declaration

### LINKED ISSUE #879: Constants in default struct fields are not accessible

### LINKED ISSUE #1477: Not possible to use a constant as the value of another constant if the first is declared before the second
This code does not compile:
```tact
const A: Int = B;
const B: Int = 10;
```

With this error:
```
Error: sources/contract.tact:22:16: Cannot evaluate expression to a constant: cannot evaluate declared constant "B" as it does not have a body
Line 22, col 16:
  21 | 
> 22 | const A: Int = B;
                      ^
  23 | const B: Int = 10;
```

We are not in the 70s to depend on the order of declarations 😄 

See also #879

i582:
@jeshecdom I don't know any of the nuances of your vision for this interpreter, so maybe this PR is complete nonsense and it should be done other way. I'd be glad to hear any suggestions!

The main idea of ​​this PR is that we recursively (on demand) calculate the value of a constant without depending on the order. The most important part is to handle circular dependencies correctly, otherwise the compiler will crash with an exception due to the depth of the recursion.

## Issue

Closes #1477.
Closes #879.

## Checklist


src/optimizer/interpreter.ts
```diff
@@ -711,6 +711,17 @@ that binds a variable name to its corresponding value.
 export class Interpreter {
     private envStack: EnvironmentStack;
     private context: CompilerContext;
+
+    /**
+     * Stores all visited constants during the current computation.
+     */
+    private visited: Set<string> = new Set();
```
jeshecdom: Probably change the name to `visitedConstants` just to ensure that these fields are related to constant computation.

src/optimizer/interpreter.ts
```diff
@@ -711,6 +711,17 @@ that binds a variable name to its corresponding value.
...
+
+    /**
+     * Stores all constants that were calculated during the calculation of some constant.
+     * Used only in case of circular dependencies to return a clear error.
+     */
+    private dependenciesPath: string[] = [];
```
jeshecdom: Also, maybe rename it to `constantDependencies`.

src/types/stmts/constant-as-default-value-of-struct-field.tact
```diff
@@ -0,0 +1,7 @@
+primitive Int;
+
+const C: Int = 100;
+
+struct Foo {
```
jeshecdom, line 5: What happens if the struct is declared before the constant `C`?

jeshecdom (commented):
It is a nice solution, I actually like it. It is treating the constants as inlining their definition, as long as they have not been evaluated already. Since evaluation is on demand, it handles cases like this:
```
const A: Int = B;
const B: Int = foo(1);

function foo(v: Int) {
   if (v == 1) {
      return 0;
   } else {
      return A;   // The else never executes, so no circular dependence at compile-time
   }
}
```
I was wondering if we could do this process for contract constants, but then I realized that the solution already evaluates contract constants. Note however that contract constants cannot depend on each other, because `self` is not allowed in their initializers and only calls to global functions are allowed (which cannot reference the contract's constants). Nevertheless, the solution already handles any dependencies contract constants may have with global constants, which is nice.

Just a couple of minimal comments below. Suggestions from other team members are welcomed.

anton-trunov:
> Note however that contract constants cannot depend on each other, because self is not allowed in their initializers

@jeshecdom Why is `self` not allowed?

src/optimizer/interpreter.ts
anton-trunov: What about cases like this?

```tact
const Foo: Int = 42;

contract C {
  const Foo: Int = 43;
}
```

Are these handled correctly with `visited`?

src/optimizer/interpreter.ts
i582: Let's call it `constantComputationPath` since it actually contains function calls as well

src/optimizer/interpreter.ts
i582: Done

src/types/stmts/constant-as-default-value-of-struct-field.tact
i582: Works fine

i582:
> > Note however that contract constants cannot depend on each other, because self is not allowed in their initializers
> 
> Why is `self` not allowed?

Looks like this code is not working:

```tact
contract Master {
    const C: Int = 10;
    const D: Int = self.C;

    get fun foo(a: Int): Int {
        if (a == D) {
            return 10;
        }
        return 0;
    }
}
```

With this error:
```
test/main.tact:6:20: Unable to resolve id 'self'
  5 |     const C: Int = 10;
> 6 |     const D: Int = self.C;
                         ^~~~
  7 | 
```

So we need to support this case in the resolver first in order to work with it in the interpreter
#1481

jeshecdom:
> So we need to support this case in the resolver first in order to work with it in the interpreter
> #1481

We need to be careful though, because as soon as `self` is allowed in constant's initializers, the contract constants could be defined using contract fields, meaning that their value will be dependent on what the Interpreter has in its environment at the moment of interpretation:
```
contract Test {
   f: Int = 5;
   const A: Int = self.foo();
   const B: Int = 10 + self.f;  
}
.........
```
One may think that in the above code `B = 15`, but what if `self.foo()` changes the value of `self.f` to, say `2`?. Then, `B = 12`. Even worse, changing the order of the constant declarations:
```
contract Test {
   f: Int = 5;
   const B: Int = 10 + self.f;  
   const A: Int = self.foo();
}
.........
```
would now give `B = 15`, because `B` would be interpreted first. So, I think that references to fields should be avoided in constant declarations (and probably other expressions involving `self` that I have not thought of).

i582:
I think that a contract constant cannot (and shouldn't) depend on a contract field

jeshecdom (commented):
I approve these changes, unless @anton-trunov and @verytactical have additional comments.

.github/workflows/tact.yml
```diff
@@ -116,102 +112,6 @@ jobs:
         run: |
           yarn link
 
-      - name: CLI Test | Check tact --version
```
verytactical: Something's off with a rebase. These changes are from a different branch.

i582: Looks like this wasn't the latest main branch, fixed

src/types/resolveDescriptors.ts
```diff
@@ -2141,6 +2141,22 @@ function initializeConstants(
     constants: ConstantDescription[],
     ctx: CompilerContext,
     util: AstUtil,
+): CompilerContext {
+    for (const constant of constants) {
+        if (constant.ast.kind === "constant_def") {
+            constant.value ??= evalConstantExpression(
```
verytactical, line 2180: I'm unsure it's even safe to do it here: typechecking is not over, how can we evaluate it?

i582: But this function is called only after `checkConstants`, where we check types

jeshecdom: Mmmm.... @verytactical has a good point. What if the constant initializer is calling a function that has not been typechecked yet? Or is there a guarantee that before constant evaluation all functions have been typechecked?

i582: > Mmmm.... @verytactical has a good point. What if the constant initializer is calling a function that has not been typechecked yet? Or is there a guarantee that before constant evaluation all functions have been typechecked?

Looks like `ctx = initializeConstantsAndDefaultContractAndStructFields(ctx, util);` is the last step in type checking, so this is not the case

jeshecdom: It is correct that it is the last step in `resolveDescriptors`. But then I checked `precompile.ts` and noticed that `resolveDescriptors` is done before `resolveStatements`, which means that functions are actually not type-checked before evaluating constants. 

This explains why the tests did not complain about assigning an integer to `self` (which has the type of a contract), like `self = 100`. I was wandering about that when reading the tests.

If we want to keep this behavior, then the interpreter should not rely on the type-checker and carry out type checks while it interprets the code (we will need to extensively test the interpreter).  The other option is to move constant evaluation as the last step in `precompile.ts`.

@anton-trunov, we need your opinion on this, because I do not know if some process after `resolveDescriptors` in `precompile.ts` requires constants to be already evaluated.

anton-trunov: const eval should be done after typechecking, it's just historically it wasn't the case :)

src/types/stmts/const-eval-almost-constant-circular-dependency.tact
anton-trunov, line 1: const eval tests should not go to the typechecker tests

anton-trunov (changes requested):

src/test/compilation-failed/const-eval-failed.spec.ts
anton-trunov, line 1: let's add a test with nested function calls `foo() -> bar() -> ...`

anton-trunov: Also, let's add a test like `const A: Int = A`

src/optimizer/interpreter.ts
```diff
@@ -870,18 +882,42 @@ export class Interpreter {
...
                 );
             }
+            this.visitedConstants.add(name);
+
+            if (constant.ast.kind === "constant_def") {
+                this.constantComputationPath.push(name);
+                constant.value = this.interpretExpression(
+                    constant.ast.initializer,
+                );
+                this.constantComputationPath.pop();
```
anton-trunov: Since these `push` / `pop` operations need to be always balanced, let's define a helper that ensures that

i582: Done

src/test/compilation-failed/const-eval-failed.spec.ts
i582: Done

src/types/stmts/const-eval-almost-constant-circular-dependency.tact
i582: Done

src/types/resolveDescriptors.ts
i582: > const eval should be done after typechecking, it's just historically it wasn't the case :)

Can we separate this into another issue and do it separately?

anton-trunov: @i582 definitely!

src/test/compilation-failed/const-eval-failed.spec.ts
```diff
@@ -193,4 +193,35 @@ describe("fail-const-eval", () => {
...
+    });
+    itShouldNotCompile({
+        testName: "const-eval-constant-circular-dependency-with-function",
+        errorMessage:
+            "Cannot evaluate expression to a constant: cannot evaluate C as it has circular dependencies: [foo() -> C -> A -> foo() -> C]",
+    });
+    itShouldNotCompile({
+        testName: "const-eval-constant-circular-dependency-with-functions",
+        errorMessage:
+            "Cannot evaluate expression to a constant: cannot evaluate C as it has circular dependencies: [foo() -> bar() -> baz() -> C -> A -> foo() -> bar() -> baz() -> C]",
```
anton-trunov: As a user, I would expect to see an error message like this:

```
cannot evaluate C as it has circular dependencies: C -> A -> foo() -> bar() -> baz() -> C
```

so it starts with C and ends with C as soon as possible.

anton-trunov (changes requested):

src/test/compilation-failed/const-eval-failed.spec.ts
i582: Fixed

src/test/compilation-failed/const-eval-failed.spec.ts
```diff
@@ -193,4 +193,35 @@ describe("fail-const-eval", () => {
...
+    itShouldNotCompile({
+        testName: "const-eval-constant-circular-dependency-with-functions",
+        errorMessage:
+            "Cannot evaluate expression to a constant: cannot evaluate C as it has circular dependencies: [C -> A -> foo() -> bar() -> baz() -> C]",
+    });
+    itShouldNotCompile({
+        testName:
+            "const-eval-constant-circular-dependency-with-recursive-function",
+        errorMessage:
+            "Cannot evaluate expression to a constant: cannot evaluate C as it has circular dependencies: [C -> A -> foo() -> foo() -> foo() -> C]",
```
anton-trunov, line 220: looks like this error message can be very long for deep enough recursion

i582: Fixed

```
[C -> A -> foo() -> foo() -> foo() -> ... -> foo() -> foo() -> foo() -> foo() -> C
```

src/test/compilation-failed/contracts/const-eval-constant-circular-dependency-with-recursive-function.tact
```diff
@@ -0,0 +1,15 @@
+const A: Int = foo(20);
```
anton-trunov: let's have both error messages for `foo(3)` and `foo(20)`

i582: Done

anton-trunov (approved):

jeshecdom (approved):

PR #1482, opened 2025-01-22
fix: forbid read and write to `self` in contract `init` function

### LINKED ISSUE #816: `self` should not be assignable in `init` method

### Tact source code

```tact
contract Test {
   a: Int;

   init() {
      let contract_copy = self;
      self.a = 5;
      self = contract_copy;    // Does this line "uninitialize" self.a?
   }
}
```

```shell
This is the generated func code of the init method:

(int) $Test$_contract_init() impure inline_ref {
    var (($self'a)) = (null());
    var ($contract_copy'a) = ($self'a);
    $self'a = 5;
    ($self'a) = ($contract_copy'a);
    return ($self'a);
}
```

### What happened?

Field `$self'a` will be initialized to `null()` after `$Test$_contract_init` executes. Is `null()` a valid value for an `Int`?

### What did you expect?

Probably an error message explaining that `self` should not be assigned in the `init` method.

### Steps to reproduce

_No response_

### How do you run Tact?

Tact CLI

### Anything else?

_No response_

i582:
## Issue

Closes #816.

## Checklist


src/types/resolveErrors.ts
```diff
@@ -29,10 +29,10 @@ function resolveStringsInAST(
     ctx: CompilerContext,
     util: AstUtil,
 ) {
-    traverse(ast, (node) => {
+    traverse(ast, (node): boolean => {
         if (node.kind === "static_call" && isRequire(node.function)) {
             if (node.args.length !== 2) {
-                return;
+                return true;
```
verytactical: It would be better to get more than 1 compiler error per compiler run.

Please don't do early exit in `traverse`.

i582: `true` here means continue traversing, so this shouldn't change the behavior 🤔

src/types/test-failed/read-self-in-init-function-assign.tact
```diff
@@ -0,0 +1,11 @@
+primitive Int;
+trait BaseTrait {}
+
+contract Contract {
+    a: Int;
+
+    init() {
+        let a = 100;
+        self.a = self;
```
jeshecdom, line 9: What would happen if the assignment to `self` occurs after an assignment to another variable in, say, the true case of an if? Like:
```
contract Contract {
    init() {
        let a = 100;
        if (a == 100) {
            a = 30;
            self = 20;
        }
    }
}
```
I noticed that `checkNode` returns false after processing an assignment statement like `a = 30`. Hence, `traversal` on node `a = 30` would return false. Therefore, when the control flow is returned to the statements in the true case of the conditional (i.e., case `statement_condition`, line 187 in `traversal` function), the `for` loop will get interrupted and will not process the next statement `self = 20`.

Something similar would happen for expressions of the form `a.f + self` or `a.foo() + self` which would not check `self`, since `checkNode` returns false in `a.f` and `a.foo()`.

But I may be wrong.

src/types/test-failed/init-self-reassign.tact
```diff
@@ -0,0 +1,7 @@
+trait BaseTrait {}
+
+contract Contract {
+    init() {
+        self = 100;
```
jeshecdom, line 5: What happens if we use an augmented assign here `self += 100`? I noticed that `checkNode` in `resolveDescriptors` does not have a case for augmented assigns. Would it enter the `kind = "id"` case once the augmented assign gets traversed?

src/ast/iterators.ts
```diff
@@ -85,203 +98,222 @@ export function traverse(node: AstNode, callback: (node: AstNode) => void) {
...
+            if (!traverse(node.type, callback)) return true;
             break;
         case "struct_decl":
         case "message_decl":
-            traverse(node.name, callback);
-            if (node.kind === "message_decl" && node.opcode !== null) {
-                traverse(node.opcode, callback);
+            if (!traverse(node.name, callback)) return true;
+            if (
+                node.kind === "message_decl" &&
```
jeshecdom: Is there a reason for this check of `kind` here if the switch case already is `message_decl`?

src/types/test-failed/read-self-in-init-function-assign.tact
i582: Nice catch! We don't actually need to return the result from `traverse` 🤔

i582: Fixed

src/types/test-failed/init-self-reassign.tact
i582: Fixed

src/ast/iterators.ts
i582: Fixed

src/types/resolveErrors.ts
verytactical: If so, existing calls to `traverse` would return `undefined`, and _not_ continue their execution.

verytactical: Anyway, I still don't understand how this works. Can we avoid changing `traverse`?

jeshecdom: @verytactical I think I get why `traverse` needs to stop traversing the children of a node when `checkNode` (the callback function) returns false. Let me explain, and I apologize for the long explanation. 

Observe the first conditional in `checkNode`:
```
if (node.kind === "field_access" || node.kind === "method_call") {
    return false;
```
If we remove the `return false`, then a valid expression like `self.a`, would later visit its child node `self` and report an error, which is an incorrect outcome. 

Similarly, in an assignment statement:
```
if (
    node.kind === "statement_assign" ||
    node.kind === "statement_augmentedassign"
   ) {
          const left = node.path;
          if (left.kind === "id" && left.text === "self") {
                throwCompilationError(
                        "cannot reassign `self` in `init` function",
                        left.loc,
                    );
           }

           traverse(node.expression, checkNode);    // Label: A

           return false;                            // Label: B
     }
```
If we remove the `return false` at label `B`, a valid statement like `self.a = foo()` would later visit the child node `self` on the expression to the left of = and report an error, which is an incorrect outcome. When the expression to the left of = is valid (i.e., when it is not explicitly the `self` identifier), we should only traverse the expression to the right of = (otherwise we risk an invalid error on the left of =), which is the reason for the call `traverse(node.expression, checkNode)` at label `A` and then `return false` to avoid revisiting the children of the assign statement.

Note that the only change to `traverse` is the signature of the `callback` function to `AstNode => boolean` and the early return after checking the callback at line 10 in `src/ast/iterators.ts`.
If such modification is not acceptable, is it possible to have two `traverse`, one as in this PR (call it `traverseCheck`) and another with callback `AstNode => void` implemented as:
```
function traverse(node: AstNode, callback: (node: AstNode) => void) {
    // Returning true in callback ensures no early return in the traverse
    traverseCheck(node, (n) => {callback(n); return true;});
}  
```
Would this be acceptable to you @verytactical? Or maybe you have a suggestion on how to change the callback `checkNode`.

i582: @verytactical, @jeshecdom is right, we need it to not traverse some nodes, like `self` in `self.a`

i582: > If so, existing calls to `traverse` would return `undefined`, and _not_ continue their execution.

I checked all the uses of `traverse` and changed callbacks to return `true` to continue traversing

src/ast/iterators.ts
```diff
@@ -6,8 +6,10 @@ import { AstNode } from "./ast";
  * @param node The node to traverse.
  * @param callback The callback function to apply to each AST element.
  */
-export function traverse(node: AstNode, callback: (node: AstNode) => void) {
-    callback(node);
+export function traverse(node: AstNode, callback: (node: AstNode) => boolean) {
```
jeshecdom: In order to keep the original `traverse` function, change the name of this function to `traverseAndCheck`. But add the new function:
```
function traverse(node: AstNode, callback: (node: AstNode) => void) {
    // Returning true in callback ensures no early return in the traverse
    traverseAndCheck(node, (n) => {callback(n); return true;});
}  
```

src/types/resolveDescriptors.ts
```diff
@@ -1859,7 +1896,7 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
 
     for (const [k, t] of types) {
         const dependsOn: Set<string> = new Set();
-        const handler = (src: A.AstNode) => {
+        const handler = (src: A.AstNode): boolean => {
```
jeshecdom: Change this back to return `void` instead of `boolean`.

src/types/resolveErrors.ts
```diff
@@ -30,10 +30,10 @@ function resolveStringsInAST(
     ctx: CompilerContext,
     util: AstUtil,
 ) {
-    traverse(ast, (node) => {
+    traverse(ast, (node): boolean => {
```
jeshecdom: Change this back to return `void` instead of `boolean`.

src/types/resolveDescriptors.ts
```diff
@@ -1019,6 +1019,43 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
...
+                throwCompilationError(
+                    "cannot read whole `self` in `init` function",
+                    node.loc,
+                );
+            }
+            return true;
+        }
+
+        ast.statements.forEach((stmt) => {
+            traverse(stmt, checkNode);
```
jeshecdom: Change this call so that it calls `traverseAndCheck`.

jeshecdom (commented):
The main unresolved issue here is if we should change the signature of the `traverse` function. I say, we keep `traverse` as it was before, but add a new `traverseAndCheck` that receives a `callback` that returns `boolean`. Comments below.

Let us do this and merge this PR.

src/ast/iterators.ts
i582: Done

src/types/resolveDescriptors.ts
i582: Done

src/types/resolveErrors.ts
i582: Done

src/types/resolveDescriptors.ts
i582: Done

jeshecdom (approved):

PR #1503, opened 2025-01-22
example: relativify imports

verytactical:


requires.build.ts
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+};
+
+const transformLiteral = (fullPath: string, importLiteral: t.StringLiteral) => {
+    const importedName = importLiteral.value;
+    if (!importedName.startsWith("@/")) {
+        return false;
+    }
+    const result = relative(
```
verytactical, line 69: Thing to thing about: importing `foo` might mean `foo/index`. Does it matter here?

requires.build.ts
```diff
@@ -0,0 +1,77 @@
+import { glob, readFile, writeFile } from "fs/promises";
+import { parse, ParserOptions } from "@babel/parser";
+import generate from "@babel/generator";
+import { dirname, extname, join, relative, resolve } from "path";
+import * as t from "@babel/types";
+
+const main = async () => {
+    for await (const file of glob("./dist/**/*.{js,ts}")) {
+        const fullPath = join(__dirname, file);
```
verytactical, line 9: Working with paths has to be checked on windows.

requires.build.ts
```diff
@@ -0,0 +1,77 @@
...
+                if (!t.isCallExpression(init)) {
+                    continue;
+                }
+                const callee = init.callee;
+                if (!t.isIdentifier(callee) || callee.name !== "require") {
+                    continue;
+                }
+                const importLiteral = init.arguments[0];
+                if (init.arguments.length !== 1 || !importLiteral) {
+                    return;
```
verytactical, line 38: This should be a `continue`. Oopsie.

requires.build.ts
```diff
@@ -0,0 +1,77 @@
...
+            } else if (t.isImportDeclaration(node)) {
+                if (transformLiteral(fullPath, node.source)) {
+                    wasChanged = true;
+                }
+            }
+        }
+        if (wasChanged) {
+            await writeFile(
+                fullPath,
+                generate(ast, {
```
verytactical, line 55: What exactly did we lose except for comments? Should check diff.

requires.build.ts
```diff
@@ -0,0 +1,77 @@
...
+import generate from "@babel/generator";
+import { dirname, extname, join, relative, resolve } from "path";
+import * as t from "@babel/types";
+
+const main = async () => {
+    for await (const file of glob("./dist/**/*.{js,ts}")) {
+        const fullPath = join(__dirname, file);
+        const source = await readFile(fullPath, "utf-8");
+        const options: ParserOptions =
+            extname(file) === ".ts"
```
verytactical, line 12: Are there imports in other files in `dist/` that are neither js nor ts?

requires.build.ts
verytactical: Also from here on `continue`s should be `throw new Error()`s, so that we'll know if there are any weird things happening when they happen.

requires.build.ts
verytactical: Also search for `@/` in transformed file, and throw an error if it was found.

requires.build.ts
```diff
@@ -0,0 +1,77 @@
+import { glob, readFile, writeFile } from "fs/promises";
+import { parse, ParserOptions } from "@babel/parser";
+import generate from "@babel/generator";
+import { dirname, extname, join, relative, resolve } from "path";
+import * as t from "@babel/types";
+
+const main = async () => {
+    for await (const file of glob("./dist/**/*.{js,ts}")) {
```
verytactical, line 8: Async glob from `"glob"` is _really nasty_, and I used Node22's glob. We have a CLI that checks everything works on older node, and it will fail. The point of that CI was to check that the compiler we produce doesn't use these features, not our build scripts, so it should probably be fixed.

requires.build.ts
novusnota: PowerShell works fine with `/`

.eslintignore
```diff
@@ -12,3 +12,4 @@ jest.setup.js
 jest.teardown.js
 /docs
 version.build.ts
+requires.build.ts
```
verytactical, line 15: Obviously we shouldn't disable linter on it.

requires.build.ts
verytactical: `relative` doesn't.

anton-trunov:
what issue does this pr solve? looks like it should either be a draft pr or just a branch with a link in the team chat

Issue #1215, opened 2024-12-19
Remove extraneous files from npm package


anton-trunov:
`dist/func/funcfiftlib.*` are expected to be large.

* [x] Remove `.npmignore` and `.yarnignore`, because we use `files` in package.json
* [x] Remove `dist/tsconfig.tsbuildinfo` from package
* [x] Remove src/test from package (edit `files` in package.json to remove)
* [x] Remove `*.build.js` and related build scripts
* [x] Do not pack `funcfiftlib.wasm` (the generated `funcfiftlib.wasm.js` already contains it)
* [x] Add a script to `package.json` to show the top 10 largest files in the packed archive (for debugging purposes)

Run `yarn gen && yarn build && npm pack`, see if there are any other leftovers

anton-trunov:
The size of Tact’s NPM package:
- before: 29.4 MB
- after all the fixes: 8.8 MB 

More than 3x improvement.

PR #1510, opened 2025-01-23
chore: don't add `dist/**/output` to package

i582:
## Issue

#1215.

## Checklist


package.json
```diff
@@ -46,6 +46,7 @@
     "bin/**/*",
     "!src/**/*",
     "!**/test",
+    "!**/benchmarks",
```
anton-trunov: aren't these already ignored by `"!**/test"` above? benchmarks should only live inside tests/

i582: I run `npm pack` and get this without those changes: 🤔 
Screenshot 2025-01-23 at 13 35 50

anton-trunov: you should remove `./src/benchmarks` locally, it's some leftovers most likely

i582: I thought so too, but I checked without `src/benchmarks` and got the same result

i582: Oops, I need to delete the `dist/` folder

package.json
```diff
@@ -46,6 +46,7 @@
     "bin/**/*",
     "!src/**/*",
     "!**/test",
+    "!dist/**/output/**/*",
```
anton-trunov:
```suggestion
    "!dist/**/output",
```

anton-trunov (approved):

Issue #1255, opened 2024-12-25
Attributes don't have locations


verytactical:
Function and constant attributes are represented as strings in AST. In order to emit error that involves an attribute, there is no way to gets its location.

Attributes should be represented as objects.

```
type AstFooAttribute = {
    kind: 'foo';
    loc: SrcInfo;
}
```

i582:
Looks like we already have locations for attributes 🤔 

https://github.com/tact-lang/tact/blob/a0c838122c3cdabba50d5b3bbbc9030245966cbb/src/ast/ast.ts#L720-L752

Commit (28.12): https://github.com/tact-lang/tact/commit/da4b8d82128cf4b6f9b04d93a93a9382407112c2

PR #1524, opened 2025-01-23
feat: underline only the attribute itself in compilation errors about attributes

i582:
Not sure if this is a good way to pass location there when we have only `SrcInfo`

## Issue

#1255.

## Checklist

### DIFF
--- a/src/grammar/next/index.ts
+++ b/src/grammar/next/index.ts
@@ -950,7 +950,7 @@ const parseConstantDefInModule =
                 $: "range",
                 start: firstAttribute.loc.interval.startIdx,
                 end: firstAttribute.loc.interval.endIdx,
-            } as $.Loc);
+            });
             result.attributes = [];
         }
         return result;
--- a/src/grammar/next/index.ts
+++ b/src/grammar/next/index.ts
@@ -946,9 +946,11 @@ const parseConstantDefInModule =
         const result = parseConstantDef(node)(ctx);
         const firstAttribute = result.attributes[0];
         if (typeof firstAttribute !== "undefined") {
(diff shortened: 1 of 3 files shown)


src/grammar/next/index.ts
```diff
@@ -946,9 +946,11 @@ const parseConstantDefInModule =
...
         const firstAttribute = result.attributes[0];
         if (typeof firstAttribute !== "undefined") {
-            // FIXME: should be `firstAttribute.loc`
-            // https://github.com/tact-lang/tact/issues/1255
-            ctx.err.topLevelConstantWithAttribute()(node.loc);
+            ctx.err.topLevelConstantWithAttribute()({
+                $: "range",
+                start: firstAttribute.loc.interval.startIdx,
+                end: firstAttribute.loc.interval.endIdx,
+            } as $.Loc);
```
verytactical: Casts are not allowed by styleguide.

verytactical (changes requested):

src/grammar/next/index.ts
i582: Fixed

src/grammar/next/index.ts
```diff
@@ -946,9 +946,11 @@ const parseConstantDefInModule =
         const result = parseConstantDef(node)(ctx);
         const firstAttribute = result.attributes[0];
         if (typeof firstAttribute !== "undefined") {
-            // FIXME: should be `firstAttribute.loc`
```
verytactical, line 949: Take a look at this comment please :)

verytactical: I'll look further into this issue after I'm finished with the current one. It's odd that we have to create a `range` here.

PR #1545, opened 2025-01-23
ci: setup Node.js 23 forward compatibility testing

### LINKED ISSUE #1217: Add Node.js 23 to CI
```shell
$ yarn
yarn install v1.22.22
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
error release-it@17.10.0: The engine "node" is incompatible with this module. Expected version "^18.18.0  ^20.9.0  ^22.0.0". Got "23.1.0"
error Found incompatible module.
```

Reported by @Shvandre: he mentions it used to work until recently

novusnota:
Basic testing, which involves installation of dependencies and running some of the `gen:` scripts that do not generate contracts — I think this is enough to check that the codebase itself is compatible with Node.js 23.

## Issue

Closes #1217.

## Checklist

### DIFF
--- a/.github/workflows/tact.yml
+++ b/.github/workflows/tact.yml
@@ -62,13 +62,12 @@ jobs:
 
       - name: Forward compatibility tests
         run: |
-          # Install dependencies and perform some steps of the build pipeline
+          # Install dependencies and perform some steps of the build pipeline,
+          # skipping the contract generation and the final `yarn build` step
           yarn install
           yarn gen:grammar
           yarn gen:stdlib
           yarn gen:func-js
-          # ...skipping contract generation...
-          yarn build
           # Clean-up
           yarn cleanall
 
--- a/.github/workflows/tact.yml
+++ b/.github/workflows/tact.yml
@@ -45,6 +45,34 @@ jobs:
           yarn cleanall
           yarn config delete ignore-engines
 
+      - name: Setup Node.js 23 for forward compat tests
+        uses: actions/setup-node@v4
+        with:
+          node-version: 23
+          # without caching
+
+      - name: Forward compat tests
+        run: |
+          # Install dependencies, gen and build the compiler
+          yarn install
+          yarn clean
+          yarn gen
+          yarn build
+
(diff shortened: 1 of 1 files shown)


.github/workflows/tact.yml
```diff
@@ -45,6 +45,34 @@ jobs:
...
+        with:
+          node-version: 23
+          # without caching
+
+      - name: Forward compat tests
+        run: |
+          # Install dependencies, gen and build the compiler
+          yarn install
+          yarn clean
+          yarn gen
```
verytactical: This will increase runtime of our CI by 10 minutes.

novusnota: Would `install` be enough then?

UPD: Probably, but just to make sure I've added a couple of more steps that do not generate contracts

.github/workflows/tact.yml
```diff
@@ -45,6 +45,34 @@ jobs:
           yarn cleanall
           yarn config delete ignore-engines
 
+      - name: Setup Node.js 23 for forward compat tests
+        uses: actions/setup-node@v4
+        with:
+          node-version: 23
+          # without caching
+
+      - name: Forward compat tests
```
verytactical: Neither backwards not forwards compat tests should go in pipeline before the main build. This increases the chance we'll have wrong environment there.

novusnota: Great suggestion, I'll separate those

.github/workflows/tact.yml
```diff
@@ -45,6 +45,34 @@ jobs:
           yarn cleanall
           yarn config delete ignore-engines
 
+      - name: Setup Node.js 23 for forward compat tests
+        uses: actions/setup-node@v4
+        with:
+          node-version: 23
```
verytactical: We're checking here that both our build and package work with Node 23.

We shouldn't put that requirement on our build.

novusnota:
image

Now we're good with timings

verytactical (approved):
Looks awesome!

PR #1563, opened 2025-01-24
put the linter launch in a separate job

### LINKED ISSUE #1575: Move linters into a separate job in CI
This will allow detecting linters failing earlier

Danil42Russia:
Closes #1575.

Made the `tact` assembly split and run all the linters.
This will help to find problems with the code in advance, even before the compiler is built. Also, it is not necessary to run them on all platforms.

Yes, there will be errors now, if this idea is directed to you, I will add running the generator to the CI step for linters.

## Checklist

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -28,7 +28,10 @@ jobs:
 
       - name: Prepare linter
         run: |
-          yarn gen
+          gen:grammar
+          gen:stdlib
+          gen:func-js
+          gen:contracts:test
 
       - name: Check there are no errors to be fixed in package.json and no uncommitted changes
         run: |
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -28,9 +28,7 @@ jobs:
 
       - name: Prepare linter
         run: |
-          yarn gen:grammar
-          yarn gen:stdlib
-          yarn gen:func-js
+          yarn gen
 
       - name: Check there are no errors to be fixed in package.json and no uncommitted changes
         run: |
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -45,3 +45,11 @@ jobs:
       - name: Check there are no unused dependencies
         run: |
           yarn knip
+
+      - name: Check broken file references in internal documentation
+        uses: lycheeverse/lychee-action@v2
+        with:
+          args: "-n -s file --exclude-path node_modules --exclude-path docs './**/*.md'"
+          output: "/dev/stdout"
+          fail: true
+          failIfEmpty: false
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,47 @@
+name: Linter check
+
+env:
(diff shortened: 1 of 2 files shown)


anton-trunov:
Looks great, let's just fix the linter check and merge

.github/workflows/linter.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Setup Node.js 22
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: "yarn"
+
+      - name: Install dependencies
+        run: |
+          corepack enable
+          yarn install
```
anton-trunov, line 27: We don't want to install **all** dependencies here, but only the dev deps needed to perform the checks:
- cspell,
- eslint,
- knip,
- prettier

.github/workflows/linter.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: |
+          yarn fmt:check
+
+      - name: Spellcheck code base
+        run: |
+          yarn spell
+
+      - name: Check there are no unused dependencies
+        run: |
+          yarn knip
```
anton-trunov, line 54: Let's run all these steps (even if some previous fail) and report CI failure if any of those fails.

anton-trunov (changes requested):

.github/workflows/linter.yml
Danil42Russia: Good point, but separately they are best not put up, for a reason:
1) You will have to install not only these packages, but also additional packages: `typescript-eslint`;
2) It will be necessary to keep track of versions additionally;
3) To make CI work correctly, you need to run codegen (and this requires a package);
4) If you make a common cache for `node_modules` (which you should do), the installation will take seconds.

.github/workflows/linter.yml
Danil42Russia: I think how to do it better, there are two realizations:
1) Through `matrix`, but then it will be necessary to put dependencies on each task. This is not optimized;
2) Add for each step `continue-on-error: true`, then the failed steps will not be shown, it can be considered that it is equal to `|| true`.

anton-trunov (changes requested):
Weird CI failure.

I would also move the broken links checker here:
```
      - name: Check broken file references in internal documentation
        if: runner.os == 'Linux'
        uses: lycheeverse/lychee-action@v2
        with:
          args: "-n -s file --exclude-path node_modules --exclude-path docs './**/*.md'"
          output: "/dev/stdout"
          fail: true
          failIfEmpty: false
```

anton-trunov:
Looks like we can save about a minute or two (tops), since we have to execute `yarn gen` :)
Basically, we just skip running tests, but build the compiler and run `yarn gen` twice in CI (in parallel)

Danil42Russia:
> since we have to execute `yarn gen`

I looked up running `gen:contracts:test:map` is not necessary for linter to work correctly

verytactical (approved):
LGTM

anton-trunov (approved):

PR #1565, opened 2025-01-24
remove the `crc-32` library from dependency

### LINKED ISSUE #1437: Remove `crc-32` dependency
Write tests with a contract that uses `crc32` function that pass strings with unicode characters.

Copy source code of `crc-32` library to `utils/`. If there are tests in that library, copy them too.

Here's [another good implementation](https://github.com/ton-blockchain/tolk-vscode/blob/master/server/src/utils/crc32.ts) :)

Danil42Russia:
## Issue

Closes #1437.

The table was generated in advance, via:
```javascript
function main() {
  const POLYNOMIAL = -306674912;
  for (let i = 0; i < 256; i++) {
    let r = i;
    for (let bit = 8; bit > 0; --bit) {
      r = ((r & 1) ? ((r >>> 1) ^ POLYNOMIAL) : (r >>> 1));
    }

    console.log(("0x" + (r >>> 0).toString(16)).padStart(8, "0"))
  }
}
```

If you need to write tests, I'll add them

## Checklist


src/utils/crc32.ts
```diff
@@ -0,0 +1,58 @@
+const TABLE = new Int32Array([
```
verytactical: Please don't add generated code here.

verytactical (changes requested):
The original issue started specifically with

> Write tests with a contract that uses crc32 function that pass strings with unicode characters.

Until there are tests for unicode characters, or rather any crc32 tests at all, we can't even know that hash is computed the way it did before, and can't do any of these changes.

src/utils/crc32.ts
Danil42Russia: Why?
1. it is faster than creating a table every time;
2. it is done by analogy with `crc16`, which has the same table.

verytactical:
Imagine worst-case scenario: we have something like `message(crc32())` in tact contract, gets evaluated at compile-time as constant expression, all the contract tests pass because bindings use precisely the same value, then incorrect contract is deployed.

src/utils/crc32.ts
verytactical: Which of the two options sounds more correct? :)
- execute 2048 operations every time compiler starts
- have a table of questionable provenance that is impossible to review, that also takes 65536 operations every time TS checks types, and also increases size of js bundle

src/utils/crc32.ts
```diff
@@ -0,0 +1,58 @@
...
+    0x616bffd3, 0x166ccf45, 0xa00ae278, 0xd70dd2ee, 0x4e048354, 0x3903b3c2,
+    0xa7672661, 0xd06016f7, 0x4969474d, 0x3e6e77db, 0xaed16a4a, 0xd9d65adc,
+    0x40df0b66, 0x37d83bf0, 0xa9bcae53, 0xdebb9ec5, 0x47b2cf7f, 0x30b5ffe9,
+    0xbdbdf21c, 0xcabac28a, 0x53b39330, 0x24b4a3a6, 0xbad03605, 0xcdd70693,
+    0x54de5729, 0x23d967bf, 0xb3667a2e, 0xc4614ab8, 0x5d681b02, 0x2a6f2b94,
+    0xb40bbe37, 0xc30c8ea1, 0x5a05df1b, 0x2d02ef8d,
+]);
+
+export function crc32(data: string | Buffer): number {
+    if (!(data instanceof Buffer)) {
```
verytactical: If `src/optimizer/interpreter.ts` is the only place we call it from, we never pass `Buffer` here.

`Buffer` is Node's non-standard API for binary data, while there is platform-agnostic `Blob`. When we build browser version of `@tact-lang/compiler`, we have to include a polyfill for `Buffer`.

While there are other places that use `Buffer`, we're working hard to remove them. I'd consider removing `Buffer` support here too.

src/utils/crc32.ts
Danil42Russia: So `typescript` and type checking is called only once, when building a project. And if you don't calculate the table, you will have to do it every time you start the compiler.

Regarding the “weird” table:
1) This table is available in the LLVM project: https://github.com/llvm/llvm-project/blob/6383a12e3b4339fa4743bb97da4d51dea6d2e2ea/llvm/lib/Support/CRC.cpp#L30
2) There are many references in Microsoft repository https://github.com/search?q=org%3Amicrosoft%20%220xD9D65ADC%22&type=code

If you look at the code of the `js-crc32` library, it doesn't get any better, it doesn't even have comments https://github.com/SheetJS/js-crc32/blob/master/crc32c.js#L85

Danil42Russia: To make sure the table is correct, you can copy from anywhere else, diff will be zero.
I can also add its generator to the project.

verytactical: I realize there are quite a few mentions of this table online, because it's used for something like last half of a century, but I don't think checking numbers by hand is a sound approach to development.

> type checking is called only once

There's also `tsserver`.

src/utils/crc32.ts
Danil42Russia: > while there is platform-agnostic `Blob`

Can you elaborate, I didn't see any use of `Blob` in the code?

verytactical: There isn't any use of `Buffer` either. If we want to compute crc32 of binary data for some reason, let's make this function to work on `string | Blob` instead.

anton-trunov (changes requested):
We are looking for ways to make our code verifiable, so no tables like that should be hardcoded

Danil42Russia:
> so no tables like that should be hardcoded

Can you then create a issue to remove the table in `crc16`?

anton-trunov:
#1635

src/utils/crc32.ts
Danil42Russia: removed the `Buffer`

src/utils/crc32.ts
Danil42Russia: remake

src/utils/crc32.ts.spec.ts
anton-trunov, line 1: `src/utils/crc32.ts.spec.ts` -> `src/utils/crc32.spec.ts`

Danil42Russia: oops, fixed it

src/utils/crc32.ts
```diff
@@ -0,0 +1,27 @@
...
+        c = n;
+        for (let k = 0; k < 8; k++) {
+            c = c & 1 ? (c >>> 1) ^ polynomial : c >>> 1;
+        }
+        table[n] = c;
+    }
+    return table;
+}
+
+const CRC32C_TABLE = makeCRC32Table(0xedb88320);
```
verytactical, line 15:
```
// Reversed polynomial of ISO3309 CRC32

File: src/utils/crc32.ts [hunk 4]
```diff
@@ -0,0 +1,27 @@
+function makeCRC32Table(polynomial: number): number[] {
+    let c: number;
+    const table = [];
```
verytactical: This table has to be an unboxed array as it was before. AFAIK it's more performant.

src/utils/crc32.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const CRC32C_TABLE = makeCRC32Table(0xedb88320);
+
+export function crc32(data: string | Uint8Array): number {
+    if (typeof data === "string") {
+        data = new TextEncoder().encode(data);
+    }
+
+    let crc = 0xffffffff;
+    for (const byte of data) {
```
verytactical, line 23: This way of iterating over an array is (usually) significantly slower.

src/utils/crc32.ts
Danil42Russia: I agree. Added :)

src/utils/crc32.ts
Danil42Russia: I agree, when I wrote a version in `javascript` to compare that everything works identically, I forgot to return the type.

src/utils/crc32.ts
Danil42Russia: The project has the `@typescript-eslint/prefer-for-of` rule enabled, and I have replaced the loop with `for-of`. I can suppress this error.

verytactical (approved):

anton-trunov (approved):

PR #1567, opened 2025-01-25
feat: allow calling methods on `null` when `self` is of an optional type

### LINKED ISSUE #1283: Calling methods on `null` when `self` is of an optional type should be allowed
`null.foo()` should be allowed in the snippet below:

```
extends fun foo(self: Int?): Bool {
    if (self == null) { return false }
    else { return self!! == 42 }
}

contract Test {
    get fun test(): Bool {
        return null.foo();
    }
}
```

Now I get the following compilation error message:

```
Tact compilation failed
Error: test-null.tact:8:16: Invalid type "<null>" for function call
Line 8, col 16:
  7 |     get fun test(): Bool {
> 8 |         return null.foo();
                     ^~~~~~~~~~
  9 |     }
```

Gusarich:
## Issue

Closes #1283.

## Checklist


src/types/resolveExpression.ts
```diff
@@ -703,6 +703,41 @@ function resolveCall(
...
+        // Too many candidates found
+        if (candidates.length > 1) {
+            throwCompilationError(
+                `Ambiguous method call ${idTextErr(exp.method)}`,
+                exp.loc,
+            );
+        }
+
+        // Return the only candidate
+        return registerExpType(ctx, exp, candidates[0]!.f.returns);
```
verytactical:
```
if (candidates.length === 0)
candidates[0]!
```

->

```
const candidate = candidates[0];
if (typeof candidate === 'undefined') { ... }
// now candidate is not falsy
```

Gusarich: @verytactical have a look now

verytactical (dismissed):
Nice code!

anton-trunov (dismissed):

anton-trunov (approved):

Issue #1434, opened 2025-01-20
Move `@ton/crypto` to devDependencies


verytactical:
- `@ton/crypto` uses Node-specific Buffer that has to be polyfilled in browser
- SHA256 implementation comes from a large and slow `jssha` library
- Cannot use SHA256 built into browser, because it's asynchronous
- Other primitives from `@ton/crypto` are used in tests, so it doesn't really matter if we have them as devDependency

Write tests for every use or `sha256_sync` in runtime (`src/abi/global.ts`, `src/optimizer/interpreter.ts`, `src/types/resolveErrors.ts`, `src/types/resolveSignatures.ts`) where it's computed from a string that contains unicode.

Copy SHA256 implementation to `utils/`, for example, from [this library](https://github.com/emn178/js-sha256).

verytactical:
![Image](https://github.com/user-attachments/assets/b183eb22-af7a-45f8-81ea-0da2b46e84ac)

PR #1569, opened 2025-01-25
forbid via `eslint` to use the `Buffer` class in the code

Danil42Russia:
In a comment https://github.com/tact-lang/tact/pull/1565#discussion_r1929544444 @verytactical wrote:
> `Buffer` is Node's non-standard API for binary data, while there is platform-agnostic `Blob`. When we build browser version of `@tact-lang/compiler`, we have to include a polyfill for `Buffer`.
> 
> While there are other places that use `Buffer`, we're working hard to remove them. I'd consider removing `Buffer` support here too.

So my suggestion is to disallow it via an `eslint` rule. This is a concept, if you like it I will add ignore existing errors.

## Checklist

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -15,6 +15,13 @@ module.exports = {
     es2021: true,
   },
   rules: {
+    "no-restricted-syntax": [
+      "error",
+      {
+        selector: "MemberExpression[object.name='Buffer']",
+        message: "lol",
+      },
+    ],
     "@typescript-eslint/no-unused-vars": [
       "error",
       {


anton-trunov (changes requested):
> So my suggestion is to disallow it via an eslint rule. This is a concept, if you like it I will add ignore existing errors.

I'm in favor of disallowing `Buffer`, but we should not just ignore all the existing errors, it needs to be case-by-case.

Also, this issue is relevant here: #1434.

Danil42Russia:
> > So my suggestion is to disallow it via an eslint rule. This is a concept, if you like it I will add ignore existing errors.
> 
> I'm in favor of disallowing `Buffer`, but we should not just ignore all the existing errors, it needs to be case-by-case.
> 
> Also, this issue is relevant here: #1434.

Yes, you're right, but this will prohibit adding new code using `Buffer`.
I'm going to create an issue to replace it, and I'll specify it in the `eslint` error. Well, I must say what to use instead of it.

verytactical:
There are two approaches with adding eslint rules that fail on existing code base
- fix issues immediately, and then ban it; obviously in a change as large as banning `Buffer` it's impossible
- create another issue, leave `// FIXME: #1234` comments right next to `eslint-ignore`s; it's usually used when something is critically wrong (i.e. for banning `as` or `any`)
- just leave it be until other "remove X" issues are resolved

I'd suggest the third approach here. Let's create an issue, link this PR as a reference for how it should eventually be done, and close this for now.

Danil42Russia:
> I'd suggest the third approach here. Let's create an issue, link this PR as a reference for how it should eventually be done, and close this for now.

As far as I'm concerned, the best solution is a combination of the second and third points.
If we do the second one now, it will help us not to add more code using `Buffer`. If we add a replacement hint, it will be even better for development. That said, if we do the third one, we will reference the problem.

But in any case, the choice is yours.

verytactical:
The chance another `Buffer` will get added is pretty small outside of inlining `@ton/core`, `crc32` etc. There is just not much use for it in runtime outside fs calls and cell operations.

Danil42Russia:
> The chance another `Buffer` will get added is pretty small

Then we close the MP?

PR #1596, opened 2025-01-27
fix: very simple constant propagation for obvious cases

### LINKED ISSUE #716: FunC codegen error for obvious cases of zero division in statements

jeshecdom:
## Issue

Closes #716.

This is a simplified version of the AST-based constant propagation algorithm in PR #852 in order to have a quick fix for the issue. It solves very obvious cases of division by zero, null de-reference and integer overflow. The analysis tries to execute expressions and statements until it reaches a join point of two or more branches. At that moment, it stops. The analysis does not carry out branch joining or fix-point computation.

For example, it is able to detect the division by zero in this:
```
fun test(v: Int): Int {
   let a = 10;
   let x = v + 1;
   if (a >= 10) {       // Condition known at compile-time, so, no branching involved.
      x = v;
      a = 5;
   } else {
      a = v;
   }
   return 1 / (a - 5);   // Division by zero
 }
 ```
 But not in here:
 ```
fun test(v: Int): Int {
   let a = 10;
   let x = v + 1;
   if (v >= 10) {   // Branching: cannot determine condition at compile-time.
      x = v;
      a = 5;
   } else {
      a = 5;
   }
   return 1 / (a - 5);   // No division by zero detected, since it cannot join paths (FunC detects the error)
 }

 ```
If needed, we could get it closer to the code in PR #852 by later incremental additions. This PR replaces PR #852. 

NOTE: This solution is intended to be temporary until proper constant propagation is implemented.

## Checklist

### CODE REVIEW

File: src/optimizer/deprecated/test/failed/assignments.tact
Comment (anton-trunov) [inline src/optimizer/deprecated/test/failed/assignments.tact:1]: tests cannot be deprecated: these are our specification, so when you switch to a proper implementation, tests should still pass

**Review by anton-trunov — CHANGES_REQUESTED** [review]

**Comment by anton-trunov** [comment]
What will we need to change in the codebase to switch to a more advanced solution?

**Comment by jeshecdom** [comment]
> What will we need to change in the codebase to switch to a more advanced solution?

Depending on what we want to do with constant propagation:

- For just detecting errors: I could keep adding code for path joining and fixpoint computation, which is available in the closed PR #852 . That code just needs adapting, because of the several refactors the codebase has went through. However, there is one big change: I would need to adapt the interpreter so that it can work with arbitrary lattice values and this is tricky for structs. I haven't thought the details though, at this moment is just a hunch. I handled that problem in the closed PR #852 with several hacks, but the main one was to wrap the environment stack class with a class that handles lattice values (I do not like such solution).

- Additionally, for optimizing code: This one is harder to predict. I can predict the initial changes that would be required with more confidence.
  + Add the optimization phase as a separate phase (currently tentative in PR #1047).
  + The transformation of AST to CFG would probably not require changes to the codebase. However, there is a potential point that could impact it, but I am not sure: handling the `SrcInfo` of the nodes produced during CFG transformation.
  + I would need to adapt the interpreter so that it handles arbitrary lattice values, as in the "just for detecting errors" case.
  + The actual algorithms for computing the fixpoints I think would not require changes to the codebase.
  + For sending the CFG back into an AST I still do not have a clear image on how to do that, so I am not sure if this would make a codebase impact.... probably not, but it is just a hunch.

File: src/ast/util.ts [hunk 1]
```diff
@@ -3,33 +3,71 @@ import * as A from "./ast";
...
+        operand: A.AstExpression,
+        loc: SrcInfo,
+    ): A.AstOpUnary {
+        const result = createNode({
+            kind: "op_unary",
+            op: op,
+            operand: operand,
+            loc,
+        });
+        return result as A.AstOpUnary;
```
i582, line 54: We don't use type casts as stated in https://github.com/tact-lang/tact/blob/main/dev-docs/STYLEGUIDE.md

i582: Looks like you can just write it as:
```
return createNode({
   kind: "op_unary",
   op: op,
   operand: operand,
   loc,
});
```

src/optimizer/interpreter.ts
```diff
@@ -523,6 +537,31 @@ class EnvironmentStack {
...
+    }
+
+    private copyEnvironment(env: Environment): Environment {
+        const newMap: Map<string, BindingState> = new Map();
+
+        for (const [name, val] of env.values) {
+            newMap.set(name, this.copyValue(val));
+        }
+
+        let newParent: Environment | undefined = undefined;
```
i582:
```typescript
const newParent = 
      typeof env.parent !== "undefined" 
    ? this.copyEnvironment(env.parent) 
    : undefined;
```
To remove mutability as stated in STYLEGUIDE.md

src/types/resolveDescriptors.ts
```diff
@@ -2237,22 +2236,24 @@ function initializeConstantsAndDefaultContractAndStructFields(
...
+                                astF,
                             );
                         } else {
                             // if a field has optional type and it is missing an explicit initializer
                             // we consider it to be initialized with the null value
 
                             field.default =
                                 field.type.kind === "ref" && field.type.optional
-                                    ? util.makeNullLiteral(field.ast.loc)
+                                    ? getAstUtil(astF).makeNullLiteral(
```
i582: Maybe we can extract the `getAstUtil()` call out of the loops? To line 2216 for example

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+    AstStatementLet,
+    AstStatementRepeat,
+    AstStatementReturn,
+    AstStatementTry,
+    AstStatementUntil,
+    AstStatementWhile,
+    AstStaticCall,
+    AstStructFieldValue,
+    AstStructInstance,
+} from "../../ast/ast";
```
i582: For such long imports we use `import * as A from "../../ast/ast";`

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+    Interpreter,
+    throwNonFatalErrorConstEval,
+} from "../interpreter";
+import { getExpType } from "../../types/resolveExpression";
+import { cloneNode } from "../../ast/clone";
+import { StructFunctions } from "../../abi/struct";
+import { MapFunctions } from "../../abi/map";
+
+/**
+ * @deprecated FIXME: Remove this error class once constant propagation is implemented.
```
i582, line 35: Why is this deprecated? Is it deprecated because it should be removed when the "No-simplified" version is implemented?

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+/**
+ * @deprecated FIXME: Remove this function once constant propagation is implemented.
+ * WARNING: DO NOT IMPORT THIS FUNCTION INTO YOUR FILES.
+ */
+export function constantPropagationAnalysis(
+    ctx: CompilerContext,
+    astF: FactoryAst,
+) {
+    // Check that the builtin Functions known by the analyzer are still the ones in StructFunctions and MapFunctions
+    const knownStructBuiltInFunctions =
```
i582: Maybe following for better readability
```
const knownStructBuiltInFunctions = [
    ...knownStructBuiltInNonMutationFunctions,
    ...knownStructBuiltInMutationFunctions,
];
```

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+            });
+        } catch (e) {
+            if (!(e instanceof InterruptedAnalysis)) {
+                throw e;
+            }
+        }
+    }
+
+    function executeStatements(statements: AstStatement[]) {
+        statements.forEach((currStmt) => {
```
i582, line 203:
```typescript
statements.forEach(interpretStatement)
```

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+                envStack.deactivateBinding(idText(ast.path));
+            }
+        } else {
+            const fullPath = tryExtractPath(ast.path);
+            if (fullPath !== null) {
+                const baseName = fullPath[0];
+                if (typeof baseName !== "undefined") {
+                    envStack.deactivateBinding(idText(baseName));
+                } else {
+                    throwInternalCompilerError(
```
i582, line 298: Looks like it is impossible case:

```typescript
export function tryExtractPath(path: A.AstExpression): A.AstId[] | null {
    switch (path.kind) {
        case "id":
            return [path];
        case "field_access": {
            const p = tryExtractPath(path.aggregate);
            return p ? [...p, path.field] : null;
        }
        default:
            return null;
    }
}
```

And I don't sure that we need to handle it anyway.

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+                        "assignments allow non-empty path expressions only",
+                        ast.path.loc,
+                    );
+                }
+            } else {
+                throwInternalCompilerError(
```
i582, line 304: Here as well, I don't think we should throw such errors in constant propagation logic

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+                    executeStatements(ast.trueStatements);
+                });
+            } else {
+                if (ast.falseStatements !== null) {
+                    const falseStmts = ast.falseStatements;
+                    envStack.executeInNewEnvironment(() => {
+                        executeStatements(falseStmts);
+                    });
+                }
+                if (ast.elseif !== null) {
```
i582: It looks like here we will execute both `else` and some `else-if` if their condition is true

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+            let iterCount = 0;
+            do {
+                executeStatements(ast.statements);
+
+                iterCount++;
+                if (iterCount >= interpreter.getConfig().maxLoopIterations) {
+                    // Interrupt the computation
+                    throw new InterruptedAnalysis();
+                }
+                const rawCond = tryExpressionEvaluation(ast.condition);
```
i582: Maybe `conditionValue` for more clarity

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+            } while (!condition);
+        });
+    }
+
+    function interpretWhileStatement(ast: AstStatementWhile) {
+        envStack.executeInNewEnvironment(() => {
+            let condition: boolean;
+            let iterCount = 0;
+            do {
+                const rawCond = tryExpressionEvaluation(ast.condition);
```
i582: And here as well `conditionValue`

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+        } else {
+            // Branch point detected.
+            throw new InterruptedAnalysis();
+        }
+    }
+
+    function interpretMethodCall(expr: AstMethodCall): AstLiteral | undefined {
+        const self = tryExpressionEvaluation(expr.self);
+        const argValues = expr.args.map((e) => tryExpressionEvaluation(e));
+        if (
```
i582: I think we can use early return here for better readability
```typescript
if (self === undefined || argValues.some((v) => v === undefined)) return undefined;
// ...
```

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+                ),
+            );
+
+            // If the method is a mutates function, then the expression acts as an implicit assignment statement
+            // into expr.self.
+            // We need to delete the binding for the path expression in expr.self, because
+            // currently, method calls in the interpreter do not modify self and they do not return the new value for self.
+
+            const fullPath = tryExtractPath(expr.self);
+            if (fullPath !== null) {
```
i582: Maybe early return here as well to reduce nesting

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+                        ) {
+                            envStack.deactivateBinding(idText(baseName));
+                        }
+                        // Not a mutates function, do nothing
+                    }
+
+                    // Not a reference or map type, so, it cannot have mutates functions
+                    // Do nothing
+                } else {
+                    throwInternalCompilerError(
```
i582: I don't think constant propagation logic should make such throws

src/optimizer/deprecated/constant-propagation.ts
```diff
@@ -0,0 +1,881 @@
...
+
+        return util.makeStructValue(structValueFields, expr.type, expr.loc);
+    }
+
+    function interpretBinaryOp(expr: AstOpBinary): AstLiteral | undefined {
+        // Boolean operators are a source of branching due to short-circuiting
+
+        if (expr.op === "&&") {
+            const leftV = tryExpressionEvaluation(expr.left);
+            if (typeof leftV !== "undefined") {
```
i582, line 790: We can use early return here to reduce nesting:
```typescript
if leftValue === undefined {
    throw new InterruptedAnalysis();
}
```

src/ast/util.ts
jeshecdom: The main problem is that `createNode` returns an `AstNode`, so we are forced to do the cast, because `createNode` cannot deduce from the `kind` property that it is actually an `AstOpUnary`. I think that this is one of those gray cases where the guidelines need to add an exception: as long as the cast is isolated in a very short method where it is obvious that the cast is justified, as in the methods in `util`. I do not know how `createNode` would need to be implemented so that its return type is the correct type given the `kind` property.

src/optimizer/deprecated/constant-propagation.ts
jeshecdom: Exactly. I thought of using deprecated to deter people from importing functions from this file, just in case they did not read the huge warning at the start of the file. This is a temporary solution until we have a more complete solution.

src/optimizer/deprecated/constant-propagation.ts
jeshecdom: I did it like in the code because of the guidelines, hahaha... the guidelines say: "Don't pass functions directly to built-in functions", but I probably violated this guideline in some other part of the code. I will need to check.

src/optimizer/deprecated/constant-propagation.ts
jeshecdom: I did not fully understand the comment. `tryExtractPath` returns `null` when the input AST is "wrong" for some reason, which happens when it is neither an id nor a field access. 

Probably processing the `AstStatementAssign` statement is the most confusing part of the implementation. The following explanation may clarify a bit.

This simple implementation takes two cases, when the path expression is a single identifier, it behaves as expected. But when it is not an identifier, it attempts to extract the path in the expression (i.e., extract the path from `ast.path`, which should have the type of a field access). After extracting the path, it takes the first element in the path and deactivates it from the environment. So, an assignment like `a.b.t = some_expression`, will deactivate `a` in the environment. This means that this simple version does not support assignment to individual fields in structs. I could add it, but handling them is a bit tricky, because we need to update the path inside the struct, i.e., create the fields `b.t` inside `a` if they do not exist (for example, `a` could have been a parameter to the function, so, all its fields do not exist).

src/optimizer/deprecated/constant-propagation.ts
jeshecdom: We usually throw compiler internal error when there is something "wrong" with the AST itself. For example, here the expression in `ast.path` should be a non-empty sequence of identifiers, not some arbitrary expression, like `a.f() = 4`, or `5 + a = g()` in the left part of the assignment. We throw fatal errors when the error has to do with the execution itself. To be honest, sometimes I get confused about which kind of error should be reported XD.

src/optimizer/deprecated/constant-propagation.ts
jeshecdom: As I discussed with you, if the AST is OK, an `elseif` and an `else` will not happen simultaneously. But you are correct that I need to add checks in case the AST is NOT correct. Added them, with a compiler internal error when the case is wrong, signaling a wrong AST.

src/optimizer/deprecated/constant-propagation.ts
jeshecdom: Actually, your comment allowed me to find a bug. In case `self` or one of the arguments is actually undefined, then we should deactivate the binding in `expr.self`, because its assignment failed (as long as the method is a mutating function). In other words, in all cases we deactivate `expr.self`. Have a look at the change. As usual, this simple version deactivates the first element in a path like `a.f.g.h`.

src/optimizer/deprecated/constant-propagation.ts
jeshecdom: That would be ok if only `&&` and `||` were the operators, but for the rest of the operators there is no branching point if the left value is undefined. So, the analysis should continue for them.

src/optimizer/deprecated/constant-propagation.ts
i582: Ah, I see

i582 (dismissed):
LGTM

src/ast/util.ts
```diff
@@ -3,33 +3,71 @@ import * as A from "./ast";
...
+export const getAstUtil = ({ createNode, cloneNode }: FactoryAst) => {
+    /**
+     * @deprecated, since it uses dummySrcInfo. Use makeUnaryExpressionLoc instead.
+     *
+     */
     function makeUnaryExpression(
         op: A.AstUnaryOperation,
         operand: A.AstExpression,
-    ): A.AstExpression {
+    ): A.AstOpUnary {
```
anton-trunov: please revert all the unrelated changes, open a new issue and solve it in a new PR

anton-trunov (changes requested):

i582 (dismissed):
LGTM

anton-trunov:
Merge conflicts

PR #1607, opened 2025-01-28
fix(interpreter): short-circuiting of `&&` and `||`

### LINKED ISSUE #1606: Interpreter does not short-circuit on operators && and ||

- [x] I'm using the latest Tact version

### Tact source code

```tact
const A: Bool = true || foo();
// const B: Bool = false && foo();   // Commented, this would also fail.

fun foo(): Bool {
    while(true) {
    }
    return true;
}
```

```shell

```

### What happened?

The above code takes a while to fail compilation (but it should NOT fail). It fails because the interpreter attempts to evaluate the second operand of || and takes a while because the default max iterations for loops is set to `2^31 - 1` (which is a huge number).

This problem was not happening in other parts of the code, for example in the semantics tests, because during FunC code generation `writeExpression` explicitly sets the max loop iterations to `2^12`, and also, when there is a non-fatal failure, like loop timeout, the compiler stops simplification but continues compilation.

I probably broke short-circuiting in the interpreter while doing the refactor of the `Value` type 🤦‍♂️. This is another reason of why I need to do extensive testing of the interpreter in isolation. Currently it is only tested implicitly but it is hard to know when it produces non-fatal errors, because compilation continues.

### What did you expect?

The following code in the interpreter:
```
        case "&&": {
            const astLeft = ensureBoolean(valLeft);
            const astRight = ensureBoolean(valRightContinuation());
            const result = astLeft.value && astRight.value;
            return util.makeBooleanLiteral(result, source);
        }
        case "||": {
            const astLeft = ensureBoolean(valLeft);
            const astRight = ensureBoolean(valRightContinuation());
            const result = astLeft.value || astRight.value;
            return util.makeBooleanLiteral(result, source);
        }
```
should be changed to:
```
        case "&&": {
            const astLeft = ensureBoolean(valLeft);
            const result = astLeft.value && ensureBoolean(valRightContinuation()).value;
            return util.makeBooleanLiteral(result, source);
        }
        case "||": {
            const astLeft = ensureBoolean(valLeft);
            const result = astLeft.value || ensureBoolean(valRightContinuation()).value;
            return util.makeBooleanLiteral(result, source);
        }
```

### Steps to reproduce

_No response_

### How do you run Tact?

_No response_

### Anything else?

_No response_

jeshecdom:
## Issue

Closes #1606.

~~I also set the default max iterations configuration in the interpreter to `2^12 = 4096`.~~ (This will be addressed in issue #1617).

## Checklist


src/optimizer/interpreter.ts
```diff
@@ -675,8 +675,8 @@ export function parseAndEvalExpression(
 
 const defaultInterpreterConfig: InterpreterConfig = {
     // We set the default max number of loop iterations
-    // to the maximum number allowed for repeat loops
-    maxLoopIterations: maxRepeatStatement,
+    // to 2^12 = 4096 to increase compiler responsiveness.
+    maxLoopIterations: 2n ** 12n,
```
anton-trunov: why this change?

src/optimizer/test/short-circuit-inits.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+import { getParser } from "../../grammar";
+import { defaultParser } from "../../grammar/grammar";
+import { resolveDescriptors } from "../../types/resolveDescriptors";
+import { getAllExpressionTypes } from "../../types/resolveExpression";
+import { resolveSignatures } from "../../types/resolveSignatures";
+import { resolveStatements } from "../../types/resolveStatements";
+import { loadCases } from "../../utils/loadCases";
+
+describe("constant-propagation", () => {
+    for (const r of loadCases(__dirname + "/short-circuit/success/")) {
```
anton-trunov: we definitely do not want to add this boilerplate to each test sub-category, like "short-circuit"

anton-trunov (changes requested):

src/optimizer/test/short-circuit-inits.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+        it(
+            r.name +
+                " should pass boolean short-circuiting during initialization",
```
i582:
```
`${r.name} should pass boolean short-circuiting during initialization`
```

i582 (dismissed):
LGTM!

Just one small nitpick 😄

src/optimizer/interpreter.ts
jeshecdom: The previous default value for loop timeout is too big. When compilation of a long loop happens, it makes the user believe that the compiler entered into an infinite loop or that just broke. But probably I need to to this change in a separate PR.

src/optimizer/test/short-circuit-inits.spec.ts
jeshecdom: I did not fully get it. Do you mean that I should remove the `short-circuit` folder?

src/optimizer/test/short-circuit-inits.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    for (const r of loadCases(__dirname + "/success/")) {
+        it(`${r.name} should pass boolean short-circuiting during initialization`, () => {
+            const Ast = getAstFactory();
+            let ctx = openContext(
+                new CompilerContext(),
+                [{ code: r.code, path: "<unknown>", origin: "user" }],
+                [],
+                getParser(Ast, defaultParser),
+            );
+            ctx = featureEnable(ctx, "external");
```
anton-trunov: do you really need this feature?

src/optimizer/test/__snapshots__/short-circuit-inits.spec.ts.snap
anton-trunov, line 1: Let's split the snapshot into two: one for positive tests and one for negative tests. The positive tests snapshot is not very informative, but it is very large.

anton-trunov (changes requested):

src/optimizer/test/short-circuit-inits.spec.ts
jeshecdom: Not really, removed.

src/optimizer/test/__snapshots__/short-circuit-inits.spec.ts.snap
jeshecdom: I'll do the same for the other pending PRs.

anton-trunov (approved):

PR #1614, opened 2025-01-28
refactor: optimize `readForwardFee` stdlib function

### LINKED ISSUE #1555: Optimize `readForwardFee`
https://github.com/tact-lang/tact/issues/1212#issuecomment-2590822116

Shvandre:
## Issue

Closes #1555 .

I did some manual optimization and improved code of that function from
```asm
execute BLKDROP2 3,1
    execute LDMSGADDR
    execute SWAP
    execute DROP
    execute LDGRAMS
    execute NIP
    execute PUSHINT 1
    execute SDSKIPFIRST
    execute LDGRAMS
    execute NIP
    execute LDGRAMS
    execute DROP
    execute PUSHINT 0
    execute GETORIGINALFWDFEE
```
to
```asm
execute LDMSGADDR
    execute LDGRAMS
    execute PUSHINT 1
    execute SDSKIPFIRST
    execute LDGRAMS
    execute LDGRAMS
    execute DROP
    execute BLKDROP2 6,1
    execute PUSHINT 0
```

## Checklist

- [ ] I have added tests to demonstrate the contribution is correctly implemented: it reduces gas consumption for the common cases, makes storage improvements, etc. and does not have too much negative impact for the rest of the use-cases


src/stdlib/stdlib/std/context.tact
```diff
@@ -11,11 +11,20 @@ native context(): Context;
 @name(__tact_context_get_sender)
 native sender(): Address;
 
+asm fun internalReadForwardFee(ctx: Context): Int {
```
novusnota: Cool, but why adding another function if the parameters and return values didn't change and the whole body was replaced by a single function call? Assembly functions can have an `extends` attribute :)

Shvandre: Oh, thanks. I tried `extends asm fun`, it didn't work. 
`asm extends fun` works. Thank you)

i582: @Shvandre maybe we can improve the error message in this case, can you create an issue?

novusnota:
Re-generate the `stdlib.ts` with `yarn gen:stdlib` to fix the linter's CI

src/stdlib/stdlib/std/context.tact
```diff
@@ -11,11 +11,15 @@ native context(): Context;
...
-    sc.loadCoins(); // Skip ihr_fee
-    return (sc.loadCoins() * 3) / 2;
+asm extends fun readForwardFee(self: Context): Int {
+    LDMSGADDR
+    LDGRAMS
+    ONE
+    SDSKIPFIRST
+    LDGRAMS
+    LDGRAMS
+    DROP
```
novusnota, line 21: Aren't you dropping the fwd_fee there and using ihr_fee instead? Perhaps, instead of the `DROP` there should be a `SWAP DROP`?

novusnota (changes requested):
It seems like there are no tests that use `Context.readForwardFee()`, so could you please add those?

anton-trunov (changes requested):
need a test showing the improvement: https://github.com/tact-lang/tact/blob/main/src/test/benchmarks/benchmarks.spec.ts

anton-trunov:
let's add that test in two commits:
- one before the optimization
- and one after it

src/stdlib/stdlib/std/context.tact
Shvandre: `LDGRAMS` does this `s - x s'`, where `s` is slice, and `s'` is remaining part, (that is why we can do `SDSKIPFIRST LDGRAMS LDGRAMS` - slice is the top element). So `DROP` just removes that remaining slice

novusnota: Got it, thanks!

src/test/benchmarks/benchmarks.spec.ts
```diff
@@ -63,4 +64,29 @@ describe("benchmarks", () => {
...
+                    .storeStringTail("This is test payload")
+                    .asSlice(),
+            },
+        );
+        const gasUsed = (
+            (
+                sendResult.transactions[1]!
+                    .description as TransactionDescriptionGeneric
+            ).computePhase as TransactionComputeVm
+        ).gasUsed;
```
anton-trunov: looks like this needs to be extracted into a helper, it is used everywhere

anton-trunov (changes requested):
What is the gas difference for `readForwardFee`?

anton-trunov (approved):

PR #1621, opened 2025-01-28
fix(interpreter): environment stack max level check

### LINKED ISSUE #1618: Add deepness check in environment stack
Currently, the interpreter does not check the deepness of the environment stack. So, a code like this with infinite recursion:
```
const A: Int = recurse();

fun recurse(): Int {
    return recurse();
} 
```
will report a Typescript error like this:
```
RangeError: Maximum call stack size exceeded
    at Interpreter.evalStaticFunction (/home/jesus/tact/tact/src/optimizer/interpreter.ts:1511:30)
    at /home/jesus/tact/tact/src/optimizer/interpreter.ts:1453:42
    at Interpreter.inComputationPath (/home/jesus/tact/tact/src/optimizer/interpreter.ts:1818:21)
    at Interpreter.interpretStaticCall (/home/jesus/tact/tact/src/optimizer/interpreter.ts:1450:41)
    at Interpreter.interpretExpression (/home/jesus/tact/tact/src/optimizer/interpreter.ts:879:29)
    at Interpreter.interpretReturnStatement (/home/jesus/tact/tact/src/optimizer/interpreter.ts:1736:30)
    ..............
```

The solution is to add a deepness check in the environment stack, and set a max environment stack deepness in `InterpreterConfig`. When the environment stack exceeds its max deepness, it should throw a non-fatal error (i.e., similar behavior as with loop time out).

jeshecdom:
## Issue

Closes #1618.

## Checklist


src/optimizer/interpreter.ts
```diff
@@ -1834,4 +1897,19 @@ export class Interpreter {
...
+            return code();
+        } catch (e) {
+            // TODO: Add here handlers for different operating systems.
+            if (e instanceof RangeError) {
+                throwNonFatalErrorConstEval(
+                    "execution stack reached maximum allowed deepness",
+                    loc,
+                );
+            }
+            throw e;
```
anton-trunov: Let's keep this change, fix the TODO, keep the tests and remove all the other irrelevant changes. If you want to refactor the interpreter, please open a separate issue and solve it in a new PR

anton-trunov (changes requested):

src/optimizer/interpreter.ts
verytactical: Also code should work the same way on any OS. Checking for OS can only be done via `process.platform`, and it won't work in browser. Then we'll need a User-Agent parsing library to polyfill it, and it will get even more complicated from there.

jeshecdom: > remove all the other irrelevant changes.

The thing is that I will still need to keep the renaming of `interpretExpression` ---> `interpretExpressionInternal`, similarly for `interpretStatement` and `interpretModuleItem` because I need the top level entry to the interpreter to handle the stack overflows. If I leave it as is, I will need to add a call to `handleStackOverflow` at every single call inside the interpreter because external processes will be able to enter at any place inside the interpreter because all its methods are public.

jeshecdom: > Also code should work the same way on any OS. Checking for OS can only be done via `process.platform`, and it won't work in browser. Then we'll need a User-Agent parsing library to polyfill it, and it will get even more complicated from there.

Do you know if there is a list somewhere that lists how TypeScript reports stack overflows in different OSes? I cannot find anything searching at Google. I will also add a check on the exception message, just in case the RangeError is not a Stack overflow.

verytactical (dismissed):

src/optimizer/interpreter.ts
```diff
@@ -614,7 +617,7 @@ class EnvironmentStack {
         return this.findBindingMap("self") !== undefined;
     }
 
-    /*
+    /**
```
i582, line 619: Since this is a doc comment now, we need to add `*` to each line:

```typescript
/**
 *
 */
```

src/optimizer/interpreter.ts
```diff
@@ -1834,4 +1893,27 @@ export class Interpreter {
 
         return `${shortPath.join(" -> ")} -> ${name}`;
     }
+
+    private handleStackOverflow<T>(loc: SrcInfo, code: () => T): T {
+        try {
+            return code();
+        } catch (e) {
+            const finalErrorMessage =
+                "execution stack reached maximum allowed deepness";
```
verytactical: depth

anton-trunov (dismissed):

PR #1622, opened 2025-01-29
fix: allow constant/trait constants depend on each other

### LINKED ISSUE #924: Contract constants cannot depend on each other

### Tact source code

```tact
contract Test {
    const A: Int = 41;
    const B: Int = self.A + 1;
    get fun b(): Int { return self.B }
}
```

```shell
Syntax and type checking failed
Error: test.tact:3:20: Unable to resolve id 'self'
Line 3, col 20:
  2 |     const A: Int = 41;
> 3 |     const B: Int = self.A + 1;
                         ^~~~
  4 |     get fun b(): Int { return self.B }

    at throwCompilationError (/Users/anton_1/ton/tact/dist/errors.js:81:11)
    at resolveExpression (/Users/anton_1/ton/tact/dist/types/resolveExpression.js:501:56)
    at resolveFieldAccess (/Users/anton_1/ton/tact/dist/types/resolveExpression.js:244:11)
    at resolveExpression (/Users/anton_1/ton/tact/dist/types/resolveExpression.js:511:20)
    at resolveBinaryOp (/Users/anton_1/ton/tact/dist/types/resolveExpression.js:102:11)
    at resolveExpression (/Users/anton_1/ton/tact/dist/types/resolveExpression.js:463:20)
    at checkInitializerType (/Users/anton_1/ton/tact/dist/types/resolveDescriptors.js:1415:53)
    at initializeConstants (/Users/anton_1/ton/tact/dist/types/resolveDescriptors.js:1425:19)
    at initializeConstantsAndDefaultContractAndStructFields (/Users/anton_1/ton/tact/dist/types/resolveDescriptors.js:1456:27)
    at resolveDescriptors (/Users/anton_1/ton/tact/dist/types/resolveDescriptors.js:1320:11)
```

### What happened?

_No response_

### What did you expect?

Compilation succeeds

### Steps to reproduce

_No response_

### How do you run Tact?

Tact CLI

### Anything else?

_No response_

### LINKED ISSUE #1645: Contract constants cannot be used in contract variable initializers
Compiling the following
```tact
contract Test {
    const A: Int = 0;
    x: Int = self.A;
}
```

results in an error message like so:

```
Error: test.tact:3:14: Cannot find 'self'
  2 |     const A: Int = 0;
> 3 |     x: Int = self.A;
                   ^~~~
  4 | }
```

i582:
## Issue

Closes #924
Closes #1645 

## Checklist


src/test/e2e-emulated/contracts/self-constants.tact
```diff
@@ -0,0 +1,41 @@
...
+
+contract ConstantTester with WithConstant {
+    const A: Int = 41;
+    const B: Int = self.A + 1;
+
+    const A2: Int = 41;
+    const C2: Int = self.A2 + 1 + self.B2;
+    const B2: Int = 9;
+
+    value: Int = self.B2 + self.C2;
```
anton-trunov: let's use some of the `WithConstant` trait's constants here

src/types/resolveDescriptors.ts
```diff
@@ -2157,8 +2158,12 @@ function checkInitializerType(
     declTy: TypeRef,
     initializer: A.AstExpression,
     ctx: CompilerContext,
+    selfTypeRef: TypeRef | null,
```
anton-trunov: Styleguide: Avoid null. typeof null === 'object', and there is undefined anyway.

anton-trunov (changes requested):
@jeshecdom please review the interpreter changes

src/types/resolveDescriptors.ts
i582: Fixed

src/test/e2e-emulated/contracts/self-constants.tact
i582: Done

src/optimizer/interpreter.ts
```diff
@@ -1120,14 +1120,35 @@ export class Interpreter {
                         ast.aggregate.loc,
                     );
                 }
+
                 if (foundContractConst.value !== undefined) {
                     return foundContractConst.value;
-                } else {
+                }
+
+                const name = `self.${ast.field.text}`;
```
jeshecdom: We do it like `self.${idText(ast.field)}`

jeshecdom (commented):
What error message will you get if you do something like this?
```
contract Test {

    a: Int = 1;
    const B: Int = 10;
    const A: Int = self.B + self.test();
    
    fun test(): Int {
        return self.a;
    }
}
```

src/optimizer/interpreter.ts
i582: Thanks, fixed!

For that code we will get this error (added to tests):
```
Cannot evaluate expression to a constant: calls of "test" are not supported at this moment
```

jeshecdom: I suspected that it would give that message. The message means that the interpreter is actually attempting to run the function `test`, but currently, it does not interpret method calls and halts. This means that once we add interpretation of method calls in the interpreter we have a potential problem if `self` somehow gets added into the bindings in the interpreter.  If `self` does not get added into the bindings, then we are OK, because the function will fail at `return self.a` with something like "identifier a is not a contract constant".

Probably the safest thing to do here would be to disallow calls to contract functions while initializing constants, because `self` is still not initialized, which means that calling `self.test()` does not make sense if the function `test` reads fields from `self`. Only calls to global functions should be allowed during constant initialization. What do you think?

anton-trunov: > once we add interpretation of method calls in the interpreter

wait, what? which issue tracks the missing features of the interpreter?

i582: > Probably the safest thing to do here would be to disallow calls to contract functions while initializing constants, because self is still not initialized

We can do this (https://github.com/tact-lang/tact/issues/1638) when method calls become possible by adding a sub issue to the issue of supporting such calls (https://github.com/tact-lang/tact/issues/1637).

jeshecdom: > > once we add interpretation of method calls in the interpreter
> 
> wait, what? which issue tracks the missing features of the interpreter?

I will add a plan with a list of things that are still missing in the interpreter.

jeshecdom:
I think this PR should be put on hold after issue #1655 is resolved first, because it is related to constant initialization.

anton-trunov (approved):

PR #1626, opened 2025-01-29
fix: runtime `sha256` now works for arbitrary strings with length >= 128

### LINKED ISSUE #1056: Runtime `sha256` should work for arbitrary strings even with length >= 128
`sha256` at run-time uses the `SHA256U` TVM instruction, which only works with data with up to 1023 bits, so strings of length 128 bytes basically get truncated. This is confusing for smart-contract authors because it breaks the abstraction of the `String` type. Since we allow strings with more than 127 characters, those `sha256` should work on those.

See also, https://github.com/tact-lang/tact/issues/1085

i582:
## Issue

Closes #1056.

## Checklist


anton-trunov:
Let's resolve merge conflicts here

src/stdlib/stdlib/stdlib_ex.fc
anton-trunov, line 1: @Shvandre take a look at this code please

src/test/benchmarks/contracts/benchmark_sha256.tact
```diff
@@ -0,0 +1,29 @@
+asm fun sha256_native(s: String): Int {
+    SHA256U
+}
```
anton-trunov: we should probably add this as a gas optimizations for short strings in the docs (and perhaps in stdlib)

@novusnota wdyt?

anton-trunov: or, actually, we can suggest using `sha256(s.asSlice())` (but we need to check the result uses `SHA256U`)

anton-trunov (changes requested):
merge conflicts

src/test/benchmarks/contracts/benchmark_sha256.tact
i582: > or, actually, we can suggest using `sha256(s.asSlice())` (but we need to check the result uses `SHA256U`)

Done

src/test/benchmarks/benchmarks.spec.ts
```diff
@@ -67,4 +71,97 @@ describe("benchmarks", () => {
...
+        const result = await sha256.send(
+            treasure.getSender(),
+            { value: toNano(1) },
+            { $$type: "HashData", value: s },
+        );
+
+        return getUsedGas(result);
+    }
+
+    async function hashStringBig(
```
i582, line 88: These functions are hard to unify without any type :(

src/test/benchmarks/contracts/benchmark_sha256.tact
novusnota: > or, actually, we can suggest using `sha256(s.asSlice())` (but we need to check the result uses `SHA256U`)

It seems like it uses `SHA256U`, so we can suggest that, yeah.

Besides, with this change the description of the [`sha256()`](https://docs.tact-lang.org/ref/core-math/#sha256) should be split in two:
1. The usage prior to this change, "prior to 1.6.0"
2. "Since 1.6.0"

anton-trunov (approved):

verytactical (approved):

PR #1632, opened 2025-01-29
feat(parser): add parseStatement method for parsing statements

xpyctumo:
## Issue

Closes nothing

But it will helps for fuzzer, because we can handle Statements directly from parser.
It also small part which is working, and might be helpful for other fuzzy-issues

## Checklist

### DIFF
--- a/src/grammar/grammar.ts
+++ b/src/grammar/grammar.ts
@@ -13,6 +13,7 @@ export type Parser = {
         path: string,
         origin: ItemOrigin,
     ) => A.AstImport[];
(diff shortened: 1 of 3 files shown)


verytactical (approved):
While we need a better approach, and though the code is only needed for testing but changes runtime, the change is small and self-contained enough to merge it.

If there will be other methods added to parser's API, we'll have to block them until `doParse` is properly exposed in the interface.

anton-trunov (approved):

PR #1633, opened 2025-01-29
Add Jetton to benchmarks

### LINKED ISSUE #1540: Add `@tact-lang/jetton` to `src/test/benchmarks`

### LINKED ISSUE #1629: Separate script to run benchmarks with automatic contracts build
`yarn bench` should run the benchmarks after building all the dependencies, including stdlib and the benchmark contracts.

i582:
Tact code from: https://github.com/tact-lang/jetton

## Output

Screenshot 2025-01-29 at 20 24 37

## Issue

Closes #1540.
Closes #1629.

## Checklist


Shvandre:
LGTM

Shvandre (dismissed):

spell/cspell-list.txt
```diff
@@ -1,5 +1,6 @@
 Brujin
 bocchi
+consuption
```
anton-trunov: consumption?

spell/cspell-list.txt
```diff
@@ -188,6 +190,7 @@ unboc
 uncons
 uninit
 unixfs
+Unsufficient
```
anton-trunov: insufficient

src/test/benchmarks.build.ts
```diff
@@ -0,0 +1,12 @@
+import { __DANGER__disableVersionNumber } from "../pipeline/version";
+import { allInFolder } from "./utils/all-in-folder.build";
+
+const main = async () => {
+    // Disable version number in packages
+    __DANGER__disableVersionNumber();
+
+    await allInFolder(__dirname, ["benchmarks/contracts/*.tact"]);
+    await allInFolder(__dirname, ["benchmarks/jetton/sources/*.tact"]);
```
anton-trunov: let's put both contracts and tests into one folder so we don't create too nested directories

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,462 @@
...
+    return sendEnough.transactions
+        .slice(1)
+        .map((t) => {
+            return (
+                (t.description as TransactionDescriptionGeneric)
+                    .computePhase as TransactionComputeVm
+            ).gasUsed;
+        })
+        .reduceRight((prev, cur) => prev + cur);
+}
```
anton-trunov: This stuff looks like it can be reused in all the benchmarks

anton-trunov (changes requested):
I'm wondering if we should put all contracts into just one file.

Also, the backwards compatibility tests failed

src/test/benchmarks.build.ts
verytactical: `allInFolder` takes an array, there's no need to have 2 of them anyway.

src/test/benchmarks/jetton/utils/JettonMinterExtension.d.ts
```diff
@@ -0,0 +1,38 @@
+import { Address, Cell, ContractProvider, Sender } from "@ton/core";
+
+declare module "../output/Jetton_JettonMinter" {
```
verytactical: Why do we need this?

Can we define those functions not as extensions to the class, but just as separate functions called on objects of that class?

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,462 @@
...
+} from "@ton/core";
+import { Blockchain, SandboxContract, TreasuryContract } from "@ton/sandbox";
+
+import {
+    JettonMinter,
+    TokenUpdateContent,
+    TokenBurn,
+    Mint,
+    ProvideWalletAddress,
+} from "./sources/output/jetton_minter_discoverable_JettonMinter";
```
verytactical: Github Actions are failing on this file.

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,462 @@
...
+    });
+    process.stdout.write("\n");
+}
+
+function getUsedGas(sendEnough: SendMessageResult) {
+    return sendEnough.transactions
+        .slice(1)
+        .map((t) => {
+            return (
+                (t.description as TransactionDescriptionGeneric)
```
verytactical: Could you please add a comment explaining why the `as` is required here? It can be a reference to a Github issue in microsoft/typescript or the library.

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,462 @@
...
+        head: ["Run", "Transfer", "Burn", "Discovery"],
+        style: {
+            head: ["cyan"],
+            border: ["gray"],
+        },
+    });
+
+    const changes = calculateChanges(results);
+
+    results.forEach((result, i) => {
```
verytactical: `forEach`+`push` is a `map`

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,462 @@
...
+    const last = results[results.length - 1]!;
+    const compareMetrics = ["transfer", "burn", "discovery"] as const;
+
+    process.stdout.write("\nComparison with FunC implementation:\n");
+    compareMetrics.forEach((metric) => {
+        const ratio = (
+            (Number(last[metric]) / Number(first[metric])) *
+            100
+        ).toFixed(2);
+        process.stdout.write(
```
verytactical: Why do we use low-level API here?

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,462 @@
...
+            `${metric.charAt(0).toUpperCase() + metric.slice(1)}: ${
+                parseFloat(ratio) > 100
+                    ? `\x1b[91m${ratio}%\x1b[0m`
```
verytactical: ANSI coloring should be done via a library. Consider `picocolor`, `chalk`, or our own file with functions that wrap text in these codes.

That said, not all terminals support these codes (for example, Github CI doesn't), and you should detect terminal mode before emitting these codes.

package.json
```diff
@@ -17,6 +17,7 @@
     "gen:func-js": "ts-node src/func/func.build.ts",
     "gen:contracts:examples": "ts-node examples/contracts.build.ts",
     "gen:contracts:test": "ts-node src/test/contracts.build.ts",
+    "gen:contracts:benchmarks": "ts-node src/test/benchmarks.build.ts",
```
verytactical, line 20: Why do we need to build them separately?

i582: For much faster local benchmark execution

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,462 @@
...
+];
+
+function calculateChanges(results: BenchmarkResult[]): string[][] {
+    type MetricKey = Exclude<keyof BenchmarkResult, "label">;
+    const headers = ["transfer", "burn", "discovery"] as const;
+    const changes: string[][] = results.map(() => headers.map(() => ""));
+
+    for (let i = 1; i < results.length; i++) {
+        headers.forEach((key, index) => {
+            const prevValue = results[i - 1]![key as MetricKey];
```
verytactical: These `!` and `as` will become and issue quite soon, when we'll be removing them from codebase.

This code should be rewritten into an immutable code without any unsafe cast.

src/test/benchmarks/jetton/jetton.spec.ts
i582: But we push it not into an array, but into a `Table`:

```typescript
const table = new Table({
    head: ["Run", "Transfer", "Burn", "Discovery"],
    style: {
        head: ["cyan"],
        border: ["gray"],
    },
});
```

src/test/benchmarks/jetton/jetton.spec.ts
i582: Because for some reason for `console.log` we have this output:
```
  ● Console

    console.log
      
      Comparison with FunC implementation:

      at log (src/test/benchmarks/jetton/jetton.spec.ts:130:13)
```

spell/cspell-list.txt
i582: Fixed

spell/cspell-list.txt
i582: Fixed

src/test/benchmarks.build.ts
i582: Done

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

package.json
verytactical: Sorry, I don't understand the explanation.

src/test/benchmarks/jetton/jetton.spec.ts
verytactical: This is expected, as all the output of jest tests is wrapped in this way.

Unfortunately, `process.stdout.write` and `console.*` don't really mix and can produce garbage output, so the only valid options are
- not use jest for benchmarks
- form the output in a single string, then emit it with `console.log`, even though `jest` will add its own header on top.

src/test/benchmarks/jetton/jetton.spec.ts
verytactical: Consider wrapping this Table library into a function that just takes a `string[][]`, or not adding the library with mutable API.

package.json
i582: For benchmarks, we only need to build a subset of contracts, so we extract this build step into another command so that only it is run when benchmarks are executed

anton-trunov: We need to document this workflow in CONTRIBUTING.md

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

src/test/benchmarks/jetton/jetton.spec.ts
i582: Done (second option)

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed with explicit checks

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

src/test/benchmarks/jetton/utils/JettonMinterExtension.d.ts
i582: Fixed

src/test/benchmarks/jetton/jetton.spec.ts
i582: Moved to benchmark `util.ts`

package.json
i582: Done

package.json
```diff
@@ -65,6 +67,7 @@
     "@ton/crypto": "^3.2.0",
     "@tonstudio/parser-runtime": "^0.0.1",
     "blockstore-core": "1.0.5",
+    "chalk": "4.1.2",
```
Danil42Russia: why not in `devDependencies`?

i582: Fixed

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,464 @@
...
+import { getUsedGas } from "../util/util";
+
+type BenchmarkResult = {
+    label: string;
+    transfer: bigint;
+    burn: bigint;
+    discovery: bigint;
+};
+
+const results: BenchmarkResult[] = [
```
verytactical: Please put this into a separate JSON file.

When we'll be doing more gas optimization, we'll want to add more entries into this list, and preferably automatically (with mentioning commit id).

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,464 @@
...
+        discovery: 11063n,
+    },
+];
+
+type MetricKey = "transfer" | "burn" | "discovery";
+const METRICS: readonly MetricKey[] = [
+    "transfer",
+    "burn",
+    "discovery",
+] as const;
```
verytactical: If you explicitly set the type of the variable, there's no need to keep `as const` here.

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,464 @@
...
+    const changes = calculateChanges(results);
+
+    results
+        .map((result, i) => [
+            result.label,
+            `${result.transfer} ${changes[i]?.[0] ?? ""}`,
+            `${result.burn} ${changes[i]?.[1] ?? ""}`,
+            `${result.discovery} ${changes[i]?.[2] ?? ""}`,
+        ])
+        .forEach((arr) => {
```
verytactical, line 95: :+1:

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,464 @@
...
+            table.push(arr);
+        });
+
+    let output = "";
```
verytactical:
```suggestion
const output: string[] = [];
output.push(table.toString());
```

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,464 @@
...
+    output += table.toString();
+    output += "\n";
+
+    const first = results[0]!;
+    const last = results[results.length - 1]!;
+    const compareMetrics = ["transfer", "burn", "discovery"] as const;
+
+    output += "\nComparison with FunC implementation:\n";
```
verytactical:
```suggestion
output.push("Comparison ...");
```

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,464 @@
...
+        ).toFixed(2);
+        output += `${metric.charAt(0).toUpperCase() + metric.slice(1)}: ${
+            parseFloat(ratio) > 100
+                ? chalk.redBright(`${ratio}%`)
+                : chalk.green(`${ratio}%`)
+        } of FunC gas usage\n`;
+    });
+    output += "\n";
+
+    console.log(output);
```
verytactical:
```suggestion
console.log(output.join('\n'));
```

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,464 @@
...
+    let output = "";
+    output += table.toString();
+    output += "\n";
+
+    const first = results[0]!;
+    const last = results[results.length - 1]!;
+    const compareMetrics = ["transfer", "burn", "discovery"] as const;
+
+    output += "\nComparison with FunC implementation:\n";
+    compareMetrics.forEach((metric) => {
```
verytactical:
```suggestion
output.push(...compareMetrics.map(?));
```

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,464 @@
...
+    sendBurn(
+        via: Sender,
+        value: bigint,
+        jetton_amount: bigint,
+        responseAddress: Address,
+        customPayload: Cell | null,
+    ): Promise<SendMessageResult>;
+}
+
+interface ExtendedJettonMinter extends SandboxContract<JettonMinter> {
```
verytactical: Now these are separate functions. We don't have to extend wrappers anymore.

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,464 @@
...
+            table.push(arr);
+        });
+
+    let output = "";
+    output += table.toString();
+    output += "\n";
+
+    const first = results[0]!;
+    const last = results[results.length - 1]!;
+    const compareMetrics = ["transfer", "burn", "discovery"] as const;
```
verytactical: This map is already defined above.

src/test/benchmarks/jetton/jetton.spec.ts
i582: See `userWallet = async ...`

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

src/test/benchmarks/util/util.ts
anton-trunov, line 1: `src/test/benchmarks/util/util.ts` -> `src/test/benchmarks/util.ts`

src/test/benchmarks/jetton/jetton.spec.ts
i582: 👍

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

src/test/benchmarks/jetton/jetton.spec.ts
i582: Done

src/test/benchmarks/util/util.ts
i582: Fixed

src/test/benchmarks/contracts/jetton_minter_discoverable.tact
```diff
@@ -0,0 +1,142 @@
...
+    }
+    receive(msg: Mint) {
+        self.requireOwner(); // Allow minting only by owner
+        require(self.mintable, "Not mintable");
+        //Maybe we should check that msg.value is enough to cover the gas fees
+        //Or, maybe we should do self.totalSupply -= msg.amount if bounced.
+        //But there is no any check in Howard's code and in official funC code,
+        self.totalSupply += msg.amount; // Update total supply
+
+        let winit: StateInit = self.getJettonWalletInit(msg.receiver);
```
verytactical: Instead of adding `winit` to cspell exceptions, better call it `walletInit`.

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,419 @@
...
+            `${result.burn} ${changes[i]?.[1] ?? ""}`,
+            `${result.discovery} ${changes[i]?.[2] ?? ""}`,
+        ])
+        .forEach((arr) => {
+            table.push(arr);
+        });
+
+    const output = [];
+    output.push(table.toString());
+    output.push("\n");
```
verytactical: The whole point of output is to have `\n` down there in `.join`, so that we know there are as many lines as elements in `output`.

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,419 @@
...
+
+    output.push("\nComparison with FunC implementation:\n");
+    output.push(
+        ...METRICS.map((metric) => {
+            const ratio = (
+                (Number(last[metric]) / Number(first[metric])) *
+                100
+            ).toFixed(2);
+
+            return `${metric.charAt(0).toUpperCase() + metric.slice(1)}: ${
```
verytactical, line 110: If we don't have a function in utils or `change-case` library to pascalcase a string, we should have it.

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,419 @@
...
+                parseFloat(ratio) > 100
```
verytactical: Number was converted to string with `toFixed` and immediately parsed back into a number

verytactical: Consider keeping `ratio` a number, using it here, and doing `.toFixed` where you need to display it.

src/test/benchmarks/jetton/jetton.spec.ts
verytactical: And after that I'm still confused. There are 3 mentions of `sendMint`: here, where a regular `sendMint` function is defined, and where it's used directly from the scope.

src/test/benchmarks/jetton/jetton.spec.ts
```diff
@@ -0,0 +1,419 @@
...
+        head: ["Run", "Transfer", "Burn", "Discovery"],
+        style: {
+            head: ["cyan"],
+            border: ["gray"],
+        },
+    });
+
+    const changes = calculateChanges(results);
+
+    results
```
verytactical, line 88:
```ts
const show = (results: BenchmarkResult[]): string[][] =>
    results
        .map(({ label, transfer, burn, discovery }) => (
            [label, [transfer, burn, discovery]]
        ))
        .map(([label, columns], i) => {
            const change = (value: number, j: number) =>
                i === 0
                    ? String(value)
                    : `${value} ${calculateChange(rest[i - 1][j], value)}`;
            return [label, ...columns.map(change)];
        });
```

src/test/benchmarks/contracts/jetton_minter_discoverable.tact
i582: Fixed

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

src/test/benchmarks/jetton/jetton.spec.ts
i582: Okey

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

src/test/benchmarks/jetton/jetton.spec.ts
i582: Okey

src/test/benchmarks/jetton/jetton.spec.ts
i582: Fixed

anton-trunov:
The table looks incredible! 🚀

PR #1634, opened 2025-01-30
docs: clarify interactions of message sending functions and their modes

### LINKED ISSUE #1093: Clarify `SendRemainingValue` applies when first message is `emit`
A follow-up on tact-lang/tact-docs#303. See the code example in https://github.com/tact-lang/tact/issues/279

novusnota:
Also listed functions with implicit mode, plus made some minor fixes on related pages

## Issue

Closes #1093.

## Checklist


docs/src/content/docs/book/send.mdx
```diff
@@ -112,30 +112,69 @@ Consider the following example:
...
+            mode: SendRemainingValue,
+            body: "Not this again!".asComment(),
+        });
+    } // exit code 37 during action phase!
+}
+```
+
+:::note
+
+  To avoid dealing with similar cases and to simplify future [debugging sessions](/book/debug), consider having only one call to one of the [message sending functions](#message-sending-functions) per [receiver function](/book/receive), preferably at the end of the function body.
```
anton-trunov: We should suggest a different solution based on `nativeReserve` here. @Shvandre can you please help with this part

docs/src/content/docs/ref/core-common.mdx
```diff
@@ -228,7 +228,9 @@ send(SendParameters{
 fun emit(body: Cell);
 ```
 
-[Queues the message](/book/send#outbound-message-processing) `body` to be sent to the outer world with the purpose of logging and analyzing it later off-chain. The message does not have a recipient and is gas-efficient compared to using any other message sending functions of Tact.
+[Queues the message](/book/send#outbound-message-processing) `body` to be sent to the outer world with the purpose of logging and analyzing it later off-chain. The message does not have a recipient and is gas-efficient compared to using any other [message sending functions](/book/send#message-sending-functions) of Tact.
```
anton-trunov:
```suggestion
[Queues the message](/book/send#outbound-message-processing) `body` to be sent to the outer world with the purpose of logging and analyzing it later off-chain. The message does not have a recipient and is more gas-efficient compared to using any other [message sending functions](/book/send#message-sending-functions) of Tact.
```

anton-trunov (changes requested):

docs/src/content/docs/book/send.mdx
Shvandre: For this specific case (when contract initially has 0 on balance). Such code will work properly
```tact
send(SendParameters{
    bounce: false,
    body: "TEST1".asComment(),
    to: sender(), //Actually any address
    value: ton("0.5"),
    mode: SendDefaultMode,
});

send(SendParameters{
    bounce: false,
    body: "TEST2".asComment(),
    to: myAddress(), //Actually any address
    value: 0,
    mode: SendRemainingBalance,
});
 ```
 And `nativeReserve` is unnecessary here

Comment (Shvandre) [reply docs/src/content/docs/book/send.mdx]: @novusnota FYI

Comment (novusnota) [reply docs/src/content/docs/book/send.mdx]: The point is in `bounce: false` and using `SendRemainingBalance` instead?

Comment (Shvandre) [reply docs/src/content/docs/book/send.mdx]: No, bounce does not affect anything.
Point is that SendRemainingBalance works correctly, taking previous actions into account,
while SendRemainingValue does not

Comment (anton-trunov) [reply docs/src/content/docs/book/send.mdx]: nice! super-useful, thanks Andrei

let's add a solution what to do when the balance is not zero

Comment (novusnota) [reply docs/src/content/docs/book/send.mdx]: Yup, and I'll also add info on `SendRemainingValue` vs. `SendRemainingBalance` to the book/message-mode, under the table in "Base modes" section

Comment (anton-trunov) [reply docs/src/content/docs/book/send.mdx]: @Shvandre can you help finalize this PR please?

Comment (Shvandre) [reply docs/src/content/docs/book/send.mdx]: @novusnota Check this example, it describes what to do, if you want to keep non-zero balance of your contract (for some reasons, because usually it's not needed) https://github.com/Shvandre/TonFUND

@anton-trunov FWIW

File: docs/src/content/docs/book/send.mdx [hunk 3]
```diff
@@ -112,30 +112,108 @@ Consider the following example:
...
+
+    // And the rest of the logic is expressed here
+    receive() {
+        // 1st outbound message evaluated and queued (but not yet sent)
+        // with the mode 0, which is the default
+        emit("Have you seen this message?".asComment());
+
+        // Let's keep at least 0.1 Toncoin on the balance
+        // Notice that nativeReserve() only queues an action to be performed during action phase
+        nativeReserve(ton("0.1"), ReserveExact);
```
novusnota:
```suggestion
        nativeReserve(ton("0.1"), ReserveExact); // or ReserveAtMost if you're ok with reserving less
```

Shvandre (dismissed):
LGTM

novusnota:
Modifying that example with SendRemainingBalance (will probably revert it to SendRemainingValue or introduce a better solution). Additionally, will add some sort of warning for being very cautious when using `SendRemainingBalance`

novusnota:
@anton-trunov @Shvandre improved examples and emphasized the dangers of improper or careless use of base modes. Please, take a look!

P.S.: macOS CI at it again :)

docs/src/content/docs/book/send.mdx
```diff
@@ -112,40 +112,126 @@ Consider the following example:
...
+If a subsequent message is then sent with a [`SendRemainingValue{:tact}`](/book/message-mode#base-modes) base mode, it will cause the same error as before:
+
+```tact
+// This contract initially has 0 nanoToncoins on the balance
+contract IfItDiesItDies {
+    // And all the funds it gets are obtained from inbound internal messages
+    receive() {
+        // 1st outbound message evaluated and queued (but not yet sent)
+        // with the mode 0, which is the default
+        emit("Have you seen this message?".asComment());
```
Shvandre, line 160: Maybe someone will be confused with external-out messages and it is worth it to send usual message here? 
Anyway it is not crucial and up to you. Feel free to just click "resolve".

Shvandre (dismissed):

docs/src/content/docs/book/send.mdx
novusnota: Well, the idea was to show different functions and we also needed more visibility of `emit()`, so there's that :)

docs/src/content/docs/book/send.mdx
```diff
@@ -112,40 +112,126 @@ Consider the following example:
...
+        //            |            |               keeping the balance from before the compute phase start
+        //            |            might keep less, but won't fail in doing so
+        //            just a tad more on top of the balance, for the fees
+
+        // 2nd outbound message evaluated and queued
+        // with the SendRemainingBalance mode
+        send(SendParameters{
+            to: sender(),
+            value: 0,
+            mode: SendRemainingBalance, // because of the prior nativeReserve(),
```
anton-trunov, line 213: here we go again

there should not be `SendRemainingBalance` here!

we are trying to solve the following problem here: use only incoming funds and send two messages

anton-trunov (changes requested):

docs/src/content/docs/book/send.mdx
novusnota: The `SendRemainingBalance` there won't send out our balance, unlike the previous example — because of the prior `nativeReserve()` we'll use only incoming funds for those two messages.

UPD: Will add tests showing this. And a proposal for a new function to the stdlib, just so this stuff is easier to manage.

novusnota: Done, please take a look! Decided not to expand the `semantics.tact` and instead make `send.tact` and `send.spec.ts` more suitable for further message-sending tests

Additionally, opened an issue to consider adding a new function to the stdlib that'd behave similarly: #1746

novusnota:
Important bit of testing knowledge: Jest fails with bigints in `.toBe()` matchers, prefer doing all checks inside the `except()` or even in the contract (preferably) and then having a simple `.toBe(true)` check

See: https://github.com/tact-lang/tact/pull/1634/commits/f9d40fa9365c3a367e3e78ed6cce8db50a10106d

novusnota:
@anton-trunov please have a look :)

src/test/e2e-emulated/contracts/send.tact
```diff
@@ -10,4 +10,43 @@ contract SendTester with Deployable {
...
     }
-}
\ No newline at end of file
+
+    /// A no-op receiver for incoming funds
+    receive("topup") {}
+
+    /// A getter to get contracts balance
+    get fun balance(): Int { return myBalance(); }
+
```
anton-trunov: You can query any contract's balance via the Sandbox API like so:

```ts
const contractBalance = (await blockchain.getContract(contract.address)).balance;
```


```suggestion
```

anton-trunov (changes requested):

src/test/e2e-emulated/contracts/send.tact
```diff
@@ -10,4 +10,43 @@ contract SendTester with Deployable {
...
+        send(SendParameters{
+            to: sender(),
+            value: 0,
+            mode: SendRemainingBalance,
+            body: "I give you my all! Well, all that's not mine!".asComment(),
+        });
+    }
+
+    /// Sends a message, keeps the original balance, sends the rest in the next message
```
anton-trunov: The same comment as for `receive("ReserveAtMost_1")`. Let's clarify the difference

src/test/e2e-emulated/send.spec.ts
```diff
@@ -61,4 +61,66 @@ describe("send", () => {
...
+        // emit, nativeReserve, send
+        let balanceBefore = await contract.getBalance();
+        await expectMessageFromTo(
+            treasure,
+            contract,
+            textMsg("ReserveAtMost_1"),
+        );
+        let balanceAfter = await contract.getBalance();
+        // The difference is exactly 0.05 that were reserved on top of the balance
+        expect(abs(balanceBefore - balanceAfter) <= 50000000n).toBe(true);
```
anton-trunov: Let's remove `abs`, it seems in this case it should be clear if the balance "after" is smaller or greater than "before"

PR #1650, opened 2025-01-30
feat!: optimize and rework `send()` function

### LINKED ISSUE #1558: Version of `send()` using `SENDRAWMSG` instead of `SENDMSG`
https://github.com/tact-lang/tact/issues/1212#issuecomment-2594050153

> the return value of `send()` is rarely used, while there is `SENDMSG` under the hood, which consumes significantly more gas than `SENDRAWMSG`

This requires special support in the compiler to not break backwards compatibility: 
we need to detect if `send`'s return value (forward fee) is not used and use a different implementation of `send` in this case.

Another option would be to introduce a breaking change, which is fine here, since it wasn't documented.

### LINKED ISSUE #1649: `send()` function needs a better implementation
It consumes lots of gas, and may be optimized

Shvandre:
## Issue

Closes #1649 
Closes #1558 

## Checklist

Results (two last rows):
image


src/stdlib/stdlib/std/context.tact
```diff
@@ -11,15 +11,29 @@ native context(): Context;
 @name(__tact_context_get_sender)
 native sender(): Address;
 
+/*
```
novusnota: Thanks for adding `/**/` comments describing what's going on, but I'll rewrite them all properly tomorrow, if you don't mind :)

src/stdlib/stdlib/std/send.tact
```diff
@@ -21,55 +21,131 @@ struct SendParameters {
...
-    // Send
-    let c: Cell = b.endCell();
-    return nativeSendMessageReturnForwardFee(c, params.mode);
+bounce: s6
+to: s5
+value: s4
+mode: s3
+body: s2
+code: s1
+data: s0
```
novusnota: Interesting way to describe the stack from the perspective of arguments, but it conflicts with the notion of the stack itself — `s0` is the top. I suggest we use the notation already applied in the docs here: https://docs.tact-lang.org/book/assembly-functions/#examples.

That said, I respect your idea and will contemplate on better ways to present it.

See https://github.com/tact-lang/tact/pull/1650/files#r1936329961

Shvandre: It was just more convenient to me to write that in backwards way)
If it is crucial, I can reverse them, but please, let them be in a column, not a row. Row form is hard to read in complex functions (like this)

Shvandre: Moreover, in such notation function arguments are closer to a function, that is why I like that)

src/stdlib/stdlib/std/send.tact
```diff
@@ -21,55 +21,131 @@ struct SendParameters {
...
+    ISNULL
+    SWAP
+    ISNULL
+    MUL
+    /*
+    mode: s4
+    body: s3
+    MessageBuilder: s2
+    data: s1
+    code: s0
```
Shvandre: Oh, spotted a typo here. 
s0 is 0 or -1, and code is s1, data is at s2 etc

src/stdlib/stdlib/std/send.tact
novusnota: > let them be in a column, not a row

Sure!

> If it is crucial, I can reverse them

Not at all, that were just my initial thoughts. I'll think some more :)

anton-trunov (changes requested):
Super-awesome stuff 🎸 Let's update the Jetton benchmarks

src/stdlib/stdlib/std/context.tact
i582: And please, use `//` instead of `/**/`

novusnota: > And please, use `//` instead of `/**/`

ofc

src/test/benchmarks/jetton/results.json
```diff
@@ -48,11 +48,17 @@
       "burn": "17686",
       "discovery": "11039"
     },
+    {
+      "label": "Optimized send()",
```
anton-trunov:
```suggestion
      "label": "1.5.3 with optimized send()",
```

anton-trunov:
I think all those inline comments make the code harder to read. We should
- review the functional equivalence of the old and new code
- make only high-level comments only explaining the approach, if needed

novusnota:
> I think all those inline comments make the code harder to read. We should
> 
> * review the functional equivalence of the old and new code
> * make only high-level comments only explaining the approach, if needed

I've edited the `context.tact` part, see: https://github.com/tact-lang/tact/pull/1650/commits/736f532b4d886eda52b1479c258c846f5dd3606d. Feel free to comment on that. During this hour I'll edit the `send.tact` part and address possible comments for `context.tact`.

By the way, we also need a note on `Context.readForwardFee()` now working only for the workchains, not the masterchain! I'll add that in.

novusnota:
> By the way, we also need a note on Context.readForwardFee() now working only for the workchains, not the masterchain! I'll add that in.

UPD: Value of `first_frac` is the same for masterchain and other workchains as of now, so I was technically wrong there. For now.

---

Now that I think about it, we're (potentially) making a breaking change here compared to the non-asm version of the function: https://github.com/tact-lang/tact/pull/1650#issuecomment-2627883312

Instead of adding the `ZERO GETORIGINALFWDFEE` by the end of `Context.readForwardFee()`, we should probably keep the `fwd_fee` value that it gives, and only then apply the `getOriginalFwdFee()` function on top.

Otherwise:
1. [`getOriginalFwdFee()`](https://docs.tact-lang.org/ref/core-advanced/#getoriginalfwdfee) is rendered useless, and the example given there gives higher fees than before
2. `Context.readForwardFee()` returns not the `fwd_fee * 3 / 2` as it used to do (for whatever reason), but the `(fwd_fee * 2^16) / first_frac`, where first_frac is a value listed in config param 25 of the Blockchain: https://tonscan.org/config#25.

P.S.: But I may be totally off here, so I'll write some tests :)

Shvandre:
I actually left previous behaviour

novusnota:
> I actually left previous behaviour

Yeah, and I overlooked it in the previous PR. The behavior prior to all asm changes was:

```tact
extends fun readForwardFee(self: Context): Int {
    let sc: Slice = self.raw;
    sc.loadAddress(); // Skip destination
    sc.loadCoins(); // Skip value
    sc.skipBits(1); // Skip extra currency collection
    sc.loadCoins(); // Skip ihr_fee
    return (sc.loadCoins() * 3) / 2;
}
```

Which, roughly speaking, was producing halved result of the current iteration of `.readForwardFee()`. Or at least it seems that way to me, assuming TON Docs are correct.

anton-trunov:
@novusnota can you please suggest tests to see if your considerations are taken into account here

novusnota:
Hmm, I did some local checks and @Shvandre is right, things are still ok. Sorry everyone, false alarm from me :):

That said, I'll add e2e tests to encode the behavior

src/stdlib/stdlib/std/send.tact
novusnota: I think it's better now, take a look

src/test/benchmarks/jetton/results.json
```diff
@@ -48,11 +48,17 @@
...
+      "discovery": "10093"
+    },
     {
       "label": "master",
-      "transfer": "25860",
-      "burn": "17686",
-      "discovery": "11039"
+      "transfer": "21464",
+      "burn": "16029",
+      "discovery": "10093"
```
anton-trunov:
```suggestion
      "transfer": "21402",
      "burn": "16029",
      "discovery": "10085"
```

anton-trunov: From my local tests

anton-trunov (approved):

PR #1652, opened 2025-01-31
feat(interpreter): method calls

### LINKED ISSUE #1646: Implement method calls in interpreter
It should be fairly similar to static calls, but with the extra check that at the end the self argument should be assigned if the method call is a mutates function.

jeshecdom:
## Issue

Closes #1646.

Code is ready, but I am adding tests.

## Checklist

### DIFF
--- a/src/optimizer/interpreter.ts
+++ b/src/optimizer/interpreter.ts
@@ -20,7 +20,12 @@ import {
     hasStaticFunction,
 } from "../types/resolveDescriptors";
 import { getExpType } from "../types/resolveExpression";
-import { showValue, TypeRef } from "../types/types";
+import {
+    FunctionDescription,
+    showValue,
+    TypeDescription,
+    TypeRef,
+} from "../types/types";
 import { sha256_sync } from "@ton/crypto";
 import { defaultParser, getParser, Parser } from "../grammar/grammar";
 import { dummySrcInfo, SrcInfo } from "../grammar";
@@ -31,6 +36,7 @@ import {
     getAstFactory,
     idText,
     isSelfId,
+    tryExtractPath,
 } from "../ast/ast-helpers";
 import { divFloor, modFloor } from "./util";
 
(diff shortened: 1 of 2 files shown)


src/optimizer/interpreter.ts
```diff
@@ -933,11 +939,180 @@ export class Interpreter {
...
-                    `calls of ${idTextErr(ast.method)} are not supported at this moment`,
-                    ast.loc,
+            default: {
+                const selfT = getExpType(this.context, ast.self);
+                if (selfT.kind !== "ref") {
+                    throwNonFatalErrorConstEval(
+                        `currently, method calls only support reference type methods`,
+                        ast.loc,
+                    );
+                }
```
anton-trunov, line 949: why not support any sensible type and let the typechecker reject some programs?

src/optimizer/interpreter.ts
```diff
@@ -933,11 +939,180 @@ export class Interpreter {
...
+                        break;
+                    case "function_decl":
+                        throwNonFatalErrorConstEval(
+                            `${idTextErr(ast.method)} cannot be interpreted because it does not have a body`,
+                            ast.loc,
+                        );
+                        break;
+                    case "native_function_decl":
+                        throwNonFatalErrorConstEval(
+                            "native function calls are currently not supported",
```
anton-trunov, line 980: most likely will never be supported

jeshecdom: I'll change the message to "native functions are not supported".

src/optimizer/interpreter.ts
jeshecdom: Typechecking is actually not executed before interpretation in some parts of the compiler, like constant initialization. This causes problems. I just opened an issue #1655.

jeshecdom:
Requires PR #1678 to be merged into main for new tests to succeed.

src/optimizer/interpreter.ts
novusnota: > I'll change the message to "native functions are not supported".

I'd just remove the `currently`: "native function calls are not supported".

Otherwise, it may seem as if we don't support them at all (outside the interpreter) :)

anton-trunov:
Closing this for now, feel free to fix the merge conflicts and finish it after we release Tact 1.6

PR #1678, opened 2025-02-03
refactor(pipeline): move constant evaluation after typechecking

### LINKED ISSUE #1655: Constant initialization must be moved after typechecking
PR #1652 introduces the ability to execute method calls, but it is now blocked due to the following issue caused by early constant initialization. In the following Tact code:
```
const A: Int = callIncrement(10);

fun callIncrement(v: Int): Int {
    return v.incr();   // Execution fails here because v does not have yet a type registered in CompilerContext
}                      // and it is required in order to obtain function incr assigned to reference type Int.

extends fun incr(self: Int): Int {
    self += 1;
    return self;
}
```
Since constant initialization is carried out **before** function `callIncrement` is type-checked, the indicated line causes an internal compiler error because variable `v` still does not have a registered type in `CompilerContext`:
```
Error: [INTERNAL COMPILER ERROR]: Expression 1784 not found
    at throwInternalCompilerError (/home/jesus/tact/tact/src/error/errors.ts:66:11)
    at getExpType (/home/jesus/tact/tact/src/types/resolveExpression.ts:37:35)
    at Interpreter.interpretMethodCall (/home/jesus/tact/tact/src/optimizer/interpreter.ts:943:41)
    at Interpreter.interpretExpression (/home/jesus/tact/tact/src/optimizer/interpreter.ts:851:29)
.....
```
The same happens if we replace variable `v` with a number, for example:
```
fun callIncrement(v: Int): Int {
    return 10.incr();   // Execution fails here because 10 does not have yet a type registered in CompilerContext
}
```
Constant initialization must be moved after typechecking.

jeshecdom:
## Issue

Closes #1655.

Moves constant, default constract field and default struct field evaluation after typechecking to ensure that any function called during initialization has already typechecked.
 
This PR is required for PR #1652.

## Checklist


src/types/resolveDescriptors.spec.ts
```diff
@@ -31,6 +32,7 @@ describe("resolveDescriptors", () => {
             ctx = featureEnable(ctx, "external");
             ctx = resolveDescriptors(ctx, Ast);
             ctx = resolveSignatures(ctx, Ast);
+            evaluateDeclarationsInitializers(ctx, Ast);
```
anton-trunov: Is this necessary here? What happens if you don't call `evaluateDeclarationsInitializers` here and below?

src/storage/resolveAllocation.spec.ts
```diff
@@ -79,6 +80,7 @@ describe("resolveAllocation", () => {
         ctx = resolveSignatures(ctx, ast);
         ctx = resolveStatements(ctx, ast);
         ctx = resolveAllocations(ctx);
+        evaluateDeclarationsInitializers(ctx, ast);
```
anton-trunov: Is this necessary here? What happens if you don't call `evaluateDeclarationsInitializers`?

anton-trunov (commented):
What about evaluating message opcodes and getters' method_id?

src/types/resolveDescriptors.spec.ts
jeshecdom: `evaluateDeclarationsInitializers` is now the last step in `precompile`. Function `evaluateDeclarationsInitializers` contains the logic that was previously in `resolveDescriptors` for evaluating constants, default fields in contracts and structs. If the function is not added, some tests will fail because they are expecting that logic that was part of `resolveDescriptors`.

src/storage/resolveAllocation.spec.ts
jeshecdom: Some tests will fail because they are expecting the logic that was in `resolveDescriptors`.

jeshecdom:
> What about evaluating message opcodes and getters' method_id?

Mmmm... good point. I did not consider that. I see that there are several other places in the typechecker where `evalConstantExpression` is called. I will need to check them in more detail.

src/types/resolveDescriptors.spec.ts
anton-trunov: could you provide some examples of failing tests?

src/storage/resolveAllocation.spec.ts
anton-trunov: could you provide some examples of failing tests?

src/types/resolveDescriptors.spec.ts
jeshecdom: You were correct. It was just a snapshot that now has "undefined" in "value". Similar for the other test in the other comment. See below comment for an example.

src/storage/resolveAllocation.spec.ts
jeshecdom: You were correct. It was a snapshot. After removing that line. I get: 
```
 FAIL  src/storage/resolveAllocation.spec.ts
  ● resolveAllocation › should write program

    expect(received).toMatchSnapshot()

    Snapshot name: `resolveAllocation should write program 1`

    - Snapshot  - 7
    + Received  + 1

    @@ -2184,17 +2184,11 @@
                    "kind": "type_id",
                    "loc": {},
                    "text": "Int",
                  },
                },
    -           "default": {
    -             "base": 10,
    -             "id": 90,
    -             "kind": "number",
    -             "loc": {},
    -             "value": 0n,
    -           },
    +           "default": undefined,
                "index": 0,
                "loc": {},
                "name": "v",
                "type": {
                  "kind": "ref",

      82 |         ctx = resolveAllocations(ctx);
      83 |         //evaluateDeclarationsInitializers(ctx, ast);
    > 84 |         expect(getAllocations(ctx)).toMatchSnapshot();
         |                                     ^
      85 |     });
      86 | });
      87 |

      at Object.toMatchSnapshot (src/storage/resolveAllocation.spec.ts:84:37)
```
The only difference in the above is that "default" is now undefined. So, the line is actually not necessary.

src/pipeline/precompile.ts
```diff
@@ -42,6 +43,9 @@ export function precompile(
     // This creates allocations for all defined types
     ctx = resolveAllocations(ctx);
 
+    // Evaluate all constants, default contract fields and default struct fields
+    evaluateDeclarationsInitializers(ctx, ast);
```
anton-trunov: btw, why this does not use the common `ctx = FOO(ctx)` pattern? (I'm not saying we should have that pattern everywhere, just that your code is written in a different style)

anton-trunov (changes requested):
merge conflicts

anton-trunov:
merge conflicts

src/storage/__snapshots__/resolveAllocation.spec.ts.snap
```diff
@@ -2190,13 +2190,7 @@ exports[`resolveAllocation should write program 1`] = `
...
             },
           },
-          "default": {
-            "base": 10,
-            "id": 89,
-            "kind": "number",
-            "loc": {},
-            "value": 0n,
-          },
+          "default": undefined,
```
anton-trunov, line 2193: why did this change?

anton-trunov:
LGTM (modulo one question)

Let's also wait for a feedback from @i582

src/storage/__snapshots__/resolveAllocation.spec.ts.snap
jeshecdom: The `default` field gets populated only by calling `evalComptimeExpressions`. Since the test `resolveAllocation.spec.ts` does not call `evalComptimeExpressions`, the field remains in `undefined` state. It would get populated if adding `evalComptimeExpressions` after the line `ctx = resolveStatements(ctx);` in `resolveAllocation.spec.ts`.

i582 (approved):
Looks good

PR #1683, opened 2025-02-03
refactor: parser parses imports

verytactical:
## Issue

Here we deduplicate `Source`, and start parsing relative import paths directly in parser.

Towards #1396.

## Checklist


src/grammar/next/grammar.spec.ts
```diff
@@ -26,8 +28,117 @@ describe("grammar", () => {
...
+                            stepsUp: 0,
+                        },
+                    },
+                },
+            ],
+        });
+    });
+
+    it("should detect .fc imports", () => {
+        expect(parse('import "./import.fc";')).toMatchObject({
```
i582, line 129: We also need a test for `.func` extension

src/grammar/prev/grammar.ts
```diff
@@ -115,16 +117,65 @@ semantics.addOperation<A.AstNode>("astOfModule", {
...
+function parseImportString(
+    importText: string,
+    loc: SrcInfo,
+): A.SourceReference {
+    if (importText.endsWith("/")) {
+        err().noFolderImports()(loc);
+    }
+
+    if (importText.includes("\\")) {
+        err().importWithBackslash()(loc);
```
i582, line 151: In `grammar/next` you additionally replace `\\` with `/`:
```typescript
importText = importText.replace(/\\/g, "/");
```

src/imports/path.ts
```diff
@@ -0,0 +1,82 @@
...
+ */
+export const fromString = (raw: string): RelativePath => {
+    return raw.split("/").map(parseSegment).reduce(appendPath, emptyPath);
+};
+
+/**
+ * Convert RelativePath to string.
+ */
+export const asString = ({ stepsUp, segments }: RelativePath): string => {
+    return [...new Array(stepsUp).fill(0).map(() => ".."), ...segments].join(
```
i582: Following code may be more readable
```typescript
if (stepsUp === 0) return segments.join("/")
return "../".repeat(stepsUp) + "/" + segments.join("/")
```

i582: Also for an empty path we may want to return `.`, as stated in:
```typescript
**
 * Empty path, equivalent to "."
 */
export const emptyPath = RelativePath(0, []);
```

src/ast/ast.ts
```diff
@@ -547,6 +549,13 @@ export type AstBoolean = {
     loc: SrcInfo;
 };
 
+// Reference to source file
+export type SourceReference = {
```
anton-trunov: Let's not use the word "reference" here: it's too overloaded. `ImportPath`?

src/ast/ast.ts
```diff
@@ -547,6 +549,13 @@ export type AstBoolean = {
...
+    readonly path: RelativePath;
+    readonly type: "stdlib" | "relative";
```
anton-trunov: 1. Maybe it should be

```ts
readonly root: "stdlib" | "project";
```

?

2. It also feels like an overlap with `ItemOrigin`.

anton-trunov (changes requested):

src/grammar/next/grammar.spec.ts
anton-trunov, line 1: stdlib import tests are missing

anton-trunov (changes requested):

anton-trunov (approved):

PR #1704, opened 2025-02-04
fix(test): allow generate deep (really huge) Ast Expressions with custom weighted AST childrens in nodes

### LINKED ISSUE #1662: Make fuzzer uniform
Currently fuzzer generates ASTs that are heavily skewed: a lot of `null`, `boolean` and (to the less degree) other leaf nodes, and very few nodes that are actually interesting.

Research into generation of more uniform/interesting ASTs.

Example with collection of node statistics [can be found here](https://github.com/tact-lang/tact/commit/021d46318bab3a8eb2cd5f1ddea3b8804de2c9fa).

xpyctumo:
AST expressions will be generated with deeper AST trees, however this has a direct impact on performance, which is the reason I reduced the `numRun` value in test.

Some fixes:
- duplicate AST are no longer generated
- added a small chance of generating huge numbers (**256), but still prioritize readable numbers (this gave a small performance boost 🚀 )
- Reworked depth and implemented `fc.letrec` instead `fc.memo`. Changed depth [size](https://fast-check.dev/docs/configuration/larger-entries-by-default/#depth-size-explained) of AST expression by default to +1 (which be 3 - between small and medium). It's optimal to run in CI right now
- Fixed a major glitch with entering infinity recursion 
- Every AST expression child now have some weight. We can provide custom weights to change fuzzer rules.

But we sacrificed typification in some places...

## Issue

Closes #1662 

## Checklist

### DIFF
--- a/src/ast/random.infra.ts
+++ b/src/ast/random.infra.ts
@@ -80,75 +80,17 @@ function randomAstString(): fc.Arbitrary<A.AstString> {
 }
 
 function randomAstNumber(): fc.Arbitrary<A.AstNumber> {
-    const values = [
-        ...Array.from({ length: 10 }, (_, i) => [0n, BigInt(i)]).flat(),
-        ...Array.from({ length: 256 }, (_, i) => 1n ** BigInt(i)),
-    ];
-
     return dummyAstNode(
         fc.record({
             kind: fc.constant("number"),
             base: fc.constantFrom(2, 8, 10, 16),
-            value: fc.oneof(...values.map((value) => fc.constant(value))),
-        }),
-    );
-}
-
-function randomAstOpUnary(
-    operand: fc.Arbitrary<A.AstExpression>,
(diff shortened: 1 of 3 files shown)


src/ast/random.infra.ts
```diff
@@ -186,158 +128,163 @@ function randomAstNull(): fc.Arbitrary<A.AstNull> {
...
-                randomAstStaticCall(subExpr()),
-                randomAstStructInstance(
-                    randomAstStructFieldInitializer(subExpr()),
+    return fc.letrec((tie) => ({
+        literal: fc.oneof(
+            { arbitrary: randomAstNumber(), weight: 100 },
+            { arbitrary: randomAstBoolean(), weight: 20 },
+            { arbitrary: randomAstNull(), weight: 1 },
+        ),
+        methodCall: dummyAstNode(
```
i582, line 140: Why do we need to inline all these constructors?

PR #1714, opened 2025-02-05
refactor(interpreter): make non-top level methods in interpreter private

### LINKED ISSUE #1712: Refactor: make intermediate methods in interpreter private
This is a follow up to PR #1621. 

Make private all non-top level method accesses to the interpreter, so that external processes cannot bypass the stack overflow check. In other words, the only public methods should be `interpretModuleItem`, `interpretStatement` and `interpretExpression`, since those are the methods that check for stack overflows.

jeshecdom:
## Issue

Closes #1712.

## Checklist


src/optimizer/constEval.ts
```diff
@@ -138,7 +134,7 @@ export const getOptimizer = (util: AstUtil) => {
         switch (ast.kind) {
             case "id":
                 try {
-                    return interpreter.interpretName(ast);
+                    return interpreter.interpretExpression(ast);
```
i582, line 137: Should we include these changes in this PR? After all, it's about changing visibility of methods to private

jeshecdom: `interpretName` is now private in interpreter. We should only call `interpretExpression`, or `interpretStatement` or `interpretModuleItem` from outside the interpreter.

i582: Ah, I see

i582 (approved):

verytactical (approved):

PR #1737, opened 2025-02-07
feat: update error message for reserved `self` function argument

### LINKED ISSUE #1735: Error `Parameter name "self" is reserved` should suggest `extends` modifier for function
Example:
```
asm(other self) fun testFunc(self: Int, other: String) { 2 BLKDROP }
```

Gusarich:
## Issue

Closes #1735.

## Checklist


src/types/resolveDescriptors.ts
```diff
@@ -844,7 +844,7 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
         for (const param of params) {
             if (isSelfId(param.name)) {
                 throwCompilationError(
-                    'Parameter name "self" is reserved',
+                    'Parameter name "self" is reserved, consider using "extends" function modifier',
```
i582: I think we should additionally check that `self` is the first parameter

novusnota:
```suggestion
                    'Parameter name "self" is reserved for functions with "extends" modifier',
```

Same idea, but now we make it clear that only functions with "extends" modifier, i.e. extension functions, can use `self` as the name of the first parameter

Gusarich: You mean we should not throw this error if `self` occurs not as first parameter?

i582: We need to add this extra note only for the first `self` parameter

Gusarich: Have a look please!

Gusarich: Have a look please!

i582 (dismissed):

dev-docs/CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Better error message for `extend function without parameters` error: PR [#1624](https://github.com/tact-lang/tact/pull/1624)
 - Don't generate `lazy_deployment_completed` by default: PR [#1717](https://github.com/tact-lang/tact/pull/1717)
 - Optimized `emptyCell()` and `emptySlice()` functions: PR [#1696](https://github.com/tact-lang/tact/pull/1696)
+- Error message for reserved `self` function argument now suggests using `extends` function modifier: PR [#1737](https://github.com/tact-lang/tact/pull/1737)
```
```suggestion
- Error message for reserved `self` function parameter now suggests using `extends` function modifier: PR [#1737](https://github.com/tact-lang/tact/pull/1737)
```

src/types/resolveDescriptors.ts
```diff
@@ -841,12 +841,19 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
...
             if (isSelfId(param.name)) {
-                throwCompilationError(
-                    'Parameter name "self" is reserved',
-                    param.loc,
-                );
+                if (i === 0) {
+                    throwCompilationError(
+                        'Parameter name "self" is reserved for functions with "extends" modifier',
+                        param.loc,
+                    );
```
anton-trunov: why do we check this at each iteration? the proper way to do it is to split the head and the tail of the parameters list and process those separately

anton-trunov (changes requested):

src/types/resolveDescriptors.ts
Gusarich: have a look please

src/types/resolveDescriptors.ts
anton-trunov, line 1: I think this will produce a confusing error message
```fun
extends fun foo(self: Int, self: String) { }
```

> Parameter name "self" is reserved for functions with "extends" modifier

Here it *is* an `extends` function, it's just we have duplicate parameter names

src/types/resolveDescriptors.ts
```diff
@@ -839,22 +839,44 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
             );
         }
 
-        // Check parameter names
         const exNames: Set<string> = new Set();
-        for (const param of params) {
-            if (isSelfId(param.name)) {
+
+        if (params.length > 0) {
+            const [firstParam, ...restParams] = params;
```
anton-trunov:
```suggestion
        const [head, ...restParams] = params;
        if (!isUndefined(head)) {  // from ../utils/array
```

anton-trunov (changes requested):

src/types/__snapshots__/resolveDescriptors.spec.ts.snap
```diff
@@ -495,6 +495,24 @@ exports[`resolveDescriptors should fail descriptors for native-fun-getter 1`] =
...
+"<unknown>:3:30: Parameter name "self" is reserved for functions with "extends" modifier
+  2 | 
+> 3 | asm(other self) fun testFunc(self: Int, other: Int) { 2 BLKDROP }
+                                   ^~~~~~~~~
+  4 | 
+"
+`;
+
+exports[`resolveDescriptors should fail descriptors for not-extend-function-self-param-2 1`] = `
+"<unknown>:4:28: Parameter name "self" is reserved
```
anton-trunov: confusing error message

anton-trunov (changes requested):

src/types/__snapshots__/resolveDescriptors.spec.ts.snap
Gusarich: have a look now please!

anton-trunov (approved):

PR #1751, opened 2025-02-08
fix: remove ast tools

### LINKED ISSUE #1750: Remove legacy AST tools
These are extremely annoying when changing AST in any way: every node should be manually handled in several files, every time in a slightly different style.

AST tooling must be generated, but before we even can get there, there's a lot of things to move around, and every time these tools have to be updated.

verytactical:
## Issue

Closes #1750.

## Checklist


novusnota (changes requested):
There's an entry in the CHANGELOG from when those tools were first added: https://github.com/tact-lang/tact/blob/main/dev-docs/CHANGELOG.md#added-3. Do we want to add a new entry about their removal?

novusnota (approved):

anton-trunov (approved):

PR #1752, opened 2025-02-08
fix: fail on any error on Windows

### LINKED ISSUE #1341: tact-template can fail in CI on Windows but CI is still green
See https://github.com/tact-lang/tact/actions/runs/12773479327/job/35605455026?pr=1339

This happens because GitHub Actions on Windows do not fail if an intermediate command in a script fails and it just executes the script further. (In that particular run it continues testing successfully because the Tact build artifacts are committed into the tact-template repo)

It seems to me we should split the `tact.yml` file into `tact.unix-like.yml` and `tact.windows.yml` and maintain two GH Action files: the Windows scripts would have something like `$ErrorActionPreference = 'Stop'` command before each step. (We should not just switch to WSL or use Unix-like shells on Windows because we need to make sure everything works on the most simple standard setup. We actually might want to *add* WSL, etc. but not replace Powershell with it). This split would also allow us to test CLI on Windows, which we currently do not do.

Btw, related issue: #318

verytactical:
## Issue

Closes #1341.

## Checklist


.github/workflows/tact.yml
```diff
@@ -76,52 +76,33 @@ jobs:
...
-        uses: actions/checkout@v4
-
-      - name: Setup Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v4
-        with:
-          node-version: ${{ matrix.node-version }}
-          cache: "yarn"
-
-      - name: Install dependencies
+      - uses: ../actions/test
```
novusnota, line 82: The previous path was correct: `./.github/actions/test`. You just need to make the `test/` folder, and move the current `test.yml` as `action.yml` there.

Such that there would be a `.github/actions/test/action.yml` file with the contents of this `tact.yml` file

novusnota (changes requested):
See the suggested fix in a prior comment. UPD: All good now.

novusnota (approved):

anton-trunov (approved):

PR #1753, opened 2025-02-08
feat: Improve error messages with Jaro-Winkler similarity matching

### LINKED ISSUE #1697: Improving Compiler Error Messages with Similarity Matching
#### **Problem Statement**

Currently, compiler error messages often fail to provide meaningful suggestions when encountering an unknown function. For example, consider the following compilation error:

```
Compilation error: sources/contract.tact:38:9: Map function "s1et" not found
Line 38, col 9:
  37 |     fun testReturn1(m: map<Int, Int>): map<Int, Int> {
> 38 |         m.s1et(3, 3); // OK
               ^~~~~~~~~~~~
  39 |         if (WHATEVER == 0) {
```

The error message merely states that `s1et` is not found, but it does not provide any helpful suggestions about what the user might have meant.

#### **Proposed Solution**

The compiler should intelligently suggest possible alternatives based on similarity matching algorithms, such as:

1. **Word Match** – Finds exact matches or matches based on word components.
2. **Partial Match** – Identifies functions that share common substrings or structures.
3. **Levenshtein Distance** – Detects likely typos by measuring how different two strings are.

Using these methods, the compiler could enhance its error message by suggesting possible corrections. For example:

```
Compilation error: sources/contract.tact:38:9: Map function "s1et" not found
Line 38, col 9:
  37 |     fun testReturn1(m: map<Int, Int>): map<Int, Int> {
> 38 |         m.s1et(3, 3); // OK
               ^~~~~~~~~~~~
  39 |         if (WHATEVER == 0) {
Did you mean:
  - set(self: map<K, V>, key: K, val: V)
  - get(self: map<K, V>, key: K): V?
```

This approach makes debugging significantly easier, particularly for beginners, by reducing the guesswork involved in fixing syntax errors.

#### **Impact**

By implementing similarity-based suggestions, we can:

- Improve the user experience of AI-powered development tools.
- Reduce frustration for novice programmers.
- Enhance the efficiency of error resolution in the compiler.

This methodology ensures that even in cases of typos or incorrect function names, developers receive clear and relevant guidance, ultimately streamlining the development process.

Gusarich:
## Issue

Towards #1697.

## Progress

I first tried implementing Levenshtein distance but couldn't tune it for good results. Jaro-Winkler, on the other hand, produced good results with minimal parameter tuning, so I think it's a better choice. Another available option is to combine those to achieve potentially even better results.

- [x] Implement basic Jaro-Winkler similarity  
- [x] Tune parameters to improve matching accuracy  
- [x] Add extensive test coverage  
- [ ] Integrate into type checking to enhance error messages

## Checklist

### DIFF
--- a/src/utils/suggestNames.ts
+++ b/src/utils/suggestNames.ts
@@ -1,4 +1,4 @@
-export interface FunctionSignature {
+export interface Signature {
     name: string;
     signature: string;
 }
@@ -151,7 +151,7 @@ function computeCompositeScore(
 }
 
 /**
- * Suggests function signatures similar to the unknown function name.
+ * Suggests signatures similar to the unknown name.
  *
  * - Returns [] if the unknown query is empty.
  * - Computes a composite score for each candidate.
@@ -162,10 +162,10 @@ function computeCompositeScore(
  * - Finally, returns only candidates whose score is within TOLERANCE of the best score.
  * - Limits the final returned list to at most 3 suggestions.
  */
-export function suggestFunctions(
+export function suggestSignatures(
     unknown: string,
-    functions: FunctionSignature[],
-): FunctionSignature[] {
+    signatures: Signature[],
+): Signature[] {
     if (unknown.trim() === "") return [];
 
     const ABS_MIN_SCORE = 0.7;
@@ -173,7 +173,7 @@ export function suggestFunctions(
     const TOLERANCE = 0.05;
 
     // Score candidates.
-    const scoredCandidates = functions.map((f) => {
+    const scoredCandidates = signatures.map((f) => {
         const { score, normQuery, normCandidate, contains } =
             computeCompositeScore(unknown, f.name);
         return { ...f, score, normQuery, normCandidate, contains };
--- a/cspell.json
+++ b/cspell.json
@@ -56,6 +56,7 @@
     "src/test/exit-codes/contracts/compute-phase-errors.tact",
     "src/test/e2e-emulated/map-tests/build",
     "src/test/e2e-emulated/map-tests/map-properties-key-value-types.ts",
+    "src/utils/suggestNames.spec.ts",
     "/docs"
   ]
 }
--- a/spell/cspell-list.txt
+++ b/spell/cspell-list.txt
@@ -57,6 +57,7 @@ Ints
 ipfs
 IPFS
 ipld
+Jaro
 Jesús
 jetton
 Jetton
@@ -163,6 +164,7 @@ varint
 varuint
 viiii
 vogons
+Winkler
 workchain
 workchains
 xffff
--- a/src/utils/__snapshots__/suggestFunctions.spec.ts.snap
+++ b/src/utils/__snapshots__/suggestFunctions.spec.ts.snap
@@ -1,0 +1,602 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`suggestFunctions Snapshot for query "  storeInt  " - Query with leading/trailing whitespace; should match storeInt 1`] = `
+[
+  {
+    "name": "storeInt",
+    "signature": "@name(store_int) extends native storeInt(self: Builder, value: Int, bits: Int): Builder;",
+  },
+]
+`;
+
+exports[`suggestFunctions Snapshot for query "AS_slice" - mixed case with underscore in asSlice 1`] = `
+[
+  {
+    "name": "asSlice",
+    "signature": "@name(__tact_address_to_slice) extends native asSlice(self: Address): Slice;",
+  },
+]
+`;
+
+exports[`suggestFunctions Snapshot for query "GAS_conumed" - Mixed case with underscore; should now return only gasConsumed 1`] = `
+[
+  {
+    "name": "gasConsumed",
+    "signature": "asm fun gasConsumed(): Int { GASCONSUMED }",
+  },
+]
+`;
+
+exports[`suggestFunctions Snapshot for query "GAS_conumed" - underscore and missing letter in gasConsumed 1`] = `
+[
+  {
+    "name": "gasConsumed",
+    "signature": "asm fun gasConsumed(): Int { GASCONSUMED }",
+  },
+]
+`;
+
+exports[`suggestFunctions Snapshot for query "STOREINT" - All uppercase query; should match storeInt 1`] = `
+[
+  {
+    "name": "storeInt",
+    "signature": "@name(store_int) extends native storeInt(self: Builder, value: Int, bits: Int): Builder;",
+  },
+]
+`;
+
+exports[`suggestFunctions Snapshot for query "Stroe_Int" - transposition and underscore in storeInt 1`] = `
+[
+  {
+    "name": "storeInt",
+    "signature": "@name(store_int) extends native storeInt(self: Builder, value: Int, bits: Int): Builder;",
+  },
+]
+`;
+
+exports[`suggestFunctions Snapshot for query "asSlic" - missing letter in asSlice 1`] = `
+[
+  {
+    "name": "asSlice",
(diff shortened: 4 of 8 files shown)


Gusarich:
@anton-trunov @verytactical @skulidropek Could you please take a look at the current outputs (test snapshots) and provide some feedback while I work on the integration? We might need to fine-tune something or add more interesting test cases.

Gusarich:
I also think it makes sense to integrate it into all kinds of `not found` errors, not just for functions. This includes types, contracts, structs, etc.

src/utils/suggestNames.ts
```diff
@@ -0,0 +1,226 @@
...
+ * - Returns [] if the unknown query is empty.
+ * - Computes a composite score for each candidate.
+ * - Discards candidates with a score below MIN_SCORE.
+ * - If any candidate's normalized name exactly equals the normalized query, returns only those.
+ * - If the best candidate's score is below an absolute threshold (ABS_MIN_SCORE), returns [].
+ * - Additionally, if the normalized query is at least 3 characters long, restrict to candidates whose normalized form starts with the first 3 letters of the normalized query (if any do).
+ * - Finally, returns only candidates whose score is within TOLERANCE of the best score.
+ * - Limits the final returned list to at most 3 suggestions.
+ */
+export function suggestSignatures(
```
reverofevil, line 165: Probably the most interesting PR I've seen so far.

On a practical side of things though, could you please add a benchmark on all identifiers from standard library, randomly mutating them, and counting sum of their indices in suggestion list?

If we go this far, it's better to compare its performance against other edit distances objectively, in hard numbers.

The other consideration is that even though it's great as technology demo, we might also hit another problem with performance: this takes roughly length of unknown multiplied by a sum of lengths of signatures. If for popular dynamic programming approach to edit distance we can probably find an approach where one of DP axis is a trie (or even something more advanced like automaton), it might be more problematic in Jaro case.

Long things short, to get to any conclusions we'd need "quality" and performance benchmarks.

verytactical, line 165: Probably the most interesting PR I've seen so far.

On a practical side of things though, could you please add a benchmark on all identifiers from standard library, randomly mutating them, and counting sum of their indices in suggestion list?

If we go this far, it's better to compare its performance against other edit distances objectively, in hard numbers.

The other consideration is that even though it's great as technology demo, we might also hit another problem with performance: this takes roughly length of unknown multiplied by a sum of lengths of signatures. If for popular dynamic programming approach to edit distance we can probably find an (asymptotically) better algorithm where one of DP axes is a trie (or even something more advanced like automaton), it might be more problematic in Jaro case.

Long things short, to get to any conclusions we'd need "quality" and performance benchmarks.

skulidropek:
> @anton-trunov @verytactical @skulidropek Could you please take a look at the current outputs (test snapshots) and provide some feedback while I work on the integration? We might need to fine-tune something or add more interesting test cases.

```
user@arch ~/tact (1697-similiarity-matching-improvement) [30]> node bin/tact.js /home/user/tact-template/sources/contract.tact
💼 Compiling project contract.tact ...
Tact compilation failed
Error: ../tact-template/sources/contract.tact:18:9: Map function "s1et" not found
  17 |     fun testReturn1(m: map<Int, Int>): map<Int, Int> {
> 18 |         m.s1et(3, 3); // OK
               ^~~~~~~~~~~~
  19 |         return m;

    at throwCompilationError (/home/user/tact/dist/error/errors.js:58:11)
    at resolveCall (/home/user/tact/dist/types/resolveExpression.js:427:48)
    at resolveExpression (/home/user/tact/dist/types/resolveExpression.js:606:20)
    at processStatements (/home/user/tact/dist/types/resolveStatements.js:265:69)
    at processFunctionBody (/home/user/tact/dist/types/resolveStatements.js:509:17)
    at resolveStatements (/home/user/tact/dist/types/resolveStatements.js:645:23)
    at precompile (/home/user/tact/dist/pipeline/precompile.js:22:53)
    at build (/home/user/tact/dist/pipeline/build.js:74:43)
    at run (/home/user/tact/dist/cli/tact/index.js:220:47)
    at compile (/home/user/tact/dist/cli/tact/index.js:201:30)
```

I installed your version of the compiler, but I don't see that the compiler offers me possible changes?

skulidropek:
> I also think it makes sense to integrate it into all kinds of `not found` errors, not just for functions. This includes types, contracts, structs, etc.

Yes, this should be done for everything. I provided the functions as an example. This will greatly help users understand what the problem is. This will help not only users but also AI programmers to write better code.

I have a sample code in C# that does exactly what I described. Maybe it will help you.  [StringSimilarity C# Example](https://github.com/publicrust/rust-analyzer/blob/main/src/Utils/StringSimilarity.cs)

Gusarich:
@skulidropek It's not yet integrated. I meant just looking through https://github.com/tact-lang/tact/blob/cef7d9dc5bc29ced9f87d81a2539ee9d431a6d00/src/utils/__snapshots__/suggestNames.spec.ts.snap

skulidropek:
> @skulidropek It's not yet integrated. I meant just looking through https://github.com/tact-lang/tact/blob/cef7d9dc5bc29ced9f87d81a2539ee9d431a6d00/src/utils/__snapshots__/suggestNames.spec.ts.snap

Yes, great, everything seems to work as expected. I'll wait until you integrate this into the project itself and I'll test it again.

src/utils/suggestNames.ts
Gusarich: > and counting sum of their indices in suggestion list?

But what about the number of irrelevant suggestions? Most setups I tried either included too many irrelevant suggestions or failed to find the correct one. The second problem is easy to check (as you suggested), while the first required some manual review of the generated suggestions on my part to improve the final results.

skulidropek: > The other consideration is that even though it's great as technology demo, we might also hit another problem with performance: this takes roughly length of unknown multiplied by a sum of lengths of signatures.

Will it give a big load if it is called only on errors? Even if it becomes slower only on errors, we will improve the entry of newcomers into using Tact. In any case, the load will be only when outputting errors. The compiler itself worked quickly and will work quickly.

verytactical: > Will it give a big load

That's the whole point: we don't know :)

At least minimal benchmarking is necessary.

The discussed limits are something like ~1s worst-case for CLI, and ~15ms if we want to integrate it into IDE. That's while we have a single error. After we start emitting several errors budget gets proportionally smaller. Especially it's important if newcomers don't have modern computers. That cuts our time budget by a lot too.

verytactical: > too many irrelevant suggestions

They don't really matter if they don't push the relevant suggestion down the list. That's why "distance to the top of the list of suggestions" is the metric. If you want to be strict about this (Knuth-level strict), sum not the indices, but their squares.

verytactical: If we're talking about CLI, and not IDE autocompletion, users might not really want to read the long list. Benchmark becomes important here: we can draw a graph of "cutout distance" vs "quality", and choose the cutout distance so that quality doesn't suffer.

skulidropek: > If we're talking about CLI, and not IDE autocompletion, users might not really want to read the long list. Benchmark becomes important here: we can draw a graph of "cutout distance" vs "quality", and choose the cutout distance so that quality doesn't suffer.

I think it is enough to display the first 3-5 elements.
We must take into account that this is done not only for users, but also for AI programmers who require context (Like Cursor & Windsurf)

anton-trunov:
@Gusarich @verytactical Any progress here?

Gusarich:
> @Gusarich @verytactical Any progress here?

Maybe push it to 1.7? I think it's not the first priority

PR #1755, opened 2025-02-08
feat(stdlib): add `getPrecompiledGas` function

Adamos123421:
Missing GETPRECOMPILEDGAS TVM fucntion necessary for contract like USDT contract :
https://docs.ton.org/v3/documentation/tvm/changelog/tvm-upgrade-2024-04#opcodes-to-process-config-parameters
And new TVM fucntion UNPACKEDCONFIGTUPLE missing too:
https://docs.ton.org/v3/documentation/tvm/changelog/tvm-upgrade-2024-04#opcodes-to-work-with-new-c7-values

### DIFF
--- a/src/stdlib/stdlib/std/internal/contract.tact
+++ b/src/stdlib/stdlib/std/internal/contract.tact
@@ -346,7 +346,6 @@ asm fun myCode(): Cell { MYCODE }
 /// fucntion available in the TVM update 
 asm fun getPrecompiledGas(): Int { GETPRECOMPILEDGAS }
 
-asm fun getFeeConfigs(): Tuple {UNPACKEDCONFIGTUPLE}
 
 
 
--- a/src/stdlib/stdlib/std/internal/contract.tact
+++ b/src/stdlib/stdlib/std/internal/contract.tact
@@ -343,3 +343,13 @@ asm fun setSeed(seed: Int) { SETRAND }
 /// See: https://docs.tact-lang.org/ref/core-advanced#mycode
 ///
 asm fun myCode(): Cell { MYCODE }
+/// fucntion available in the TVM update 
+asm fun getPrecompiledGas(): Int { GETPRECOMPILEDGAS }
+
+asm fun getFeeConfigs(): Tuple {UNPACKEDCONFIGTUPLE}
+
+
+
+
+
+


src/stdlib/stdlib/std/internal/contract.tact
```diff
@@ -343,3 +343,12 @@ asm fun setSeed(seed: Int) { SETRAND }
 /// See: https://docs.tact-lang.org/ref/core-advanced#mycode
 ///
 asm fun myCode(): Cell { MYCODE }
+/// fucntion available in the TVM update 
+asm fun getPrecompiledGas(): Int { GETPRECOMPILEDGAS }
```
anton-trunov, line 347:
```suggestion

/// function available in the TVM update 
asm fun getPrecompiledGas(): Int? { GETPRECOMPILEDGAS }
```

src/stdlib/stdlib/std/internal/contract.tact
```diff
@@ -343,3 +343,12 @@ asm fun setSeed(seed: Int) { SETRAND }
...
+
+
+
+
+
+
+
```
anton-trunov, line 354:
```suggestion
```

anton-trunov (changes requested):

src/stdlib/stdlib/std/internal/contract.tact
novusnota: Adjusted your suggestion to `Int?`, because for non-precompiled contracts the `null` is returned instead of the gas usage

novusnota:
I'm not sure that we need this in the stdlib — it seems to be too specific. Will wait for the input from @Shvandre though

Adamos123421:
> I'm not sure that we need this in the stdlib — it seems to be too specific. Will wait for the input from @Shvandre though

It's used for stable coin contracts . check https://github.com/ton-blockchain/stablecoin-contract/blob/main/contracts/gas.fc
+ int jetton_wallet_gas_consumption = get_precompiled_gas_consumption();
line 70

Shvandre:
> > I'm not sure that we need this in the stdlib — it seems to be too specific. Will wait for the input from @Shvandre though
> 
> It's used for stable coin contracts . check https://github.com/ton-blockchain/stablecoin-contract/blob/main/contracts/gas.fc
> + int jetton_wallet_gas_consumption = get_precompiled_gas_consumption();
> line 70

Yeah, and it is the only place where this function is used. I think is someone need those instructions, one can use asm functions

anton-trunov:
Let's close this for now, we can reconsider in the future

Adamos123421:
By the way guys, I opened this bounty are you responsible for approving it ? https://github.com/ton-society/grants-and-bounties/issues/954

anton-trunov:
@Vodka2134156 nope, it's a different entity

Adamos123421:
> @Vodka2134156 nope, it's a different entity

Thnaks for reply , they stopped approving bounties for two months ago , I wanted to make some important contract that are madable in tact but still there are not in examples , like the wallets contract and nft contracts , simple distibutors , comprressed nfts , maybe liquid staking too. But I guess it won't get approved as they are not active anymore .

Adamos123421:
@anton-trunov I am working on implementing TON-DNS in tact can it be added to tact by examples when it's done ? Even though I doubt new beginners will benefit from it alot.

anton-trunov:
@Vodka2134156 There is already something on DNS in the standard library: https://docs.tact-lang.org/ref/stdlib-dns. Please have a look

delovoyhomie:
@Vodka2134156 go ahead! Do you have an approximate deadline?

PR #1758, opened 2025-02-10
fix: support `T` and `T?` methods with the same name

### LINKED ISSUE #1285: Methods for non-optional and optional types clash
The following snippet

```tact
extends fun equal(self: Int, other: Int): Bool {
    return self == other;
}

extends fun equal(self: Int?, other: Int?): Bool {
    if (self == null && other == null) { return true }
    if (self == null || other == null) { return false }
    return self!!.equal(other!!);
}
```

produces the following type error:

```
Error: test.tact:5:1: Function "equal" already exists in type "Int"
Line 5, col 1:
  4 | 
> 5 | extends fun equal(self: Int?, other: Int?): Bool {
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  6 |     if (self == null && other == null) { return true }
```

i582:
## Issue

Closes #1285.

## Checklist


src/abi/map.ts
```diff
@@ -353,7 +353,7 @@ export const MapFunctions: ReadonlyMap<string, AbiFunction> = new Map([
                 const [self] = args;
                 checkMapType(self, ref);
 
-                return { kind: "ref", name: "Cell", optional: true };
+                return { kind: "ref", name: "Cell", optional: false };
```
anton-trunov, line 357: so, this wouldn't fail at runtime?

```tact
contract Foo {
    get fun foo(): Slice {
        let x: map<Int, Int> = emptyMap();
        return x.asCell().asSlice();
    }
}
```

Gusarich (changes requested):
The snippet below throws an internal compiler error.
```tact
extends fun foo(self: Int) {
    return self * 2;
}

fun bar(x: Int?) {
    x.foo();
}
```

```
[INTERNAL COMPILER ERROR]: Type Int_optional not found

       96 |     const msg = `[INTERNAL COMPILER ERROR]: ${message}\nPlease report at https://github.com/tact-lang/tact/issues`;
       97 |     throw source === undefined
    >  98 |         ? new TactInternalCompilerError(msg)
          |           ^
       99 |         : new TactInternalCompilerError(
      100 |               `${locationStr(source)}\n${msg}\n${source.interval.getLineAndColumnMessage()}`,
      101 |               source,

      at throwInternalCompilerError (src/error/errors.ts:98:11)
      at getType (src/types/resolveDescriptors.ts:2113:35)
      at resolveCall (src/types/resolveExpression.ts:644:29)
      at resolveExpression (src/types/resolveExpression.ts:947:20)
      at processStatements (src/types/resolveStatements.ts:434:48)
      at processFunctionBody (src/types/resolveStatements.ts:763:17)
      at resolveStatements (src/types/resolveStatements.ts:803:19)
      at Object.<anonymous> (src/types/resolveStatements.spec.ts:24:36)
```

anton-trunov:
@i582 ping :)

i582:
> The snippet below throws an internal compiler error.
> 
> ```tact
> extends fun foo(self: Int) {
>     return self * 2;
> }
> 
> fun bar(x: Int?) {
>     x.foo();
> }
> ```

Fixed

src/abi/map.ts
i582: If I understand correctly, now we just generate the qualifier expression as is, so `mp.asCell()` becomes `mp` in the FunC code:

https://github.com/tact-lang/tact/blob/7b1dd5463113bfe9b678c145a0d7507eacfd288c/src/abi/map.ts#L374

Thus, `asCell` simply returns non optional `Cell` type

src/types/resolveDescriptors.ts
```diff
@@ -2080,9 +2097,23 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
 export function getType(
     ctx: CompilerContext,
     ident: A.AstId | A.AstTypeId | string,
+    optional?: boolean,
```
anton-trunov, line 2100: > Don't use optional parameters

:)

why do we actually want this change?

anton-trunov:
let's move this to 1.7

Issue #1736, opened 2025-02-07
Add an extension function `asAddress(self: Slice)` to the stdlib


Shvandre:
Idk, maybe slice should be checked if it is a valid address
Or it may be just "NOP"

novusnota:
Perhaps, we should add two: with and without a check. Like, `.asAddress()` won't check, but `.asValidAddress()` (or something) will. That's because we've previously removed lots of Address checks, but having some that user can tap in if they want seems beneficial to me.

Ideas of names for the latter:
* `.asAddressCheck()`
* `.asAddressAssert()` (akin to `!!` in name and spirit) — or even `.assertAddress()`
* `.asAddressSafe()`

Shvandre:
Great idea!
Maybe it's better to make `.asAddress` safe version, and `.asAddressUnsafe` - without any checks.

PR #1766, opened 2025-02-10
refactor(stdlib): rewrite `contractAddressExt`, `newAddress` and `Address.asSlice` functions to Tact; add `asAddressUnsafe` and `contractHash`

### LINKED ISSUE #1763: Rewrite `contractAddressExt` (`__tact_compute_contract_address`) in Tact

### LINKED ISSUE #1764: Rewrite `Address.asSlice` (`__tact_address_to_slice`) in Tact

### LINKED ISSUE #1765: Rewrite `newAddress` (`__tact_create_address`) in Tact

i582:
Screenshot 2025-02-13 at 16 00 12

## Issue

Closes #1763.
Closes #1764.
Closes #1765.
Towards #1736.
Towards #573.

## Checklist


src/stdlib/stdlib/std/internal/contract.tact
```diff
@@ -1,21 +1,38 @@
...
+        .storeUint(0, 1)
+        .storeInt(chain, 8)
+        .storeUint(hash, 256)
+        .endCell()
+        .asSlice()
+        .toAddressUnsafe();
+}
+
+asm extends fun toAddressUnsafe(self: Slice): Address { NOP }
+asm extends fun asSlice(self: Address): Slice { NOP }
```
Shvandre: Idk, does those NOPs presented in resulting code? Maybe we should define those by just curly braces, without NOP? (NOP actually consumes gas. Insignificant amount, but still consumes)

i582:
Some changes from #1804, let's wait for #1804 to be merged and merge master into this PR

src/stdlib/stdlib/std/internal/contract.tact
i582: Looks like yes:
Screenshot 2025-02-13 at 14 42 38

But currently, empty assembly functions result in the following FunC compilation error:

```
FunC compilation error: /Users/petrmakhnev/tact/src/test/benchmarks/contracts/output/address_Addresses.code.fc:51:4: error: string with assembler instruction expected
  """;
```

Let's create and issue for such cases

src/stdlib/stdlib/std/internal/contract.tact
```diff
@@ -1,21 +1,36 @@
...
+
+inline fun contractAddressExt(chain: Int, code: Cell, data: Cell): Address {
```
Shvandre, line 174: Lets optimize this function. It shouldn't use endCell().
Let's use HASHEXT_SHA256 to calculate hash of builder

i582: Done

src/stdlib/stdlib/std/internal/contract.tact
```diff
@@ -1,21 +1,60 @@
...
+        // depth_descriptors
+        .storeUint(code.depth(), 16)
+        .storeUint(data.depth(), 16)
+        // ref hashes
+        .storeUint(code.hash(), 256)
+        .storeUint(data.hash(), 256);
+    return builderSha256(builder);
+}
+
+asm fun builderSha256(b: Builder): Int {
```
novusnota: Do we need this as a new stdlib library function? Perhaps let's use the approach from: https://github.com/tact-lang/tact/pull/1626

novusnota: Besides, the `.toAddressUnsafe()` and `calculateAddressHash()` are introduced too. If we want to keep them in the stdlib, let's add an entry in the CHANGELOG.

If not, let's use the approach from https://github.com/tact-lang/tact/pull/1626

i582: Let's not add any extra FunC functions as we want to remove them for the native backend

novusnota: Ok, then what about naming `builderSha256()` like `sha256Builder()`? Or, even better, re-using the on-chain `sha256()` we have, but with converting the Builder to a Slice?

i582: >  but with converting the Builder to a Slice?

It's too expensive because of `ENDC`

i582: Renamed to `Builder.hash()`

anton-trunov:
merge conflicts

dev-docs/CHANGELOG.md
```diff
@@ -24,7 +24,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `nullChecks` config option to disable run-time null checks for the `!!` operator in order to save gas: PR [#1660](https://github.com/tact-lang/tact/pull/1660)
 - `loadVarInt16`, `loadVarUint16`, `loadVarInt32`, `loadVarUint32` methods for the `Slice` type: PR [#1667](https://github.com/tact-lang/tact/pull/1667)
 - New functions in stdlib from `stdlib.fc` and `math.fc`: `Builder.depth`, `Slice.skipLastBits`, `Slice.firstBits`, `Slice.lastBits`, `Slice.depth`, `Cell.computeDataSize`, `Slice.computeDataSize`, `Cell.depth`, `curLt`, `blockLt`, `setGasLimit`, `getSeed`, `setSeed`, `myCode`, `sign`, `divc`, `muldivc`, `mulShiftRight`, `mulShiftRightRound`, `mulShiftRightCeil`, `sqrt`: PR [#986](https://github.com/tact-lang/tact/pull/986)
+- New functions `toAddressUnsafe`, `calculateAddressHash`, `Builder.hash` in stdlib: PR [#1766](https://github.com/tact-lang/tact/pull/1766)
 - The `--output` CLI flag for specifying custom output directory in single-contract compilation: PR [#1793](https://github.com/tact-lang/tact/pull/1793)
+- New functions `toAddressUnsafe`, `calculateAddressHash`, `Builder.hash` in stdlib: PR [#1766](https://github.com/tact-lang/tact/pull/1766)
```
anton-trunov, line 28: looks like a duplicate line

i582:
> merge conflicts

Fixed

src/stdlib/stdlib/std/internal/cells.tact
```diff
@@ -163,6 +163,11 @@ asm extends fun refs(self: Builder): Int { BREFS }
 
 asm extends fun bits(self: Builder): Int { BBITS }
 
+asm extends fun hash(self: Builder): Int {
```
anton-trunov: we need correctness tests for this new method: the hash we compute here should have a certain relationship with `.hash()` for Cells / Slices

anton-trunov (changes requested):

src/stdlib/stdlib/std/internal/cells.tact
Gusarich: This one will calculate data hash, not cell hash. Exactly like `SHA256U` for slices.

I think we should either rename this method to something like `dataHash` or make it compute cell hash to match the logic of `.hash` for cells and slices.

novusnota:
Added docs, please take a look!

Renamed `calculateAddressHash()` to `contractHash()` for clarity and conformance to existing functions. It's not the `Address` yet only the `account_id` part of the address, hence the simplification of the name.

Also renamed `Slice.toAddressUnsafe` to `Slice.asAddressUnsafe` because:
* there is `map.asCell()` method that does similar type casting in spirit, i.e. NOP
* there is a mirror function `Address.asSlice()`, also a NOP
* issue 1736 originally discussed the `.asAddressUnsafe()` name

In general, I think at some point we might carefully split all the `asSomething()` casts and `toSomething()` conversions, where the former are no-ops and simply instruct the compiler to start viewing the underlying type differently, while the latter actually require some operations and transformations to be performed.

P.S.: We can revert those renames if you wish, but I believe they are on point :)

novusnota:
_I've renamed `calculateAddressHash()` to `contractHash()`, see the rationale [here](https://github.com/tact-lang/tact/pull/1766#issuecomment-2658464256)._

The current implementation of `contractHash()` produces the following Fift assembly. Notice lots of `SWAP`s and such:

```ts
$global_contractHash PROCINLINE:<{
      NEWC
  131380 PUSHINT
  SWAP
  24 STU
  s2 PUSH
      CDEPTH
  SWAP
  16 STU
  OVER
      CDEPTH
  SWAP
  16 STU
  s0 s2 XCHG
      HASHCU
  ROT
  256 STU
  SWAP
      HASHCU
  SWAP
  256 STU
      ONE HASHEXT_SHA256
}>
```

If we rewrite it as a Tact `asm` function, we can easily get rid of lots of `SWAP`s and similar instructions by placing the data optimally, thus optimizing the resulting gas consumption. Furthermore, we would not need to introduce a separate `Builder.hash()` function at all, since it would be inlined.

That's why instead of writing docs for `Builder.hash()` function, I've efficiently rewrote the `contractHash()` function to Tact assembly:

```tact
asm fun contractHash(code: Cell, data: Cell): Int {
    // According to the https://docs.tact-lang.org/book/cells#cells-representation,
    // the layout for the Builder to hash goes as follows:
    // 1) refs_descriptor:bits8 | bits_descriptor:bits8 | data:bitsN
    //
    //  refs_descriptor: ref_count + ((exotic? & 1) * 8) + (mask * 32)
    //                   2 refs (code + data), non-exotic, zero-mask
    //
    //  bits_descriptor: floor(bit_count / 8) + ceil(bit_count, 8)
    //                   floor (5 bits / 8) + ceil(5 bits / 8) = 0 + 1 = 1
    //
    //  data: [0b00110] + [0b100] = [0b00110100] = 0x34 (data + augmented bits)
    //        0b00110 - data (split_depth, special, code, data, Library)
    //        0b100 - augmented bits (Leading 1 + zeroes to make section multiple of eight)
    //
    //  That is: (2 << 16) | (1 << 8) | 0x34 = 131380 for all three.
    //
    // 2) and 3) depth_descriptors: CDEPTH of `code` and CDEPTH of `data`
    // 4) and 5) ref hashes: HASHCU of `code` and HASHCU of `data`

    // Group 1: Computations and arrangements
    s0 PUSH HASHCU // `data` hash
    s2 PUSH HASHCU // `code` hash
    SWAP2
    CDEPTH         // `data` depth
    SWAP
    CDEPTH         // `code` depth
    131380 INT     // (2 << 16) | (1 << 8) | 0x34

    // Group 2: Composition of the Builder
    NEWC
    24 STU  // store refs_descriptor | bits_descriptor | data
    16 STU  // store depth_descriptor for `code`
    16 STU  // store depth_descriptor for `data`
    256 STU // store `code` hash
    256 STU // store `data` hash

    // Group 3: SHA256 hash of the resulting Builder
    ONE HASHEXT_SHA256 // a.k.a. Builder.hash()
}
```

In this form, it consumes 90 **less** gas than its non-`asm` counterpart, while producing identical results.

@anton-trunov @Shvandre @i582 @Gusarich If you agree with this vision, let's remove the `Builder.hash()` and replace the current `contractHash()` with my `asm` implementation :)

src/stdlib/stdlib/std/internal/cells.tact
novusnota: :eyes: https://github.com/tact-lang/tact/pull/1766#issuecomment-2658464665

novusnota:
Fixing benchmarks, uno momento por favor

anton-trunov (approved):

novusnota (approved):

PR #1794, opened 2025-02-12
feat: `isqrt` stdlib function

### LINKED ISSUE #1710: Add `isqrt()` to the standard library
A follow-up on https://github.com/tact-lang/tact/issues/770, as discussed in https://github.com/tact-lang/tact/pull/986. Basically, our `sqrt()` has the same implementation as [in `mathlib.fc`](https://github.com/ton-blockchain/ton/blob/master/crypto%2Fsmartcont%2Fmathlib.fc#L205), which rounds to the nearest integer, while it would've been nice to have the `isqrt()` function that'll floor, i.e. [integer square root](https://en.wikipedia.org/wiki/Integer_square_root).

See:
* https://en.wikipedia.org/wiki/Integer_square_root
* https://github.com/Aisse-258/bigint-isqrt (mini-library with 1 function)

Gusarich:
## Issue

Closes #1710.

## Checklist


docs/src/content/docs/ref/core-math.mdx
```diff
@@ -105,6 +105,30 @@ sqrt(0);  // 0
...
+## isqrt
+
+<Badge text="Gas-expensive" title="Uses 500 gas units or more" variant="danger" size="medium"/>
+<Badge text="Available since Tact 1.6" variant="tip" size="medium"/><p/>
+
+```tact
+fun isqrt(num: Int): Int;
+```
+
+Computes the [integer square root](https://en.wikipedia.org/wiki/Integer_square_root) of the [`Int{:tact}`][int] value `num`. Returns the result rounded down (floor). This is equivalent to finding the largest integer `x` such that `x * x <= num`.
```
novusnota:
```suggestion
Computes the [integer square root](https://en.wikipedia.org/wiki/Integer_square_root) of the [`Int{:tact}`][int] value `num`. Returns the [rounded down](https://en.wikipedia.org/wiki/Rounding#Rounding_down) result. This is equivalent to finding the largest integer `x` such that `x * x ≤ num`.
```

src/stdlib/stdlib/std/internal/math.tact
```diff
@@ -424,3 +424,46 @@ fun sqrt(num: Int): Int {
 
     return x;
 }
+
+/// Global function. Available since Tact 1.6.0.
+///
+/// Computes the integer square root of the `Int` value `num`. Returns the result rounded down (floor).
+/// This is equivalent to finding the largest integer `x` such that `x * x <= num`.
```
novusnota:
```suggestion
/// Computes the integer square root of the `Int` value `num`. Returns the rounded down result.
/// This is equivalent to finding the largest integer `x` such that `x * x ≤ num`.
```

novusnota (changes requested):
:fire: Awesome work on the docs!

Just minor nits (re-gen the `stdlib.ts` afterwards, pls)

novusnota (dismissed):

src/test/e2e-emulated/math.spec.ts
anton-trunov, line 1: let's also add a test for `2**257 - 1`

src/stdlib/stdlib/std/internal/math.tact
```diff
@@ -424,3 +424,46 @@ fun sqrt(num: Int): Int {
...
+/// ```
+///
+/// See: https://docs.tact-lang.org/ref/core-math#isqrt
+///
+fun isqrt(num: Int): Int {
+    if (num == 0) {
+        return 0;
+    }
+
+    let s: Int = log2(num);
```
verytactical: Could you please add a reference to where this algorithm was taken from?

Understanding this is Newton's method doesn't really help to unpack how exactly it works.

src/test/e2e-emulated/math.spec.ts
```diff
@@ -564,6 +564,18 @@ describe("math", () => {
                 "Unable to execute get method. Got exit_code: 5",
             );
         }
+
+        // isqrt: floor(sqrt(num))
+        for (let num = 0n; num <= 100n; num++) {
```
verytactical, line 573: Let's add a test where this loop is moved into the contract and goes over way larger range, and then skip it.

src/test/e2e-emulated/math.spec.ts
```diff
@@ -564,6 +564,18 @@ describe("math", () => {
...
+            expect(await contract.getIsqrt(num)).toBe(
+                BigInt(Math.floor(Math.sqrt(Number(num)))),
```
verytactical: I would not rely on JS semantics here.

When we compute `isqrt(x) = i`, it means that `i^2 <= x < (i+1)^2`, and I'd check that with `bigint`s.

src/stdlib/stdlib/std/internal/math.tact
Gusarich: exactly same as for `sqrt` function, with the only difference being last condition that checks `x * x > num` to make it floor.
Perhaps we can just call `sqrt` from it to get `x`.

src/test/e2e-emulated/math.spec.ts
Gusarich: reworked this one

src/test/e2e-emulated/math.spec.ts
Gusarich: why do we need it?

src/test/e2e-emulated/math.spec.ts
Gusarich: done

src/stdlib/stdlib/std/internal/math.tact
anton-trunov: hmmm, let's try that. what would be the difference between these two implementations?

Gusarich: pushed the commit. have a look at how it's done.

it will probably increase gas consumption a little bit until we have optimized inlining

src/stdlib/stdlib/std/internal/math.tact
```diff
@@ -430,3 +430,34 @@ fun sqrt(num: Int): Int {
...
+///     isqrt(-1); // ERROR! Exit code 5: Integer out of expected range
+/// }
+/// ```
+///
+/// See: https://docs.tact-lang.org/ref/core-math#isqrt
+///
+fun isqrt(num: Int): Int {
+    let x = sqrt(num);
+
+    if (x > num / x) {
```
anton-trunov, line 457: it's not clear why `isqrt(0) == 0` and not an exception

Gusarich: why should it be an exception?

anton-trunov:
Let's not add this function, since it can be easily expressed with `sqrt`

src/stdlib/stdlib/std/internal/math.tact
anton-trunov: `let x = sqrt(0)`, so x = 0 and `num / x` should produce an exception

PR #1759, opened 2025-02-10
feat: better errors

verytactical:
## Issue

Towards #66.
Towards #183.
Towards #1547.

## Checklist

- [ ] I have updated CHANGELOG.md
- [x] I have run all the tests locally and no test failure was reported
- [x] I have run the linter, formatter and spellchecker
- [x] I did not do unrelated and/or undiscussed refactorings

PR #1799, opened 2025-02-12
feat: rewrite oldstyle throws

### LINKED ISSUE #678: Replace `throw Error(...)` with appropriate Tact error in the rest of codebase
A follow-up on #645.

Gusarich:
## Issue

Closes #678.

## All places I skipped

Those are skipped because they are related to work with files, CLI argument parsing, testing etc.

- https://github.com/tact-lang/tact/blob/081cbc5f833916a41fccdd5e91a81bca27025532/version.build.ts#L28
- https://github.com/tact-lang/tact/blob/081cbc5f833916a41fccdd5e91a81bca27025532/examples/contracts.build.ts#L28
- https://github.com/tact-lang/tact/blob/081cbc5f833916a41fccdd5e91a81bca27025532/src/test/utils/all-in-folder.build.ts#L44

## Checklist

### DIFF
--- a/src/func/funcCompile.ts
+++ b/src/func/funcCompile.ts
@@ -1,4 +1,8 @@
 import { ILogger } from "../context/logger";
+import {
+    throwCompilationError,
+    throwInternalCompilerError,
+} from "../error/errors";
 
 // Wasm Imports
 // eslint-disable-next-line @typescript-eslint/no-require-imports
@@ -112,7 +116,9 @@ export async function funcCompile(args: {
                                 (v) => v.path === data,
                             );
                             if (!fl) {
-                                throw Error("File not found: " + data);
+                                throwCompilationError(
+                                    "File not found: " + data,
+                                );
                             }
                             allocatedPointers.push(
                                 writeToCStringPtr(mod, fl.content, contents),
@@ -181,7 +187,7 @@ export async function funcCompile(args: {
         }
     } catch (e) {
         args.logger.error(e as Error);
-        throw Error("Unexpected compiler response");
+        return throwInternalCompilerError("Unexpected compiler response");
     } finally {
         for (const i of allocatedFunctions) {
             mod.removeFunction(i);
@@ -190,5 +196,5 @@ export async function funcCompile(args: {
             mod._free(i);
         }
     }
-    throw Error("Unexpected compiler response");
+    throwInternalCompilerError("Unexpected compiler response");
 }
--- a/src/generator/writers/cast.ts
+++ b/src/generator/writers/cast.ts
@@ -2,6 +2,7 @@ import { getType } from "../../types/resolveDescriptors";
 import { TypeRef } from "../../types/types";
 import { WriterContext } from "../Writer";
 import { ops } from "./ops";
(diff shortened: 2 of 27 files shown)


src/generator/writers/cast.ts
```diff
@@ -11,7 +12,7 @@ export function cast(
 ) {
     if (from.kind === "ref" && to.kind === "ref") {
         if (from.name !== to.name) {
-            throw Error("Impossible");
```
verytactical, line 14: OMG

src/grammar/next/index.ts
```diff
@@ -1388,10 +1388,7 @@ export const getParser = (ast: FactoryAst) => {
             display,
             (message: string) => (source: $.Loc) => {
                 const srcInfo = locationToSrcInfo(source);
-                throw new TactCompilationError(
```
verytactical, line 1391: Is this equivalent to what it was before?

src/vfs/createVirtualFileSystem.ts
```diff
@@ -33,7 +33,7 @@ export function createVirtualFileSystem(
             const name = filePath.slice(normalizedRoot.length);
             const content = fs[name];
             if (typeof content !== "string") {
-                throw Error(`File ${name} not found at ${filePath}`);
+                throw new Error(`File ${name} not found at ${filePath}`);
```
verytactical: why not here? this is a compilation error

verytactical:
Everywhere except `.build.ts` files we're safe to throw one of those errors. In `.build.ts` files we're not even doing compilation, so `throwCompilerError` would be weird.

Also we need a linter rule, otherwise this job will be soon undone.

https://eslint.org/docs/latest/rules/no-restricted-syntax
https://astexplorer.net/

src/grammar/next/index.ts
Gusarich: why not?

verytactical: `display.at(srcInfo, message)` became `message`

Gusarich: we pass `srcInfo` in `throwCompilationError` and it formats it by itself

src/func/funcCompile.ts
```diff
@@ -181,7 +187,7 @@ export async function funcCompile(args: {
         }
     } catch (e) {
         args.logger.error(e as Error);
-        throw Error("Unexpected compiler response");
+        return throwInternalCompilerError("Unexpected compiler response");
```
Gusarich, line 190: @verytactical do you know any other way to fix linter error here? when I replaced `throw Error(..)` with `throwInternalCompilerError` it started showing errors in the next throw below it, and the only "good" way to fix it that I found was to add `return` here.

anton-trunov:
@Gusarich let's fix the merge conflicts and finalize it

anton-trunov:
Let's either fix the merge conflicts or close this PR

Gusarich:
> Let's either fix the merge conflicts or close this PR

#1759 changed internal functions for throwing errors so I think it'll be easier to do this PR from scratch rather than going through all places with errors again

Issue #132, opened 2024-02-15
Gas benchmarks


anton-trunov:
This check should be integrated to CI.

Gusarich:
what if some security-related change causes a larger gas consumption?

anton-trunov:
That's a good point. We prioritize security and correctness over gas consumption for sure. Cases like this will be documented

Gusarich:
Can that be done via artifacts? Or maybe we could just store the last gas usage in some file in the repository 🤔

anton-trunov:
@Gusarich There are [benchmarks](https://github.com/tact-lang/tact/blob/main/src/benchmarks/benchmarks.spec.ts) -- we might just revise those and see if the benchmarks need to be extended

Gusarich:
> @Gusarich There are [benchmarks](https://github.com/tact-lang/tact/blob/main/src/benchmarks/benchmarks.spec.ts) -- we might just revise those and see if the benchmarks need to be extended 

So we just have to save the gas usage in all tests to snapshots?

anton-trunov:
> So we just have to save the gas usage in all tests to snapshots?

I think we should separate functional correctness tests and gas consumption benchmarks. The latter should have more realistic scenarios

Gusarich:
> > So we just have to save the gas usage in all tests to snapshots?
> 
> I think we should separate functional correctness tests and gas consumption benchmarks. The latter should have more realistic scenarios

So maybe it makes sense to wait for #131 ?

anton-trunov:
yeah, these are definitely related issue

PR #1811, opened 2025-02-13
tests: add escrow benchmarks

### LINKED ISSUE #1774: Add escrow contract to benchmarks
FunC vs Tact

Kaladin13:
## Issue

Closes #1774
Towards #132 

## Checklist


Kaladin13:
image

escrow benchmark final table

verytactical (dismissed):

src/test/benchmarks/util.ts
```diff
@@ -11,3 +13,123 @@ export function getUsedGas(sendEnough: SendMessageResult): number {
...
+        return chalk.gray(`same`);
+    }
+    return number >= 0
+        ? chalk.redBright(`(+${change}%)`)
+        : chalk.green(`(${change}%)`);
+}
+
+function calculateChanges(
+    results: BenchmarkResult[],
+    METRICS: readonly string[],
```
i582: why not `metrics`? 🤔

src/test/benchmarks/util.ts
```diff
@@ -11,3 +13,123 @@ export function getUsedGas(sendEnough: SendMessageResult): number {
         )
         .reduceRight((prev, cur) => prev + cur);
 }
+
+type BenchmarkResult = {
+    label: string;
+    commit: string | null;
```
i582: AFAIK, we use `undefined` instead of `null`:

STYLEGUIDE.md:
> Avoid null. typeof null === 'object', and there is undefined anyway.

src/test/benchmarks/util.ts
```diff
@@ -11,3 +13,123 @@ export function getUsedGas(sendEnough: SendMessageResult): number {
...
+    gas: Record<string, number>;
+};
+
+type RawBenchmarkResult = {
+    results: {
+        label: string;
+        commit?: string;
```
i582: STYLEGUIDE.md:
> Don't use optional fields foo?: Bar. Every ? doubles number of cases that should be tested. Eventually some combination of non-defined and undefined fields will be non-semantic. Prefer at least an explicit foo: Bar | undefined, or better create a type with descriptive name to make it a proper tagged union.

i582 (commented):
Great work! Just some minor notes

src/test/benchmarks/util.ts
verytactical: Well, we should have just added `"commit": null` to every result.

Kaladin13: @verytactical You mean directly in json schema?

src/test/benchmarks/util.ts
Kaladin13: fixed to lowcase, this was copied from previous version

src/test/benchmarks/util.ts
Kaladin13: Added explicit `"commit": null` to previous benchmarks for consistency, removed optional field

src/test/benchmarks/util.ts
Kaladin13: Changed in last commit to undefined

i582 (approved):
LGTM

verytactical (approved):

Issue #369, opened 2024-06-02
Tact generated FunC code is not future proof


0kenx:
Why does Tact generate the following FunC code:
```
slice __tact_verify_address(slice address) inline {
    throw_unless(136, address.slice_bits() == 267);
    var h = address.preload_uint(11);
    throw_if(137, h == 1279);
    throw_unless(136, h == 1024);
    return address;
}
```
Wouldn't that always fail when `chain ID != 0` when in the future some addresses are from a different chain ID?

It should just generate
```
-    throw_unless(136, h == 1024);
+    throw_unless(136, h >> 8 == 2);
````

anton-trunov:
Hey @0kenx, you contribution qualifies for the Tact contributor [SBT](https://society.ton.org/contribute-to-tact-compiler), would you mind sharing your TON wallet?

anton-trunov:
@0kenx Would you like to open a PR and contribute this change?

0kenx:
I can fix this single instance, but I believe a more systematic review of the codegen is necessary.

anton-trunov:
Yep, codegen review is planned actually

PR #1876, opened 2025-02-18
feat(stdlib): add safe `Slice.asAddress` function

### LINKED ISSUE #1736: Add an extension function `asAddress(self: Slice)` to the stdlib

i582:
## Issue

Closes #1736.
Towards #369.

## Checklist


novusnota (changes requested):
Doing edits related to re-entrance of the exit code 136, I'll be swift

docs/src/content/docs/ref/core-cells.mdx
```diff
@@ -1298,6 +1298,27 @@ let buzz: Cell = beginCell().storeSlice(s).endCell();
...
+
+<Badge text="Available since Tact 1.6 (not released yet)" variant="tip" size="medium"/><p/>
+
+```tact
+extends fun asAddress(self: Slice, chain: Int): Address;
+```
+
+Extension function for the [`Slice{:tact}`][slice].
+
+Safely casts the [`Slice{:tact}`][slice] to an [`Address{:tact}`][p] and returns it. The inverse of [`Address.asSlice(){:tact}`](#addressasslice).
```
anton-trunov: We should mention here it is more expensive in terms of gas compared to `.asAddressUnsafe`

src/stdlib/stdlib/std/internal/contract.tact
```diff
@@ -1,3 +1,32 @@
...
+/// * https://docs.tact-lang.org/ref/core-cells#sliceasaddress
+/// * https://docs.tact-lang.org/ref/core-cells#addressasslice
+///
+extends fun asAddress(self: Slice, chain: Int): Address {
+    // 11 bits for the prefix,
+    // 256 bits for the address itself
+    nativeThrowUnless(136, self.bits() == 267);
+
+    let basechainAddrPrefix = chain == -1 ? 1279 : 1024;
+    nativeThrowUnless(136, self.preloadUint(11) == basechainAddrPrefix);
```
anton-trunov: Does this address issue #369? @Gusarich @Shvandre @Kaladin13 can you please help with this?

Shvandre: Yeah, it is actually the same problem here. But I don't think addrVars and new chains will come in near future

src/stdlib/stdlib/std/internal/contract.tact
```diff
@@ -1,3 +1,32 @@
...
+/// See:
+/// * https://docs.tact-lang.org/ref/core-cells#sliceasaddress
+/// * https://docs.tact-lang.org/ref/core-cells#addressasslice
+///
+extends fun asAddress(self: Slice, chain: Int): Address {
+    // 11 bits for the prefix,
+    // 256 bits for the address itself
+    nativeThrowUnless(136, self.bits() == 267);
+
+    let basechainAddrPrefix = chain == -1 ? 1279 : 1024;
```
Shvandre: But maybe we don't need to accept `chain` as a parameter? 

Just check, if address is in basechain or masterchain

src/stdlib/stdlib/std/internal/contract.tact
Shvandre: And the docs also states that 136 exit code is 
> Removed since Tact 1.6 (not released yet)

I think we should either remove this from docs either use different exit code here. Maybe 134 `incorrect argument`?

novusnota: I'm adding that exit code back, no troubles here :)

novusnota: The solution described there can be applied here.

I also don't think addrVars and new chains will become available soon, but just in case we can slowly start our preparations with this little function :)

src/stdlib/stdlib/std/internal/contract.tact
novusnota: Thought about it, but that'll limit the API in the future iff we'd want to extend it for other chains (not that likely, but still)

src/stdlib/stdlib/std/internal/contract.tact
novusnota: @Shvandre hmm, we only need to support the addr_std, right? Like, there were ideas for supporting addr_none elsewhere, so now I'm not sure.

In any case, I'll push my updated changes and then it can be discussed some more. I'll mark this as resolved then.

anton-trunov: @novusnota yeah, we also want to support addr_none as an inhabitant of the Address type

docs/src/content/docs/ref/core-cells.mdx
novusnota: Done, please take a look!

src/stdlib/stdlib/std/internal/contract.tact
novusnota: > yeah, we also want to support addr_none as an inhabitant of the Address type

Ok, that can be done later in the same function or we'll have a separate `.asAddressNone()`.

As for now, I've re-introduced the exit code 136 and made it a bit more specific: to validate the standard address, i.e. [`addr_std`](https://docs.ton.org/v3/documentation/data-formats/tlb/msg-tlb#addr_std10). This removes the limitations of chain IDs described in #369 and narrows the scope of currently supported inhabitants of Address type.

novusnota:
Snapshots failed, updating those in a moment.

A brief summary: I've re-introduced the exit code 136 and made it a bit more specific: to validate the standard address, i.e. [`addr_std`](https://docs.ton.org/v3/documentation/data-formats/tlb/msg-tlb#addr_std10). This removes the limitations of chain IDs described in #369 and narrows the scope of currently supported inhabitants of Address type to just `addr_std`, with the possibility to add `addr_none` and, potentially, `addr_extern`.

src/test/exit-codes/contracts/tact-reserved-contract-errors.tact
```diff
@@ -96,4 +97,60 @@ contract ReservedContractErrorsTester with Ownable {
...
+            failed = false;
+        } catch (exitCode) {
+            nativeThrowIf(exitCode, exitCode != targetCode);
+        }
+        require(failed, "Slice.asAddress() didn't error on invalid account ID size");
+
+        // As we got here, everything above caused the target exit code,
+        // so let's throw it explicitly now for the tests on Blueprint's side
+        throw(targetCode);
+    }
```
jeshecdom, line 200: Let's add the following cases so that the tests are independent of how the function is implemented. I understand that the following cases are already covered in the possible paths of the current implementation of `asAddress`, but let us suppose that someone in the future decides to change the implementation of `asAddress`, missing some of the following cases:
- Valid tag prefix in chainID = -1, but incorrect account ID size.
- Valid tag prefix in chainID = 1, but incorrect account ID size.
- Invalid tag prefix in chainID = 0, and correct size.

novusnota: Thanks for the suggestion, covered those cases as well

jeshecdom (commented):
Docs and tests LGTM

anton-trunov (approved):

PR #1901, opened 2025-02-18
fix: contract parameters

### LINKED ISSUE #1756: Introduce contract parameters to resolve unnecessary lazy contract initialization
## Problem statement

The initial data cell of a contract is defined using the parameters of the `init(...)` function which is called right before the first incoming message is processed by a contract.

To not call `init` several times, we use the `lazy_deployment_completed` bit which is the first bit in the Tact contract storage. So, we have to make the contract storage 1 bit larger and also spend some gas on determining if we need to call `init()` for each incoming message. This can be wasteful for the common case of when the `init` parameters and contract storage variables are exactly the same:

```tact
struct ContractStorage {
  owner: Address;
  value: Int as coins;
}

contract Foo with Ownable {
  owner: Address;
  value: Int as coins;

  init(initData: ContractStorage) {
    self.owner = initData.owner;
    self.value = initData.value;
  }

  // ...
}
```

This is also a lot of boilerplate.

## Proposed solution

Allow defining contract storage as contract parameters like so:

```tact
contract Foo(
    owner: Address, 
    value: Int as coins,
) with Ownable {
    // no contract storage variables, no init
    // ...
}
```

If there are no contract storage variables, there is no need to create the `init` function and hence no need to keep the lazy deployment bit.  Notice how this reduces the boilerplate, since the new proposed syntax is basically equivalent to what we had in the problem statement section.

One of the pros of the proposed syntax is that contract parameters resemble how we use the `initOf` operator: `initOf(owner, value)`.

Also, we should still generate the `fromInit` TS wrapper to allow easy contract deployment from TS scripts.

TLDR:
- Contract has `init()` - works the same as it did
- Contract is defined with parameters - no lazy init
- Contract has neither parameters nor `init()` - no lazy init
- Contract has both parameters and `init()` - error

verytactical:
## Issue

Closes #1756.

## Checklist

### DIFF
--- a/src/types/resolveDescriptors.ts
+++ b/src/types/resolveDescriptors.ts
@@ -433,6 +433,76 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
     }
 
     for (const a of ast.types) {
+        if (a.kind === "contract" && a.params !== undefined) {
+            const s = types.get(idText(a.name))!;
+            for (const d of a.declarations) {
+                if (d.kind === 'contract_init') {
+                    throwCompilationError(
+                        `Initialization was already defined on contract ${a.name.text}`,
+                        d.loc,
+                    );
+                }
+            }
+
+            const params: InitParameter[] = [];
+            const args: A.AstTypedParameter[] = [];
+            const statements: A.AstStatement[] = [];
+            for (const r of a.params) {
+                const type = buildTypeRef(r.type, types);
+                params.push({
+                    name: r.name,
+                    type,
+                    as: r.as?.text ?? null,
+                    loc: r.loc,
+                });
+                if (isRuntimeType(type)) {
+                    throwCompilationError(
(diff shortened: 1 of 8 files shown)


src/types/resolveDescriptors.ts
```diff
@@ -432,6 +432,76 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
...
+                params.push({
+                    name: r.name,
+                    type,
+                    as: r.as?.text ?? null,
+                    loc: r.loc,
+                });
+                if (isRuntimeType(type)) {
+                    throwCompilationError(
+                        printTypeRef(type) +
+                            " is a runtime-only type and can't be used as a init function parameter",
```
anton-trunov, line 461:
```suggestion
                            " is a runtime-only type and can't be used as an init function parameter",
```

PR #1931, opened 2025-02-20
fix/refactor(bench/jetton): correct and optimize Jetton Minter benchmark

### LINKED ISSUE #1929: `src/test/benchmarks/contracts/jetton_minter_discoverable.tact` should correspond to `jetton-minter-discoverable.fc`
https://github.com/ton-blockchain/token-contract/blob/main/ft/jetton-minter-discoverable.fc

anton-trunov:
## Issue

Closes #1929.

## Benchmark results

image

## Checklist

- [x] I have added tests to demonstrate the contribution is correctly implemented: it reduces gas consumption for the common cases, makes storage improvements, etc. and does not have too much negative impact for the rest of the use-cases

### DIFF
--- a/src/test/benchmarks/contracts/jetton_minter_discoverable.tact
+++ b/src/test/benchmarks/contracts/jetton_minter_discoverable.tact
@@ -1,6 +1,5 @@
 // https://github.com/ton-blockchain/TEPs/blob/master/text/0089-jetton-wallet-discovery.md
 
-import "@stdlib/ownable";
 import "./jetton_wallet";
 import "./messages";
 
@@ -14,7 +13,7 @@ struct JettonMasterState {
     jettonWalletCode: Cell;
 }
 
-contract JettonMinter with OwnableTransferable {
+contract JettonMinter {
     totalSupply: Int as coins;
     owner: Address;
     jettonContent: Cell;
@@ -67,15 +66,13 @@ contract JettonMinter with OwnableTransferable {
         }
     }
 
-    receive(msg: JettonUpdateContent){
-        self.requireOwner(); // Allow changing content only by owner
-        self.jettonContent = msg.content; // Update content
+    receive(msg: JettonUpdateContent) {
+        nativeThrowUnless(73, sender() == self.owner);
+        self.jettonContent = msg.content;
     }
 
     receive(msg: Mint) {
-        self.requireOwner(); // Allow minting only by owner
-        // Maybe we should check that msg.value is enough to cover the gas fees
-        // Or, maybe we should do self.totalSupply -= msg.amount if bounced.
+        nativeThrowUnless(73, sender() == self.owner);
         self.totalSupply += msg.amount; // Update total supply
 
         deploy(DeployParameters{
@@ -92,6 +89,11 @@ contract JettonMinter with OwnableTransferable {
             }.toCell(),
             init: getJettonWalletInit(msg.receiver)
         });
+    }
+
+    receive(msg: ChangeOwner) {
+        nativeThrowUnless(73, sender() == self.owner);
+        self.owner = msg.newOwner;
     }
 
     get fun get_jetton_data(): JettonMasterState {
--- a/src/test/benchmarks/contracts/jetton_minter_discoverable.tact
+++ b/src/test/benchmarks/contracts/jetton_minter_discoverable.tact
@@ -95,10 +95,6 @@ contract JettonMinter with OwnableTransferable {
         });
     }
 
-    bounced(msg: bounced<JettonTransferInternal>){
-        self.totalSupply -= msg.amount;
-    }
-
     get fun get_jetton_data(): JettonMasterState {
         return JettonMasterState {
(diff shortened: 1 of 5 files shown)


anton-trunov:
A tip to review this PR: it's better either to go throw all the individual commits one by one from the beginning, or just look at the final version and compare it to the linked original FunC file

src/test/benchmarks/contracts/jetton_minter_discoverable.tact
```diff
@@ -1,85 +1,80 @@
...
-            mode: SendRemainingValue | SendIgnoreErrors, //ignore errors, because supply already been updated
-            body: JettonExcesses{
-                queryId: msg.queryId
-            }.toCell()
-        });
-    }
+    receive(msg: ProvideWalletAddress) {
+        // we use message fwdFee for estimation of forward_payload costs
+        let ctx = context();
+        let fwdFee = muldivc(fwdFee(ctx.raw), 3, 2);
```
Shvandre, line 31: Why do you use muldivc and custom fwdFee here?

We have good Context().readForwardFee() function

anton-trunov: I'm just trying to stick close to the FunC original:

```
slice sender_address = cs~load_msg_addr();
cs~load_msg_addr(); ;; skip dst
cs~load_coins(); ;; skip value
cs~skip_bits(1); ;; skip extracurrency collection
cs~load_coins(); ;; skip ihr_fee
int fwd_fee = muldiv(cs~load_coins(), 3, 2); ;; we use message fwd_fee for estimation of forward_payload costs
```

src/test/benchmarks/contracts/jetton_minter_discoverable.tact
```diff
@@ -1,85 +1,80 @@
...
-asm fun emptyAddress(): Address { b{00} PUSHSLICE }
+const ProvideAddressGasConsumption: Int = 10000000;
+const Workchain: Int = 0;
 
 struct JettonMasterState {
     totalSupply: Int as coins;
-    mintable: Bool;
     adminAddress: Address;
     jettonContent: Cell;
     jettonWalletCode: Cell;
```
Shvandre, line 14: Maybe we should also removed that from here? Strange, AFAIR it was already removed in benchmarks in some PR

src/test/benchmarks/contracts/jetton_minter_discoverable.tact
novusnota: It's just taken from this reference implementation: 

https://github.com/ton-blockchain/token-contract/blob/21e7844fa6dbed34e0f4c70eb5f0824409640a30/ft/jetton-minter-discoverable.fc#L58

Also, `Context().readForwardFee()` does almost everything the same, but it floors, not ceils and uses the `first_frac` from the config, while this estimate — does not

src/test/benchmarks/contracts/jetton_minter_discoverable.tact
anton-trunov: my bad, restored

src/test/benchmarks/contracts/jetton_minter_discoverable.tact
Shvandre: > Also, Context().readForwardFee() does almost everything the same, but it floors, not ceils and uses the first_frac from the config, while this estimate — does not

Other-words it is better in every way

anton-trunov: I'm happy to use `readForwardFee()` if you can demonstrate it saves gas here. Provided we get the same  `3 / 2` factor.

Shvandre: image
Hm, strange. I thought it would save more gas

anton-trunov: Do you also keep the `3/2` factor?

Shvandre: > Provided we get the same 3 / 2 factor.

As stated here
https://docs.ton.org/v3/documentation/smart-contracts/transaction-fees/fees-low-level#formula-1

Action fee is `first_frac / 2**16` which is very close to `1/3` (0.3333282470703125) (as first_frac is 21845 https://tonviewer.com/config#25)

and GETFORWARDFEE calculates original forward_fee based on that

anton-trunov: We are not porting TON Docs to Tact, we are porting https://github.com/ton-blockchain/token-contract/blob/main/ft/jetton-minter-discoverable.fc and https://github.com/ton-blockchain/token-contract/blob/main/ft/jetton-wallet.fc to Tact. This is not about getting things right, this is about comparing gas usage of those two implementations. To make comparisons honest, we need to keep the semantics of those contracts as close as possible.

Shvandre (approved):
LGTM

PR #1934, opened 2025-02-20
fix(stdlib): replace `Context.bounced` with `Context.bounceable` flag

### LINKED ISSUE #1294: Message context is missing bounce flag and some others
```tact
struct Context {
    bounced: Bool;
    sender: Address;
    value: Int;
    raw: Slice;
}
```
```tlb
int_msg_info$0 ihr_disabled:Bool bounce:Bool bounced:Bool
  src:MsgAddressInt dest:MsgAddressInt 
  value:CurrencyCollection ihr_fee:Grams fwd_fee:Grams
  created_lt:uint64 created_at:uint32 = CommonMsgInfo;
```
This struct is missing several needed fields. The most important one is bounce flag. 
Created_lt and created_at may be also needed in some situations, and now there is no way to get them in receivers too.

Moreover, this struct specify `bounced`. I think it is unnecessary, because special receivers are used for handling bounced messages....

Re: #1212 
Also, maybe it's better to store those in TVM tuple, but I checked FIFT and looks like global tensors are already tuples. 
@Gusarich can you please check it? 

And the last but not least, maybe it is better not to store balance in global context and access it through 
```asm
BALANCE FIRST
```
?
We already have that in stdlib function `myBalance()`

i582:
Fift code before/after:

Screenshot 2025-02-20 at 22 52 26

## Issue

Closes #1294.

## Checklist


src/generator/writers/writeContract.ts
```diff
@@ -329,11 +329,12 @@ export function writeMainContract(
             ctx.append();
             ctx.append(`;; Context`);
             ctx.append(`var cs = in_msg_cell.begin_parse();`);
-            ctx.append(`var msg_flags = cs~load_uint(4);`); // int_msg_info$0 ihr_disabled:Bool bounce:Bool bounced:Bool
-            ctx.append(`var msg_bounced = -(msg_flags & 1);`);
+            ctx.append(`cs~skip_bits(2);`); // skip int_msg_info$0 ihr_disabled:Bool
+            ctx.append(`var msg_bounceable = cs~load_int(1);`); // bounce:Bool
```
anton-trunov, line 333:
```suggestion
            ctx.append(`var msg_bounceable = cs~load_int(1);`); // bounceable:Bool
```

anton-trunov (changes requested):

src/generator/writers/writeContract.ts
i582: `bounce`!
See https://docs.ton.org/v3/documentation/smart-contracts/message-management/sending-messages

```
int_msg_info$0 ihr_disabled:Bool bounce:Bool bounced:Bool
  src:MsgAddress dest:MsgAddressInt 
  value:CurrencyCollection ihr_fee:Grams fwd_fee:Grams
  created_lt:uint64 created_at:uint32 = CommonMsgInfoRelaxed;
```

novusnota: Hmm, we should call it `bounce` then too, right?

src/test/e2e-emulated/contracts/bounceable.tact
```diff
@@ -0,0 +1,14 @@
...
+
+    receive("test") {
+        let ctx = context();
+        self.wasBounceable = ctx.bounceable;
+    }
+
+    get fun wasBounceable(): Bool {
+        return self.wasBounceable!!;
+    }
+}
```
Kaladin13, line 14: Why we need optional boolean here? Call to getter will fail (in our ts wrapper but still) before first receiver call

```suggestion
contract Test {
    wasBounceable: Bool = false;

    receive() {}

    receive("test") {
        let ctx = context();
        self.wasBounceable = ctx.bounceable;
    }

    get fun wasBounceable(): Bool {
        return self.wasBounceable;
    }
}
```

Kaladin13 (commented):
Everything good on actual logic, just small inconvenience in test contract

src/test/e2e-emulated/contracts/bounceable.tact
i582: But that's actually why I did it this way: in my code we check that the field is actually set to false, and not that something silently happened and the receiver wasn't called at all (what happened when it was implemented incorrectly)

Shvandre (approved):
LGTM

src/test/e2e-emulated/contracts/bounceable.tact
Kaladin13: Oh, I get it, make sense

Maybe then just change get method type to `Bool?` and remove `Non-null assert !!` ? 
This would be explicit and maybe we can even add another test for this, e.g.

```js
expect(bounceFieldBeforeReceiverCall).toBeNull()
```

i582: I don't think we need a test in this case, since this is not what we want to test :)

PR #1936, opened 2025-02-20
fix: `sha256()` for Slices and rework its description

### LINKED ISSUE #1879: Mark `sha256()` as gas-expensive and correct its description
In past PRs of 1.6.0 its on-chain functionality got different, so the description should probably be split in two parts: before 1.6.0 and since 1.6.0.

### LINKED ISSUE #1937: Runtime `sha256()` should work for arbitrary Slices with length >= 128
In https://github.com/tact-lang/tact/pull/1626 we've fixed the runtime behavior of `sha256()` for Strings, but forgot about Slices, which still use the `SHA256U` TVM instruction, which only works with data bits, ignoring all refs.

```tact
// Same implementation as for string_hash() from FunC's stdlib,
// which is currently used for runtime hashing of Slices in Tact.
asm fun currentHash(src: Slice): Int { SHA256U }

// Has the same issues as SHA256U — only uses data bits, ignoring refs
asm fun otherHash(src: Slice): Int { ONE HASHEXT_SHA256 }

fun example() {
    let big = "------------------------------------------------------------------------------------------------------------------------------129".asSlice();
    let first128 = "------------------------------------------------------------------------------------------------------------------------------1".asSlice();

    currentHash(big) == currentHash(first128); // true
    otherHash(big) == otherHash(first128); // true
}
```

novusnota:
Also made the passed Slices use the same function instead of the `string_hash()` from FunC's stdlib, otherwise they'd still have all the same problems we've fixed before for Strings in #1626. Added positive and negative tests, showing that previous implementation for Slices was incorrect on Slices larger than 127 bytes.

## Issue

Closes #1879.
Closes #1937.

## Checklist


anton-trunov:
Looks like this should have a different linked issue. This is a fix of a fix. Can you please add an example that fails into a separate issue?

novusnota:
> Looks like this should have a different linked issue. This is a fix of a fix. Can you please add an example that fails into a separate issue?

Opened and linked: https://github.com/tact-lang/tact/issues/1937

anton-trunov:
Is it still possible to recover the previous behavior? I.e. if the user want to hash just the slice data, without any traversals?

novusnota:
> Is it still possible to recover the previous behavior? I.e. if the user want to hash just the slice data, without any traversals?

If the Slice would not contain any references the results are the same as before.

If the user still wants to hash just the data even if the slice contains some refs, in the "Before Tact 1.6" note I've said that TVM instructions that perform SHA-256 hashing all operate on data only. With that they can simply introduce their own helper `asm` function if that's needed.

Or do we want to add a new function to the stdlib? `.hashData()` or somesuch?

anton-trunov:
> Or do we want to add a new function to the stdlib? .hashData() or somesuch?

I think we should. `.hashData` sounds good

novusnota:
> > Or do we want to add a new function to the stdlib? .hashData() or somesuch?
> 
> I think we should. `.hashData` sounds good

Sure thing, I've opened a separate issue for this — #1950. Will be resolved in a separate PR soon, let's finish this one first

anton-trunov:
merge conflicts

novusnota:
> merge conflicts

Resolved, please take a look

i582 (approved):
Looks good!

anton-trunov (approved):

Issue #1866, opened 2025-02-17
Migrate to binary receivers in Escrow


anton-trunov:
Comment (text/string) receivers use expensive `HASHSU` instructions (526 gas units).

Incidentally, where can we check the corresponding FunC contract? The Tact version should have the exact same interfaces.

PR #1949, opened 2025-02-21
tests: check all combinations of internal, external and bounced receivers for "empty" messages

jeshecdom:
## Issue

Closes #1849.
Closes #1850.
Closes #1924.
Closes #1918.

For issue #1850, this PR tests it implicitly, because this PR checks all possible combinations of receivers (internal, external and bounced).

This PR sends "empty" messages to receivers:
- A message with an op code less than 32 bits.
- An actual empty cell (not even an opcode).
- A message with opcode 0, but with exactly 32 bits in length.
- The empty string.
- An empty string inside a Message struct.
- An empty message struct.

~~In the above I said "towards", because this PR is still not checking all possible combinations of **bounced receivers**.~~
Bounced receivers are now checked.

## Checklist

- [ ] I have updated CHANGELOG.md
- [X] I have run all the tests locally and no test failure was reported
- [X] I have run the linter, formatter and spellchecker
- [X] I did not do unrelated and/or undiscussed refactorings

PR #1968, opened 2025-02-23
refactor(codegen): inline contract router

### LINKED ISSUE #1587: Inline contract routers at codegen stage
Contract routers (`X_contract_router_Y`) should most likely be `inline`, not `inline_ref` as they are called exactly once.

EDIT. It turned out not work that well. See comments.

anton-trunov:
This requires inlining receivers as well.
Reduces gas usage of the Jetton benchmarks.
But increases gas usage for Escrow (I think we can live with it for now, until #1866 is resolved). 

## Issue

Closes #1587.

## Gas Benchmarks

image

image

## Checklist

- [x] I did not do unrelated and/or undiscussed refactorings (but I really wanted to do it)


src/generator/writers/writeRouter.ts
```diff
@@ -250,26 +186,38 @@ function writeBinaryReceiver(
...
+
+        writeStoreContractVariables(contract, wCtx);
+        wCtx.append("return ();");
+        // if (
+        //     binaryReceiver.ast.statements.length === 0 ||
+        //     binaryReceiver.ast.statements[binaryReceiver.ast.statements.length - 1]!.kind !==
+        //         "statement_return"
+        // ) {
+        //     wCtx.append("return ());");
+        // }
```
Kaladin13: Commented code

src/generator/writers/writeContract.ts
```diff
@@ -252,148 +257,124 @@ export function writeInit(
...
-
-                // Throw if not handled
-                ctx.append(`;; Throw if not handled`);
-                ctx.append(
-                    `throw_unless(${contractErrors.invalidMessage.id}, handled);`,
+                wCtx.append(
+                    `var ${contractVariables} = ${ops.contractLoad(contract.name, wCtx)}();`,
                 );
-                ctx.append();
+                wCtx.append();
```
Kaladin13, line 370: > wCtx.append();

Have seen a few of these in different parts of codegen, sometimes as `wCtx.append(``);`

 I think we should add `wCtx.appendEmpty()` and change all similar entries to it
 
 (I know this isn't this PR's fault, but we need to fix it oneday)

src/generator/writers/writeRouter.ts
```diff
@@ -206,26 +145,23 @@ function writeNonBouncedRouter(
...
         wCtx.append(";; Receiver fallback");
-        wCtx.append(
-            `self~${ops.receiveAny(contractName, receivers.kind)}(in_msg);`,
-        );
-        wCtx.append("return (self, true);");
+        writeFallbackReceiver(receivers.fallback, contract, "in_msg", wCtx);
     } else {
-        wCtx.append("return (self, false);");
+        wCtx.append(`;; Throw if not handled`);
+        wCtx.append(`throw(${contractErrors.invalidMessage.id});`);
```
Kaladin13, line 153: Since we are using this a lot maybe add function `wCtx.appendThrow(...)` ?

src/generator/writers/writeRouter.ts
```diff
@@ -503,18 +488,23 @@ function writeBouncedReceiver(
...
+            funcIdOf(selector.name),
+            wCtx,
+            false,
+            selector.bounced,
         );
 
-        // Execute function
         wCtx.append(
-            `self~${ops.receiveTypeBounce(contractName, selector.type)}(msg);`,
+            `var ${msgFields} = in_msg~${selector.bounced ? ops.readerBounced(selector.type, wCtx) : ops.reader(selector.type, "no-opcode", wCtx)}();`,
```
Kaladin13: > selector.bounced ? ops.readerBounced(selector.type, wCtx) : ops.reader(selector.type, "no-opcode", wCtx)

Maybe use separate variable here? Hard to understand what is happening inside string substitution and function calls are with different arguments

Kaladin13 (commented):
General logic is solid, good thing we have #1949 since we will catch all possible router bugs there, if there are any, after this massive codegen changes

Few notes in review

src/generator/writers/writeContract.ts
anton-trunov: Good catch, I'll replace it with `wCtx.append()`, since it means basically `wCtx.appendEmpty()` and is used in like that in 15+ places.

src/generator/writers/writeRouter.ts
anton-trunov: good catch thanks

src/generator/writers/writeRouter.ts
anton-trunov: I could only find 3 places, let's keep it as it as for now

Kaladin13 (dismissed):
Good to go!

PR #1977, opened 2025-02-24
doc(stdlib): doc comments for `address` functions

### LINKED ISSUE #1920: Add doc comments to `address.tact`
- [x] `storeAddress`
- [x] `loadAddress`

anton-trunov:
## Issue

Closes #1920.
Towards #573.

## Checklist


src/stdlib/stdlib/std/internal/address.tact
```diff
@@ -183,6 +183,35 @@ asm fun parseStdAddress(slice: Slice): StdAddress { REWRITESTDADDR }
...
+/// See:
+/// * https://docs.tact-lang.org/ref/core-cells/#sliceloadaddress
+///
 asm( -> 1 0) extends mutates fun loadAddress(self: Slice): Address { LDMSGADDR }
 
+/// Extension function for the `Builder` type.
+///
+/// Stores `address` in the copy of the `self` builder. Returns that copy.
+///
+/// Attempts to store an address into `self` will result in exit code 8: Cell overflow if `self` cannot fit the address.
```
jeshecdom: "Attempts to store an address into `self` will result in exit code 8:  ......."

At first reading this confused me because it made me believe that it would throw the exit code 8 at **any** address. I have a tendency to stop interpreting sentences at signs like ":".

Probably changing the sentence to:

"Attempts to store an address into `self` when `self` cannot fit it, will result in exit code 8: Cell overflow" 

which is the pattern followed by the previous function. But maybe it is fine as it is, I am not completely sure. Feel free to dismiss this comment.

docs/src/content/docs/ref/core-cells.mdx
```diff
@@ -459,6 +459,8 @@ Extension function for the [`Builder{:tact}`][builder].
 
 Stores the `address` in the copy of the [`Builder{:tact}`][builder]. Returns that copy.
 
+Attempts to store an address into `self` will result in [exit code 8](/book/exit-codes#8): `Cell overflow` if `self` cannot fit the address.
```
jeshecdom: Same observation as in the comment for function `storeAddress` in `src/stdlib/stdlib/std/internal/address.tact`.

anton-trunov:
@novusnota can you please address @jeshecdom's suggestions?

novusnota:
> Can you please address @jeshecdom's suggestions?

Sure!

(hold on a moment or two)
(time is an illusion)

UPD: Got distracted by other things plus needed to contemplate on what is really needed to be done.

I agree with adding "type" after the actual type name to make things explicit: "Extension function for the X _type_". But to be consistent, this has to be applied everythere — I'll do that in the morning.

De jure I'm not the one who checked the _"I did not do unrelated and/or undiscussed refactorings"_, so technically it's fine if I'll slightly adjust other docs :)

src/stdlib/stdlib/std/internal/address.tact
novusnota: Done, please take a look!

docs/src/content/docs/ref/core-cells.mdx
novusnota: Done, please take a look!

PR #1955, opened 2025-02-21
tests: compatible func and tact jetton benchmarks

Kaladin13:
## Issue

Closes #1933
Closes #1902

## Checklist

- [ ] I have updated CHANGELOG.md
- [x] I have run all the tests locally and no test failure was reported
- [x] I have run the linter, formatter and spellchecker
- [x] I did not do unrelated and/or undiscussed refactorings

PR #1981, opened 2025-02-24
tests: binary compatible with func escrow benchmarks

### LINKED ISSUE #1866: Migrate to binary receivers in Escrow

### LINKED ISSUE #1980: Make FunC and Tact version of Escrow pass the same test suite
They should accept same messages

Kaladin13:
## Issue

Closes #1866
Closes #1980

## Checklist


spell/cspell-list.txt
```diff
@@ -177,4 +177,6 @@ XMODEM
 xtwitter
 привет
 extracurrency
-nobounce
\ No newline at end of file
+nobounce
+bouncable
```
anton-trunov:
```suggestion
bounceable
```

Kaladin13: Thanks, fixed

Kaladin13:
image

Current benchmark table with same test suite. With compatible tests, FunC gas usage decreased

While even now tests from FunC-only test suite and this one look very similar and without any logic errors, I think from now on (given recent #1955) we need to start with this `compatible-tests` approach to make any new FunC vs Tact benchmarks

spell/cspell-list.txt
```diff
@@ -177,4 +177,6 @@ XMODEM
 xtwitter
 привет
 extracurrency
-nobounce
\ No newline at end of file
+nobounce
+bounceable
+gauarantor
```
anton-trunov:
```suggestion
guarantor
```

src/test/benchmarks/contracts/escrow.tact
```diff
@@ -146,7 +152,7 @@ contract Escrow with Deployable {
     }
 
     // send funds back to buyer
-    receive("cancel"){
+    receive(_: Cancel){
         require(self.isFunded == true, "not funded");
```
anton-trunov: let's use `throwUnless` instead of `require`, so the exit codes are also equivalent to the ones used in the FunC version

src/test/benchmarks/contracts/escrow.tact
```diff
@@ -165,7 +171,7 @@ contract Escrow with Deployable {
     }
 
     // on-chain data access, could be useful to guarantor contract
-    receive("provideEscrowData"){
+    receive(_: ProvideEscrowData){
         self.reply(TakeEscrowData{escrowData: self.escrowInfo()}.toCell());
```
anton-trunov: I think `self.reply` is more expensive than using `send`

src/test/benchmarks/contracts/escrow.tact
```diff
@@ -6,12 +6,18 @@ fun calculateJettonWalletAddress(owner: Address, master: Address, code: Cell): A
...
     newJettonWalletCode: Cell;
 }
 message TakeEscrowData {
     escrowData: EscrowData;
 }
+
+message(0x9451eca9) Funding {}
+message(0xe8c15681) Approve {}
+message(0xcc0f2526) Cancel {}
+message ProvideEscrowData {}
```
anton-trunov: looks like we should fix the opcode here (what is its FunC equivalent?)

src/test/benchmarks/contracts/escrow.tact
Kaladin13: We didn't benchmark this function and from language perspective this is idiomatic usecase for `self.reply`

You think we still need to change it?

src/test/benchmarks/contracts/escrow.tact
```diff
@@ -146,7 +152,7 @@ contract Escrow with Deployable {
     }
 
     // send funds back to buyer
-    receive("cancel"){
+    receive(_: Cancel){
         require(self.isFunded == true, "not funded");
         let ctx: Context = context();
         require(self.guarantorAddress == ctx.sender, "not guarantor");
```
Shvandre: ctx is not used further, so using `sender()` instead of `context().sender` should be better

src/test/benchmarks/contracts/escrow.tact
Kaladin13: Since on-chain API is additional functionality FunC version does not have it. Do we need to add it to FunC?

src/test/benchmarks/contracts/escrow.tact
```diff
@@ -6,12 +6,18 @@ fun calculateJettonWalletAddress(owner: Address, master: Address, code: Cell): A
...
 message TakeEscrowData {
     escrowData: EscrowData;
 }
+
+message(0x9451eca9) Funding {}
+message(0xe8c15681) Approve {}
+message(0xcc0f2526) Cancel {}
+message ProvideEscrowData {}
+
 struct EscrowData {
```
Shvandre, line 20: IIRC we can just use `self` as struct with all data fields. It is suitable for getters, but idk if it works for your on-chain API. Could you, please, check, is that the case? 

P.S. I see that this struct also specifies fields in different order, is that done for purpose?

spell/cspell-list.txt
Kaladin13: fixed

src/test/benchmarks/contracts/escrow.tact
anton-trunov, line 1: - Let's not use `Deployable`, its more expensive (and most likely the FunC contract does it differently)
- prefer `sender()` to `context().sender`
- this might not work, but can we try inlining the contract methods? (maybe with just `inline` annotation),
- also, adding `inline` to the `calculateJettonWalletAddress` function might work (or not)

src/test/benchmarks/contracts/escrow.tact
anton-trunov: Yeah, let's at least try. `reply` gets all the contract vars on stack because it's a contract method for some reason, hence it's usually less efficient

src/test/benchmarks/contracts/escrow.tact
anton-trunov: Yep, to make benchmarking honest, we either remove it from our Tact contract or add to the FunC version

src/test/benchmarks/contracts/escrow.tact
Kaladin13: changed to this type of throws

src/test/benchmarks/contracts/escrow.tact
Kaladin13: Thanks, changed it

src/test/benchmarks/contracts/escrow.tact
Kaladin13: With #2000 this seems not possible for the on-chain API at the moment (

src/test/benchmarks/contracts/escrow.tact
Kaladin13: added

src/test/benchmarks/contracts/escrow.tact
Kaladin13: fixed

src/test/benchmarks/contracts/escrow.tact
Kaladin13: - Switched from `Deployable`
- Changed to `sender()`
- Inlining for both contract methods and `calculateJettonWalletAddress` didn't work (it only increased gas usage by 100-200)

Kaladin13:
image

With all the discussed changes we have

anton-trunov (approved):

PR #1985, opened 2025-02-24
feat: no lazy init (contract parameters)

### LINKED ISSUE #1756: Introduce contract parameters to resolve unnecessary lazy contract initialization

verytactical:
## Issue

Closes #1756.

## Checklist


Shvandre (dismissed):
We discussed this changes with @verytactical and agreed on adding more comments here, especially for field names, such as `separate`, `contract` and some others. 

Also now new constructor requires to add fake ASTNode of init() function, but this cannot be bypassed easily, so it will stay as is.

Besides new constructor lacks the possibility to make some arguments optional, but we decided not to implement that in this PR not to overcomplicate it.

src/generator/writers/writeContract.ts
```diff
@@ -64,36 +64,46 @@ export function writeStorageOps(
...
+                            `return $sc~${ops.reader(type.name, "with-opcode", ctx)}();`,
+                        );
+                    } else {
+                        ctx.append(`return null();`);
+                    }
+                });
+                ctx.append(`} else {`);
+                ctx.inIndent(() => {
+                    // Load arguments
+                    if (type.init!.params.length > 0) {
```
i582, line 84: `!.` -> `?.`

src/types/resolveDescriptors.ts
```diff
@@ -432,11 +432,87 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
         };
     }
 
+    for (const a of ast.types) {
+        if (a.kind === "contract" && a.params !== undefined) {
```
i582, line 436: early continue

src/types/resolveDescriptors.ts
```diff
@@ -1097,10 +1174,17 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
...
-                            d.loc,
-                        );
+                        if (s.init.kind !== "contract") {
+                            throwCompilationError(
+                                "Init function already exists",
+                                d.loc,
+                            );
+                        } else {
+                            throwCompilationError(
+                                "Cannot define init() on a contract that has contract initializer",
```
verytactical: contract parameters

src/types/__snapshots__/resolveDescriptors.spec.ts.snap
```diff
@@ -365,6 +365,24 @@ exports[`resolveDescriptors should fail descriptors for getter-outside-contract
 "
 `;
 
+exports[`resolveDescriptors should fail descriptors for init-on-contract-no-fields 1`] = `
+"<unknown>:6:5: All fields must be defined in contract initializer if it exists
```
verytactical: as contract parameters

src/types/__snapshots__/resolveDescriptors.spec.ts.snap
```diff
@@ -365,6 +365,24 @@ exports[`resolveDescriptors should fail descriptors for getter-outside-contract
...
+  5 | ) {
+> 6 |     foo: Int as int16 = 0;
+          ^~~~~~~~~~~~~~~~~~~~~
+  7 |     get fun getFoo(): Int {
+"
+`;
+
+exports[`resolveDescriptors should fail descriptors for init-on-contract-no-init 1`] = `
+"<unknown>:6:5: Initialization was already defined on contract Contract
```
verytactical: `init() cannot be used along with contract parameters`

src/types/__snapshots__/resolveDescriptors.spec.ts.snap
verytactical: `Cannot use contract fields along with contract parameters`

src/generator/writers/writeContract.ts
verytactical: https://github.com/tact-lang/tact/blob/995b2d028918f2c113cf028e6a061118756a9d96/src/generator/writers/writeContract.ts#L83

verytactical: I would prefer not to do any changes to legacy code here.

anton-trunov (approved):

PR #1991, opened 2025-02-25
feat: enhance error message for `extends asm fun`

### LINKED ISSUE #1619: Better error for `extends asm fun`
`extends asm fun` now produces 
```
Tact compilation failed
Error: @stdlib/std/context.tact:14:9: Expected "abstract", "extends", "fun", "get", "inline", "mutates", "override", or "virtual"
  13 | 
> 14 | extends asm fun readForwardFee(self: Context): Int {
               ^
  15 |     LDMSGADDR
```

Which is quite good. But Maybe It would be better to add something like:
> Maybe you've ment asm extends fun ...

Gusarich:
## Issue

Closes #1619.

## Checklist

- [ ] ~~I have documented my contribution in `docs/` and made the build locally~~

### DIFF
--- a/src/grammar/parser-error.ts
+++ b/src/grammar/parser-error.ts
@@ -88,6 +88,16 @@ export const syntaxErrorSchema = <T, U>(
             return handle(sub`Only full function definitions are allowed here`);
         },
         expected: (expects: ReadonlySet<string>) => {
+            if (
+                expects.has('"fun"') &&
+                [...expects].some((e) => e === '"asm"' || e === '"extends"')
+            ) {
+                return handle(
+                    text(
+                        `Expected ${getExpectedText(expects)}. Maybe you meant: "asm extends fun" instead of "extends asm fun"?`,
+                    ),
+                );
+            }
             return handle(text(`Expected ${getExpectedText(expects)}`));
         },
         invalidFuncId: () => {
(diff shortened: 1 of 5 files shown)


Gusarich:
Note that this fix does not work properly with `prev` grammar

src/grammar/test-failed/extends-asm-fun-error.tact
```diff
@@ -0,0 +1,3 @@
+extends asm fun readForwardFee(self: Context): Int {
```
jeshecdom, line 1: Just out of curiosity: what would be the error message (or not) in all the possible combinations?
- extends mutates asm
- extends asm mutates
- mutates extends asm
- mutates asm extends
- asm mutates extends
- asm extends mutates (this one is the usual order)

src/grammar/parser-error.ts
```diff
@@ -88,6 +88,16 @@ export const syntaxErrorSchema = <T, U>(
             return handle(sub`Only full function definitions are allowed here`);
         },
         expected: (expects: ReadonlySet<string>) => {
+            if (
```
verytactical, line 91: I would rather not do it this way. This is a generic error handler, and this kind of specific patch doesn't really belong here.

verytactical (changes requested):

src/grammar/parser-error.ts
Gusarich: yeah... https://github.com/tact-lang/tact/issues/1619#issuecomment-2642614715

anton-trunov:
Let's close this for now and come back to this in 1.7 milestone

PR #1998, opened 2025-02-25
fix: better error messages for empty `bounced()` and `bounced("string")`

### LINKED ISSUE #1984: Improve error message on `bounced()`
In this contract:
```
contract Test {

    bounced() {
     
    }
}
```
the compiler reports the error:
```
bounced() cannot be used as fallback
  2 | 
> 3 |     bounced() {
          ^~~~~~~~~~~
```
which made me believe that by adding the bounced fallback on slices:
```
contract Test {

    bounced() {
     
    }
    
    bounced(msg: Slice) {
     
    }
}
```
the error would go away, but it keeps saying the same error. My reasoning was as follows: "If the error message is telling me that I cannot use `bounced()` as a fallback, then adding the actual fallback `bounced(slice)` would allow me to use `bounced()`".

A clearer error message would simply say "`bounced()` is not allowed as a bounced receiver". Or even simply say the more informative error message when one adds `bounced(String)`:

"Bounce receive function can only accept bounced message, message or Slice"

i582:
## Issue

Closes #1984.

## Checklist

### DIFF
--- a/src/grammar/parser-error.ts
+++ b/src/grammar/parser-error.ts
@@ -75,12 +75,12 @@ export const syntaxErrorSchema = <T, U>(
         },
         noBouncedWithoutArg: () => {
             return handle(
-                sub`bounced() receiver should accept one of Message, bounced<Message> or Slice`,
+                sub`bounced() receiver should accept a Message, bounced<Message> or Slice`,
             );
         },
         noBouncedWithString: () => {
             return handle(
-                sub`bounced() receiver can only accept Message, bounced<Message> or Slice`,
+                sub`bounced() receiver can only accept a Message, bounced<Message> or Slice`,
             );
         },
         noConstantDecl: () => {
--- a/src/grammar/parser-error.ts
+++ b/src/grammar/parser-error.ts
@@ -74,11 +74,13 @@ export const syntaxErrorSchema = <T, U>(
             return handle(sub`Expression is not callable`);
         },
         noBouncedWithoutArg: () => {
-            return handle(sub`bounced() cannot be used as fallback`);
+            return handle(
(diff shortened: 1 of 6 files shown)


src/grammar/parser-error.ts
```diff
@@ -74,11 +74,13 @@ export const syntaxErrorSchema = <T, U>(
             return handle(sub`Expression is not callable`);
         },
         noBouncedWithoutArg: () => {
-            return handle(sub`bounced() cannot be used as fallback`);
+            return handle(
+                sub`bounced() receiver should accept one of Message, bounced<Message> or Slice`,
```
novusnota:
```suggestion
                sub`bounced() receiver accepts a Message, bounced<Message> or Slice`,
```

Not sure if it's better, but it's more concise

src/grammar/parser-error.ts
```diff
@@ -74,11 +74,13 @@ export const syntaxErrorSchema = <T, U>(
...
+            );
         },
         noBouncedWithString: () => {
             return handle(
-                sub`bounced() cannot be used with a string literal name`,
+                sub`bounced() receiver can only accept Message, bounced<Message> or Slice`,
```
novusnota:
```suggestion
                sub`bounced() receiver accepts a Message, bounced<Message> or Slice`,
```

i582: Fixed

src/grammar/parser-error.ts
i582: Fixed

anton-trunov (changes requested):
Let's fix it for the old parser as well

novusnota (approved):

Issue #2000, opened 2025-02-25
Handle contract type as field type in message


Kaladin13:
![Image](https://github.com/user-attachments/assets/cdd67b03-c511-4f15-85f5-ef697443b843)

When trying to

```tact
contract Escrow {}

message TakeEscrowData {
    escrowData: Escrow;
}
```

PR #2011, opened 2025-02-25
fix: show better error for fields with a contract type

i582:
## Issue

Towards #2000.

## Checklist

### DIFF
--- a/src/types/resolveSignatures.ts
+++ b/src/types/resolveSignatures.ts
@@ -186,6 +186,21 @@ export function resolveSignatures(ctx: CompilerContext, Ast: FactoryAst) {
             throwInternalCompilerError(`Unsupported type: ${name}`);
         }
 
+        for (const field of t.fields) {
+            const type = field.type;
+            if (type.kind !== "ref") {
+                continue;
+            }
+
+            const t = getType(ctx, type.name);
+            if (t.kind === "contract") {
+                throwCompilationError(
+                    `Fields with a contract type are not supported yet`,
+                    field.loc,
+                );
+            }
+        }
+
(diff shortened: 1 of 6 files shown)


src/types/resolveSignatures.ts
```diff
@@ -186,6 +186,21 @@ export function resolveSignatures(ctx: CompilerContext, Ast: FactoryAst) {
             throwInternalCompilerError(`Unsupported type: ${name}`);
         }
 
+        for (const field of t.fields) {
```
anton-trunov, line 189: can this be moved to `src/types/resolveDescriptors.ts` by any chance?

i582: It looks like we can't put this in the `resolveDescriptors()` function, since not all types are resolved and processed at that stage, and we can't safely get type information by name like this:

```typescript
if (a.kind === "struct_decl" || a.kind === "message_decl") {
    const type = types.get(idText(a.name))!;

    for (const f of a.fields) {
        if (type.fields.find((v) => eqNames(v.name, f.name))) {
            throwCompilationError(
                `Field ${idTextErr(f.name)} already exists`,
                f.loc,
            );
        }

        const fieldDescription = buildFieldDescription(
            f,
            type.fields.length,
        );

        if (
            fieldDescription.type.kind === "ref" &&
            getType(ctx, fieldDescription.type.name).kind === "contract"
        ) {
            throwCompilationError(
                `Fields with a contract type are not supported yet`,
                a.loc,
            );
        }

        type.fields.push(fieldDescription);
    }
...
}
```

This fails because while we are processing some stdlib structure, the `Int` type is still not processed, and `getType` throws an exception

PR #2027, opened 2025-02-26
fix: fix of optimized `initOf` with dependency in global function used inside `init()`

### LINKED ISSUE #2019: `initOf` used inside a function might not work
A user reports the following (https://t.me/tactlang/55239):

I’m getting 135 exit code during contract deploy. Looks like that because I use initOf other contract in init of my contract. Any way how I can solve that? I need other contract to be owner of my contract and I don’t want to pass it as arg in init and I don’t want to use sender() as well. I figured it out: I used separate function that returns address of contract: 

```tact
fun master_addr(): Address{
    return contractAddress(initOf MasterV0());
}
```

once I changed my `init` to use `contractAddress(initOf MasterV0())` instead of `master_addr()`, it started to work.

### LINKED ISSUE #2024: `undefined` in `*_child_get_code`
```tact
contract Test {
    owner: Address;

    init() {
        self.owner = master_addr();
    }
}

fun master_addr(): Address {
    return contractAddress(initOf MasterV0());
}

contract MasterV0 {
    state: Int = 0;

    get fun state(): Int {
        return self.state;
    }
}
```

With 1.5.4 it works just fine

i582:
## Issue

Closes #2024.
Closes #2019.

## Checklist


src/types/resolveDescriptors.ts
```diff
@@ -1965,16 +2011,32 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
                 }
                 dependsOn.add(idText(src.contract));
             }
+
+            if (src.kind === "static_call") {
```
jeshecdom, line 2100: So, was the problem caused by the contract not detecting all referenced contracts during resolve phase (i.e., it did not detect all its children)? In this case, resolve was not looking inside the code of static function calls. What would happen if `initOf` is called inside the following cases? 
- A call to an `extends` function, like in this example:
```
extends fun doSomething(self: Int): Address {
    return contractAddress(initOf MasterVO());
}
```
I suspect that calling it like `5.doSomething()` inside `init()` will go wrong because of the `5` expression still not having a type in `CompilerContext`. I hope I am wrong.
- A call to a mutates function.
- A call to a contract method.

i582: > I hope I am wrong.

Unfortunately, no, but I really hope that such code will never be written :)

The correct solution for all such cases can only be fixed in a completely new compiler, I think. Currently our `resolveDescriptors` are a bunch of bad code

novusnota: > Unfortunately, no, but I really hope that such code will never be written :)

If some code can be written it will be written :):

Shvandre (dismissed):
LGTM

src/types/resolveDescriptors.ts
verytactical: ...and I will do it personally, out of spite!

(Not. Let's merge as is.)

PR #2039, opened 2025-02-26
feat(stdlib): add `Slice.hashData()` and `String.hashData()`

### LINKED ISSUE #1950: Add `Slice.hashData()` and `String.hashData()` functions to the stdlib
In place of the `sha256()` that works for all the tail references and strings with more than 127 bytes as well: https://github.com/tact-lang/tact/pull/1936

### LINKED ISSUE #1963: Fix the caution in `book/statements/#destructuring-assignment`
The example in the caution note is missing `Second`, it should be as follows:
```tact
struct First { nested: Second }
struct Second { field: Int }

fun example() {
    let prep = First { nested: Second { field: 42 } };
    let First { nested: Second { field: thing } } = prep;
}
```

novusnota:
A follow-up to https://github.com/tact-lang/tact/pull/1936, where we've fixed the `sha256()` for Slices and that warranted introduction of gas-efficient functions for hashing only the data of a given Slice/String.

## Issue

Closes #1950.
Closes #1963 (small typo).

## Checklist


src/stdlib/stdlib/std/internal/crypto.tact
```diff
@@ -2,6 +2,43 @@ asm extends fun hash(self: Cell): Int { HASHCU }
...
+/// Calculates and returns an `Int` value of the SHA-256 hash of the data bits from the given `Slice`, which should have a number of bits divisible by 8.
+///
+/// Unlike `sha256()`, this function is gas-efficient and **only** hashes the data of the given `Slice`, i.e. up to 1023 bits, ignoring the refs.
+///
+/// Attempts to specify a `Slice` with number of bits **not** divisible by 8 throw an exception with exit code 9: `Cell underflow`.
+///
+/// ```tact
+/// fun example() {
+///     let s: Slice = beginCell().asSlice();
+///     let fizz: Int = s.hashData(); // just the hash of the data
```
i582: maybe let's use `slice()` function for example data?

src/stdlib/stdlib/std/internal/crypto.tact
```diff
@@ -2,6 +2,43 @@ asm extends fun hash(self: Cell): Int { HASHCU }
...
+///
+asm extends fun hashData(self: Slice): Int { ONE HASHEXT_SHA256 }
+
+/// Extension function for the `String` type. Available since Tact 1.6.0.
+///
+/// Calculates and returns an `Int` value of the SHA-256 hash of the data bits from the given `String`, which should have a number of bits divisible by 8.
+///
+/// Unlike `sha256()`, this function is gas-efficient and **only** hashes up to 127 bytes of the given `String`. Using longer strings would cause collisions if their first 127 bytes are the same.
+///
+/// Attempts to specify a `String` with number of bits **not** divisible by 8 throw an exception with exit code 9: `Cell underflow`.
```
i582, line 44: is it possible?

novusnota: Yes, if one would first make a slice and then type cast it via `.asString()` to a string

```tact
dump(
    // this will blow up
    beginCell().storeUint(0, 7).asSlice().asString().hashData()
);
```

src/stdlib/stdlib/std/internal/crypto.tact
anton-trunov: good suggestion, we should give these latest additions more visibility

docs/src/content/docs/ref/core-strings.mdx
```diff
@@ -165,6 +165,28 @@ let buzz: Slice = s.toCell().asSlice();
...
+
+```tact
+extends fun hashData(self: String): Int;
+```
+
+Extension function for the [`String{:tact}`][p] type.
+
+Calculates and returns an [`Int{:tact}`][int] value of the [SHA-256][sha-2] hash of the data bits from the given [`String{:tact}`][p], which should have a number of bits divisible by $8$.
+
+Unlike [`sha256(){:tact}`](/ref/core-math#sha256), this function is gas-efficient and **only** hashes up to $127$ bytes of the given [`String{:tact}`][p]. Using longer strings would cause collisions if their first $127$ bytes are the same.
```
anton-trunov: let's not over-use links here (`String` is already linked in the previous paragraphs).
```suggestion
Unlike [`sha256(){:tact}`](/ref/core-math#sha256), this function is gas-efficient and **only** hashes up to $127$ bytes of the given string. Using longer strings would cause collisions if their first $127$ bytes are the same.
```

src/stdlib/stdlib/std/internal/crypto.tact
novusnota: Used `slice()` and also made the example showcase the shortcomings of this function

PR #2065, opened 2025-02-27
refactor: separate context

### LINKED ISSUE #2064: Store context variables separately
If there isn't too many context variables, store them separately.

At the moment we store them as a tuple in the generated FunC code

```func
(int, slice, int, slice) __tact_context_get() inline {
    return __tact_context;
}
```

and in resulting assembly we unpack the tuple onto stack

```fift
__tact_context_get PROCINLINE:<{
  __tact_context GETGLOB
  4 UNTUPLE
}>
```

and then drop the components we don't need

```fift
__tact_context_get INLINECALLDICT
2 2 BLKDROP2
```

verytactical:
## Issue

Closes #2064.

## Checklist


src/stdlib/stdlib/std/internal/context.tact
```diff
@@ -60,3 +69,43 @@ asm extends fun readForwardFee(self: Context): Int {
                       // first_frac is a value listed in config param 25
                       // of the blockchain: https://tonscan.org/config#25
 }
+
+asm fun readForwardFeeAux(rawPayload: Slice): Int {
+    // Only the self.raw (Context.raw) is important,
```
novusnota, line 74: hacky copy-paste i see :yo_yo:

Issue #2045, opened 2025-02-26
Backwards compatibility tests on macOS fail with OOM


novusnota:
See: https://github.com/tact-lang/tact/actions/runs/13551615045/job/37876307749?pr=1949

And that PR in general: https://github.com/tact-lang/tact/pull/1949. In it, macOS runners for the main test suite were running in an endless loop on map tests. Perhaps, some other test triggered such behavior:

![Image](https://github.com/user-attachments/assets/8c4b961e-290b-4300-932a-dbfc8f226532)

Only the manual cancellation of the workflow made it stop.

Rumor has it this may be related to the fuzzer :)

novusnota:
If someone else has that OOM problem locally, a temporary workaround is to run `export NODE_OPTIONS=--max_old_space_size=4096` once in the terminal where you'll be running tests. That'll increase the memory limit to 4Gb.

novusnota:
Ways to fix it: https://github.com/tact-lang/tact/pull/2066#issuecomment-2689303858

anton-trunov:
It looks like it's been fixed or at least alleviated. Feel free to reopen if not.

PR #2066, opened 2025-02-27
fix: increase memory limits to 4 Gb

verytactical:
## Issue

Towards #2045, because higher memory limits aren't a sustainable way to go forward.

## Checklist


jest-ci.config.js
```diff
@@ -2,5 +2,5 @@ const mainConfig = require("./jest.config");
 
 module.exports = {
     ...mainConfig,
-    maxWorkers: "2",
+    maxWorkers: "1",
```
novusnota:
```suggestion
    maxWorkers: "25%",
```

what if?

novusnota:
Ok, 25% didn't work either. The `receiver-precedence` tests consume too much memory at once.

This is a problem with the Sandbox + Jest combo, but we can probably mitigate it by splitting this test group into several subgroups: `it("internal receivers should handle empty messages and empty strings correctly"`.

And/or if that doesn't work, we could move the `let` statements with `SandboxContract<>` into these newly created test subgroups, because only a handful from those contracts are needed at a time.

UPD: Or just increase the memory limits.

jeshecdom:
Now that I am starting to do huge tests, I think I will have to think how to split them so that this does not happen again.

"No bulk loading of contracts in Sandbox. Do one `it` test per contract."

should be the mantra. Sorry for the problems this is causing.

novusnota:
> "No bulk loading of contracts in Sandbox. Do one `it` test per contract."

Yeah, seems like a good rule of thumb :)

Or just deploy the contract when you need it and "do not declare and deploy many contracts before or within `beforeEach()`"

novusnota:
Increasing mem limits helped the general tests, but not for Node.js 18 in backwards-compat tests.

UPD: Of course it didn't — backwards-compat tests do not use the same coverage script. One moment :)
UPD2: Now this should work.

novusnota (approved):
image

CI is green, let's merge this ASAP, properly fix #2045 later

jeshecdom (approved):

Issue #2030, opened 2025-02-26
Add tests of reachability of `initOf` and `codeOf`


jeshecdom:
We need to test if `initOf` and `codeOf` are detected anywhere in the code. Specially for obtaining addresses during `init()`.

PR #2083, opened 2025-02-28
tests: check `initOf` and `codeOf` reachability

### LINKED ISSUE #2383: Post swarm testing activities for `initof` script
Do the following: 
- Simplify generators so that they do not return arrays, since they no longer are doing exhaustive search (this will also be needed for algorithms that generate uniform trees).
- Analyze the distribution of generated AST nodes and depth of trees.

jeshecdom:
## Issue

Towards #2030.
Closes #2383.

This PR uses the script `src/test/autogenerated/gen-initof-reachability.ts` to generate ASTs for the reachability test of `initOf`.

It tries to generate all possible combinations where a 'initOf` could be placed. It chains fast-check generators in order to generate all possible combinations (up to a certain depth in function calls). Then, it compiles the ASTs, runs the contracts in Sandbox and checks that the contracts send the appropriate messages to the correct contracts and no error exit codes are produced.

The current version has the following limitations, some of which will be addressed in subsequent PRs:
- It does not generate code for `codeof`.
- It does not generate complex expressions because I would need to generate well-typed expressions. For example, it does not generate cases where there is a function that does an `initOf` and such function is called inside an integer expression:
```
fun test(): Int {
   let a = initOf Deployer();
   send(.....)
   return 10;
}

contract A {
   init(v: Int) {
       let a: Int = 45 + test() + v;
   }
}
```
At the moment, it only generates function calls that return `StateInit`, like this:
```
fun test(): StateInit {
   let a = initOf Deployer();
   return a;
}

contract A {
   init(v: Int) {
       let a: StateInit = test();
      send(......)
   }
}
```
but I plan to add more complex expressions.
- Method calls, foreach loops, initOf in receivers will be added later. 

The fuzzer can be invoked with:
`yarn gen-initof-tests`

The statistics can be invoked with:
`yarn gen-initof-tests stats`

## Checklist


jeshecdom:
This PR depends on PR: https://github.com/tact-lang/tact/pull/2311. To make the AST generator script work, I included the change in the aforementioned PR. So, probably the PRs should be merged. Opinions?

PR #2110, opened 2025-03-03
refactor(tests): split the precedence tests for internal and external receivers

jeshecdom:
## Issue

Towards #2045.

This PR splits in half the precedence tests for internal and external receivers in order to decrease memory usage.

## Checklist

### DIFF
--- a/.github/workflows/tact.yml
+++ b/.github/workflows/tact.yml
@@ -42,7 +42,8 @@ jobs:
           # Test some specific things for backwards compatibility.
           # It's important to restrain from using too much of Node.js 22+ features
           # until it goes into maintenance LTS state and majority of users catches up
-          yarn cross-env NODE_OPTIONS=--max_old_space_size=4096 jest isSubsetOf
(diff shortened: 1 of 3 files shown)


jeshecdom:
Mmmm.... it is now failing after the memory decrease. However, this time it does not say which test is failing... or at least I cannot find a reference to `receiver-precedence` in the error messages. Do you think that I will need to split the tests so that it opens only one contract per test?

novusnota:
> Mmmm.... it is now failing after the memory decrease. However, this time it does not say which test is failing... or at least I cannot find a reference to `receiver-precedence` in the error messages. Do you think that I will need to split the tests so that it opens only one contract per test?

Not sure. It also seems that the `yarn build` command is one of the culprits, especially the `tsc` part of it.

Perhaps, some bigger tests have to be split up OR we should approach compilation of `.ts` test files differently — for example, by using [ts-blank-space](https://github.com/bloomberg/ts-blank-space) or experimental type stripping of Node.js. This will avoid problems by simply not having to compile lots of `.ts` tests at once — they could be processed on the fly.

PR #2221, opened 2025-03-03
fix: remove @ton/core from ast.ts

### LINKED ISSUE #1459: Remove `@ton/core` imports
We can't really make a schema out of opaque third-party types.

`Cell` and `Address` have a JS-native representation (in worst case, base64 string).

Replace `@ton/core` imports in `ast/` with JS-native types.

verytactical:
## Issue

Closes #1459.

### DIFF
--- a/src/ast/ast.ts
+++ b/src/ast/ast.ts
@@ -1,4 +1,3 @@
-// import type { Address, Cell, Slice } from "@ton/core";
 import type { SrcInfo } from "../grammar/src-info";
 import type { RelativePath } from "../imports/path";
 import type { Language } from "../imports/source";
--- a/src/ast/ast.ts
+++ b/src/ast/ast.ts
@@ -1,4 +1,4 @@
-import type { Address, Cell, Slice } from "@ton/core";
+// import type { Address, Cell, Slice } from "@ton/core";
 import type { SrcInfo } from "../grammar/src-info";
 import type { RelativePath } from "../imports/path";
 import type { Language } from "../imports/source";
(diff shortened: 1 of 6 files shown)


src/ast/ast.ts
```diff
@@ -1,4 +1,4 @@
-import type { Address, Cell, Slice } from "@ton/core";
+// import type { Address, Cell, Slice } from "@ton/core";
```
anton-trunov: commented code

src/ast/util.ts
```diff
@@ -76,7 +76,7 @@ export const getAstUtil = ({ createNode }: FactoryAst) => {
     function makeCellLiteral(c: Cell, loc: SrcInfo): A.AstCell {
         const result = createNode({
             kind: "cell",
-            value: c,
+            value: c.toString(),
```
jeshecdom, line 79: These changes introduce a slight inefficiency in the interpreter code. If you go to the interpreter `src/optimizer/interpreter.ts`, in the following switch cases of the corresponding method:

In `interpretStaticCall()`:
- cell
- slice
- rawSlice
- address
- newAddress

In `interpretMethodCall()`:
- asComment


Those cases first create a cell/slice/address and then call `util.makeXLiteral` which would transform them back into a string after the change. Is this inefficiency acceptable? A reimplementation of those switch cases would simply return the strings instead of creating the cell/slice/address. This would imply adding a version of `util.makeXLiteral` that receives the strings instead of cell/slice/address.

Issue #1909, opened 2025-02-20
Add doc comments to `cells.tact`


anton-trunov:
Some entries already come with doc comments, but not all. These are missing doc comments:

- [ ] `beginCell`
- [ ] `storeInt`
- [ ] `storeUint`
- [ ] `storeBool`
- [ ] `storeRef`
- [ ] `endCell`
- [ ] `refs` for `Cell` and `Slice`
- [ ] `bits` for `Cell` and `Slice`
- [ ] `beginParse`
- [ ] `loadRef`
- [ ] `loadBits`
- [ ] `preloadBits`
- [ ] `loadInt`
- [ ] `preloadInt`
- [ ] `loadUint`
- [ ] `loadBool`
- [ ] `skipBits`
- [ ] `endParse`
- [ ] `empty`
- [ ] `dataEmpty`
- [ ] `refsEmpty`
- [ ] `asSlice` for `Builder` and `Cell`
- [ ] `emptyCell`
- [ ] `emptySlice`

PR #2246, opened 2025-03-04
feat(stdlib): add doc comments to `cells.tact`

i582:
Towards #1909


docs/src/content/docs/ref/core-cells.mdx
anton-trunov, line 1: let's correct the signature
```tact
extends fun storeSlice(self: Builder, slice: Slice): Builder;
```

i582: Fixed

anton-trunov (approved):

PR #2280, opened 2025-03-05
refactor(stdlib): replace `MUL` with `AND` in `send()`

### LINKED ISSUE #2279: `send()` uses `MUL` instead of boolean `AND` for logic
In the `send()` function from stdlib, there's a place where two boolean (`0`/`-1`) values are computed and then their product is taken to determine whether both of them were `-1` or not.
https://github.com/tact-lang/tact/blob/163adfb5d20d27e84f8b0b5213089ebaad6f5ec0/src/stdlib/stdlib/std/internal/send.tact#L212-L218

It uses `MUL` instruction that performs simple arithmetic multiplication, and the comment says that `-1 * -1` wraps back to `-1`.
The thing is that `-1 * -1 = 1`, and even though `IFELSE` operation used later will still work with this, it's not a good thing that we calculate boolean value by multiplying numbers and also using non-boolean (`0` or `-1`) number for a logic instruction.

There's an `AND` instruction for this purpose that will result `-1` instead of `1` in this case and it consumes exactly the same gas and is exactly the same size as `MUL`.

Gusarich:
## Issue

Closes #2279.


anton-trunov (approved):

Shvandre (approved):
Awesome change, huge contribution!

novusnota:
A bit unrelated, but it'd be nice if you could also edit the CHANGELOG entry in v1.6.1 for wildcards — the correct PR number there is 2188, not 2189 :)

anton-trunov:
@novusnota can you please push the change to this branch?

anton-trunov:
ah, never mind, the CI done here, merging

novusnota:
> @novusnota can you please push the change to this branch?

I'll send a small follow-up PR then :)

PR #2300, opened 2025-03-06
fix: don't give an error for a small enough type for bounce receiver

### LINKED ISSUE #2294: Remove "small enough for bounce receiver" error message
```shell
bounced (msg: bounced<GetState>) {

Error: contracts/root/root. tact: 396:5: This message is small enough for bounce receiver, you need to remove bounced modifier.
```

i582:
## Issue

Closes #2294.

## Checklist

### DIFF
--- a/src/types/test/small-bounced-type.tact
+++ b/src/types/test/small-bounced-type.tact
@@ -1,4 +1,5 @@
 primitive Int;
+primitive Bool;
 
 trait BaseTrait {}
 
@@ -7,8 +8,16 @@ message Add {
     amount: Int;
 }
 
+message Is {
+    val: Bool;
+}
+
 contract Test {
     bounced(msg: bounced<Add>) {
+        // ...
+    }
+
+    bounced(msg: bounced<Is>) {
         // ...
     }
 }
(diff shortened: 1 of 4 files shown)


anton-trunov (dismissed):

src/types/test/small-bounced-type.tact
```diff
@@ -0,0 +1,14 @@
+primitive Int;
+
+trait BaseTrait {}
+
+message Add {
+    queryId: Int;
```
novusnota: Not that small, Int has 257 bits :)

Try booleans or `as uintX`

src/types/test/small-bounced-type.tact
```diff
@@ -0,0 +1,14 @@
...
+    amount: Int;
```
anton-trunov:
```suggestion
    queryId: Int as uint8;
```

src/types/test/small-bounced-type.tact
i582: Added single `Bool`

src/types/test/small-bounced-type.tact
i582: Done

PR #2311, opened 2025-03-06
fix: union with parsedModules

### LINKED ISSUE #2310: Compiling ASTs with references to stdlib causes errors during `resolveDescriptors`
The error is caused by the second step in `precompile`, at the first line of the `openContext` function. When passing the AST through the `parsedModules` optional argument in `openContext`, the `openContext` function drops the `sources` computed in the first step of `precompile` (which contain the `stdlib` declarations): 

```
const modules = parsedModules ?? parseModules(sources, parser);
```
This causes missing reference errors in `resolveDescriptors` later.

It seems to me that instead of dropping `sources`, the function `openContext` should take the union of `sources` plus the optional argument `parsedModules`. Like this:
```
const parsedSources = parseModules(sources, parser);
const modules = typeof parsedModules === "undefined" ? parsedSources : [...parsedSources, ...parsedModules];
```

### LINKED ISSUE #2429: Autogenerate `makeX` functions from AST types
Use the `babel` tool to autogenerate the `makeX` functions. Focus on all the cases for `AstExpression` and `AstStatement` which are the most common used cases in fuzzers.

Place the generated `makeX` functions in a new utility factory, because the current `AstUtil` is used in other places in the code base. Eventually we will need to move the code base into using the new utility factory and delete `AstUtil`.

Add another `yarn` command that runs the script and creates the `.ts` file with all the `makeX` functions.

Related to #1460 and #1327.

jeshecdom:
## Issue

Closes #2310.
Closes #2429.

To run autogeneration of `makeX` functions, execute:
`yarn gen:make-funs`
It will create the file `src/ast/generated/make-factory.ts`.

## Checklist


src/context/store.ts
```diff
@@ -63,7 +65,11 @@ export function openContext(
     parser: Parser,
     parsedModules?: A.AstModule[],
 ): CompilerContext {
-    const modules = parsedModules ?? parseModules(sources, parser);
+    const parsedSources = parseModules(sources, parser);
```
verytactical: I would rather make `parsedModules` required, and pass `parseModules(sources, parser)` in regular build.

src/test/pipeline/precompile-ast.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+    //
+    // contract Test {
+    //    f: StateInit;
+    //
+    //    init() {
+    //        self.f = initOf Test();
+    //    }
+    // }
+
+    function makeModule(): A.AstModule {
```
verytactical: It seems like this should be a set of definitions in `src/ast/no-src-factory.ts` or even a reuse of existing code for this.

It will be extremely tedious to update manually in case AST definitions change. This code should eventually be generated from AST types.

verytactical (changes requested):
^

src/test/pipeline/precompile-ast.spec.ts
jeshecdom: Agreed. I will migrate all the `makeX` functions into that new factory, because this will be a common problem in all PRs involving testing: each one of them defines their own `makeX` functions.

Is there an example on how to generate code for the `makeX` functions just from the AST types? Do you mean like a pre-compilation step, similar to `yarn gen` so that all `makeX` functions are generated?

jeshecdom: Actually, there was a `AstUtil` factory. I think I'll place the `makeX` functions there.

src/context/store.ts
jeshecdom: The suggested change caused fixes in several `spec` files.

src/test/pipeline/precompile-ast.spec.ts
jeshecdom: I added autogeneration of all makeX functions. Please have a look.

verytactical: Thanks!

package.json
```diff
@@ -56,6 +56,7 @@
     "all": "yarn clean && yarn gen && yarn build && yarn coverage && yarn lint:all",
     "next-version": "ts-node version.build.ts",
     "random-ast": "ts-node ./src/ast/random-ast.ts",
+    "gen-make-functions": "ts-node ./src/ast/generated/gen-make-functions-script.ts && yarn prettier -w ./src/ast/generated/make-factory.ts",
```
verytactical: I suggest calling this in line with other scripts, `gen:ast` or something like that.

src/ast/generated/gen-make-functions-script.ts
```diff
@@ -0,0 +1,301 @@
...
+    const createNodeArgsForMake = generalParams.map(({ id, type: _ }) =>
+        t.objectProperty(id, t.identifier(`p_${id.name}`)),
+    );
+    const createNodeArgsForMakeDummy = generalParams.map(({ id, type: _ }) =>
+        id.name === "loc"
+            ? t.objectProperty(id, t.identifier(emptySrcInfo))
+            : t.objectProperty(id, t.identifier(`p_${id.name}`)),
+    );
+    const funReturnType = t.tsTypeReference(
+        t.tSQualifiedName(t.identifier(astNamespace), id),
```
verytactical: tSQualifiedName -> tsQualifiedType

src/ast/generated/gen-make-functions-script.ts
```diff
@@ -0,0 +1,301 @@
...
+    }[] = [];
+
+    // If there is no loc field,
+    // the makeDummy function cannot be created
+    const makeDummy = rawFieldsArray.some(
+        (f) => f.key.type === "Identifier" && f.key.name === "loc",
+    );
+
+    for (const field of rawFieldsArray) {
+        if (field.key.type !== "Identifier") {
```
verytactical: nit: there is `t.isIdentifier` that returns a guard type, and might be more convenient

src/ast/generated/gen-make-functions-script.ts
```diff
@@ -0,0 +1,301 @@
...
+        .filter((stmt) => stmt?.type === "TSTypeAliasDeclaration");
+
+    const declNames = new Set(decls.map((decl) => decl.id.name));
+
+    const finalFunctions: { name: string; code: string }[] = [];
+
+    // Extract from the declarations all the union types
+    const unionTypes = decls
+        .map((decl) => decl.typeAnnotation)
+        .filter((decl) => decl.type === "TSUnionType");
```
verytactical: Guard type likely doesn't help us understand it's a union type in the body of the `for` loop below.

For most means and purposes, `.filter` is worse than just a simple `if ()` for this reason.

src/ast/generated/gen-make-functions-script.ts
```diff
@@ -0,0 +1,301 @@
...
+            } else {
+                // Currently unsupported
+                throw new Error(
+                    "TSNamedTupleMember is currently not supported in TSTupleType",
+                );
+            }
+        }
+
+        default:
+            return typ;
```
verytactical, line 302: I would rather throw an exception here, because there might be (or will be) types with identifiers inside that are not (will not be) handled.

Throwing an exception here would be slightly safer.

src/generator/writers/writeExpression.spec.ts
```diff
@@ -71,11 +72,14 @@ const golden: string[] = [
...
+        const sources: Source[] = [
+            { code: code, path: "<unknown>", origin: "user" },
+        ];
         let ctx = openContext(
             new CompilerContext(),
-            [{ code: code, path: "<unknown>", origin: "user" }],
+            sources,
             [],
-            getParser(ast),
+            parseModules(sources, getParser(ast)),
```
verytactical, line 82: Pleasure to my eyes :heart:

verytactical (dismissed):

src/ast/generated/gen-make-functions-script.ts
jeshecdom: I did not fully understand the comment. However, I changed all direct type comparisons to calls for their corresponding guard function `t.isX` (with only one exception in the switch case for `TSTupleType` in function `qualifyType`, for which I added a comment in the code).

src/ast/generated/gen-make-functions-script.ts
jeshecdom: Done

src/ast/generated/gen-make-functions-script.ts
jeshecdom: Done

src/ast/generated/gen-make-functions-script.ts
jeshecdom: Done

package.json
jeshecdom: Done

PR #2312, opened 2025-03-06
chore: replace "change-case" library with own implementation

### LINKED ISSUE #1436: Remove dependency on `change-case`
Converting to snake-case and PascalCase doesn't require a library.

i582:
## Issue

Closes #1436.

### DIFF
--- a/src/utils/change-case.ts
+++ b/src/utils/change-case.ts
@@ -1,211 +1,62 @@
-// The MIT License (MIT)
-//
-// Copyright (c) 2014 Blake Embrey (hello@blakeembrey.com)
-//
-// Permission is hereby granted, free of charge, to any person obtaining a copy
-// of this software and associated documentation files (the "Software"), to deal
-// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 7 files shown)


src/utils/change-case.ts
Danil42Russia, line 1: what about adding tests? :(

src/utils/change-case.ts
```diff
@@ -0,0 +1,211 @@
...
+        prefix + words.map(transform).join(options?.delimiter ?? "") + suffix
+    );
+}
+
+/**
+ * Convert a string to snake case (`foo_bar`).
+ */
+export function snakeCase(input: string, options?: Options) {
+    return noCase(input, { delimiter: "_", ...options });
+}
```
anton-trunov: Let's keep only `pascalCase` and `snakeCase` and their deps

src/utils/change-case.ts
i582: added

src/utils/change-case.ts
i582: redone

PR #2314, opened 2025-03-06
fix: no empty init when empty contract parameters are present

### LINKED ISSUE #2297: Contract with empty contract parameter list should not generate `init`
Otherwise, we get contract storage save operations for receivers that cannot possibly change the contract state. It should be empty but it's not because of the implicit lazy deployment bit.

This should work:

```
message Message {
    i: Int as uint64;
    b: Int as uint128;
}

contract Foo() {
    receive(msg: Message) {
        msg.toSlice();
    }
}
```

verytactical:
## Issue

Closes #2297.

### DIFF
--- a/src/grammar/next/grammar.gg
+++ b/src/grammar/next/grammar.gg
@@ -118,7 +118,7 @@ Contract =
     attributes:ContractAttribute*
     keyword<"contract">
     name:Id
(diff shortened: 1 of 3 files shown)


novusnota (approved):
pleasing linters goes brr

anton-trunov (approved):

PR #2315, opened 2025-03-06
refactor: `require()` now uses exit codes starting from 1024

### LINKED ISSUE #2245: require() should use smaller exit codes
It is more convenient and can save us some gas.

Shvandre:
## Issue

Closes #2245 .


src/cli/tact/e2e.spec.ts
```diff
@@ -184,14 +184,17 @@ describe("tact --config config.json", () => {
     });
 });
 
+/*
+ * This tests do not work in JavaScript Debug Terminal, as they use stderr.
+ * JS debug terminal adds redundant output to stderr, which breaks the tests.
+ */
```
anton-trunov:
```suggestion
```

anton-trunov (changes requested):
Add a test demonstrating

```tact
require(false, "second");
require(false, "second");
```

is handled correctly.

anton-trunov:
One more thought: this PR breaks backwards compatibility with the current behavior of require: `require(cond, "error message")` in several places will have the same exit code because the string is shared.

novusnota:
> One more thought: this PR breaks backwards compatibility with the current behavior of require: `require(cond, "error message")` in several places will have the same exit code because the string is shared.

I think that's reasonable. For posterity though, I'll push the updated docs for `require()` in a moment since I've almost finished them :)

anton-trunov:
I don't think we need one more source of implicitness in the language, especially the one that breaks backwards compatibility

PR #2327, opened 2025-03-07
fix(docs): compiler upgrades with npm or yarn v4

### LINKED ISSUE #2301: Check compiler upgrade instructions in the docs
The goal is to check that commands given to upgrade the compiler work: https://docs.tact-lang.org/book/compile/#upgrades

Besides, that the installation of newer versions also work, without any upgrade commands.

Plus, we need to support Yarn v4 there, which uses `yarn up` and not `yarn upgrade`

novusnota:
Uses idiomatic `npm` package version upgrade command over the installation one.

## Issue

Closes #2301.


novusnota:
Since we've fixed compiler updates with removing the `postinstall` and/or other lifecycle scripts, this PR will be simplified: down to the small fix in the docs and, perhaps, keeping the tact-template setup action, just in case.

novusnota:
Ok, since nobody is reviewing I've removed the action — let's just merge the fix to the docs :)

Issue #820, opened 2024-09-13
Left and right shifts in FunC 0.4.4 are buggy and need to be replaced with `asm`-functions


anton-trunov:
The example from the issue https://github.com/ton-blockchain/ton/issues/1022 should obviously map to Tact, so we need to check if Tact has the same bug and add a workaround for it (maybe even the workaround described in the linked issue).

```tact
contract Repro {
    get fun test(i: Int): Bool {
        let mask = 1 << i;
        //let mask = (1 << i) / 1; <-- this is a workaround
        return mask == 1;
    }
}

// test(0) returns false, while true is expected
```

This has been fixed in FunC v0.4.5, in the following PR: https://github.com/ton-blockchain/ton/pull/1262

PR #2340, opened 2025-03-09
feat: add updated fuzzer from tact-check repository

### LINKED ISSUE #2641: Generate unary, binary and ternary operators in the expressions for fuzzer in PR #2340
Currently, the fuzzer in PR #2340 does not generate expressions involving unary, binary and ternary operators. We need to adapt the algorithm for uniform tree generation to produce well-typed expressions.

- [X] Rewrite the uniform tree generation algorithm to compute the counts table using dynamic programming.

- [X] Adapt the algorithm so that the "structural tree grammar rules" are parametric over the types. In this way we can use the algorithm to produce well-typed expressions. Hint: note that the typing rules, when read backwards are precisely the grammar rules for the structural trees:
```
n: Int    m: Int
-----------------     ======>    E (Int) ->   + E (Int) E (Int)      "To produce an expression from non-terminal E (of type Int), generate the + symbol and two expressions of type Int"
   n + m: Int                                 | .....    <--- other cases of the non-terminal E 
   

n: Int    m: Int
-----------------    =======>    E (Bool) -> <= E (Int) E (Int)
  n <= m : Bool                              | ......
```
At the moment, the type system for Tact is finite; for example, the optional `?` and `map` type constructors do not nest. So, the above method should suffice.

- [X] Add the generators for binary, unary and ternary operators.

### LINKED ISSUE #2877: Integrate the expression generator with rest of imported fuzzer
Currently, the expression generator is isolated from the rest of generators in the imported fuzzer. We need to join the expression generator with the rest to be able to generate declarations that later can be called in expressions, like structs, functions, contracts and traits.

Mell0r:
Closes #2641
Closes #2877


Mell0r:
Right now there is the problem in one of the tests in the fuzzer - compilation.spec.ts:
**Error from the compilator:** [INTERNAL COMPILER ERROR]: Interval [0, 0[ is empty or out of source bounds (0)
**Stacktrace:**
      at throwInternalCompilerError (src/error/errors.ts:98:11)
      at show (src/grammar/src-info.ts:185:39)
      at Object.show [as toString] (src/grammar/src-info.ts:227:29)
      at Object.toString [as getLineAndColumnMessage] (src/grammar/src-info.ts:258:35)
      at getLineAndColumnMessage (src/error/errors.ts:111:62)
      at throwNonFatalErrorConstEval (src/optimizer/interpreter.ts:60:24)
      at Interpreter.throwNonFatalErrorConstEval [as interpretStaticCall] (src/optimizer/interpreter.ts:1562:21)        
      at Interpreter.interpretStaticCall [as interpretExpressionInternal] (src/optimizer/interpreter.ts:924:29)
      at interpretExpressionInternal (src/optimizer/interpreter.ts:770:18)
      at Interpreter.code [as handleStackOverflow] (src/optimizer/interpreter.ts:1933:20)
      at Interpreter.handleStackOverflow (src/optimizer/interpreter.ts:769:21)
      at interpretExpression (src/optimizer/constEval.ts:221:32)
      at writeExpression (src/generator/writers/writeExpression.ts:181:45)
      at writeStatement (src/generator/writers/writeFunction.ts:186:40)
      at writeReceiverBody (src/generator/writers/writeRouter.ts:502:23)
      at writeReceiverBody (src/generator/writers/writeRouter.ts:492:9)
      at Writer.handler [as inIndent] (src/utils/Writer.ts:9:9)
      at Writer.inIndent [as inBlock] (src/utils/Writer.ts:15:14)
      at WriterContext.inBlock (src/generator/Writer.ts:303:30)
      at inBlock (src/generator/writers/writeRouter.ts:475:10)
      at writeBouncedReceiver (src/generator/writers/writeRouter.ts:427:13)
          at Array.forEach (<anonymous>)
      at forEach (src/generator/writers/writeRouter.ts:426:33)
      at Writer.handler [as inIndent] (src/utils/Writer.ts:9:9)
      at Writer.inIndent [as inBlock] (src/utils/Writer.ts:15:14)
      at WriterContext.inBlock (src/generator/Writer.ts:303:30)
      at inBlock (src/generator/writers/writeRouter.ts:419:10)
      at src/generator/writers/writeContract.ts:422:35
      at Writer.handler [as inIndent] (src/utils/Writer.ts:9:9)
      at Writer.inIndent [as inBlock] (src/utils/Writer.ts:15:14)
      at WriterContext.inBlock (src/generator/Writer.ts:303:30)
      at inBlock (src/generator/writers/writeContract.ts:404:14)
      at handler (src/generator/Writer.ts:246:17)
      at WriterContext.handler [as body] (src/generator/Writer.ts:236:9)
      at body (src/generator/Writer.ts:245:18)
      at WriterContext.handler [as fun] (src/generator/Writer.ts:187:9)
      at WriterContext.fun [as main] (src/generator/Writer.ts:244:14)
      at main (src/generator/writers/writeContract.ts:347:10)
      at writeAll (src/generator/writeProgram.ts:437:22)
      at writeAll (src/generator/writeProgram.ts:55:5) <--- **first function called outside of the fuzzer and inside the compilator**
      at compile (fuzzer/test/testUtils.ts:62:20)
      at fuzzer/test/compilation.spec.ts:47:21
      at withNodeFS (fuzzer/src/util.ts:28:9)
      at compileProgram (fuzzer/test/compilation.spec.ts:30:5)
      at AsyncProperty.run (node_modules/fast-check/lib/check/property/AsyncProperty.generic.js:49:28)
      at asyncRunIt (node_modules/fast-check/lib/check/runner/Runner.js:37:21)
      at checkAsyncProperty (fuzzer/src/util.ts:100:5)
      at compileAndCheckProperty (fuzzer/test/compilation.spec.ts:111:17)
      at fuzzer/test/compilation.spec.ts:121:25
          at async Promise.all (index 81)

As i understand, the problem is with provioding loc to all the AST items. Before the last commit i used dummySrcInfo from compilator code for all 'loc' fields, and now i tried to replace it with my own dummySrcInfoPrintable (./fuzzer/src/utils.ts, 236 line) with non-empty line interval, but error still persists. As i doesn't provide a source code and compile ast tree directly, in the place of the error (./src/grammar/src-info.ts, 186 line) the 'str' - source code itself - is empty, so the interval is empty too. But how can it not be?

jeshecdom:
Hello. Try to define your `dummySrcInfoPrintable` like this:
```
const dummySrcInfoPrintable = getSrcInfo(" ", 0, 0, null, "user");
```
Note that the first parameter to `getSrcInfo` is not an empty string, but has a single space character (you could add as many space characters as you like). The error is caused because using an empty string as first parameter to `getSrcInfo` produces invalid limits in the second and third parameters: an empty string does not have a 0th character.

fuzzer/src/generators/constant.ts
```diff
@@ -0,0 +1,128 @@
+import type {
+    ConstantDecl as AstConstantDecl,
+    ConstantDef as AstConstantDef,
+    ConstantAttribute as AstConstantAttribute,
+    Expression as AstExpression,
+} from "../../../src/ast/ast";
```
anton-trunov:
```suggestion
import type * as Ast from "../ast/ast";
```

fuzzer/src/formatter.ts
```diff
@@ -0,0 +1,599 @@
+import type * as Ast from "../../src/ast/ast";
+
+/**
+ * PrettyPrinter class provides methods to format and indent Tact code.
+ */
+class PrettyPrinter {
```
jeshecdom: Let us delete this file and update all imports so that they refer to the pretty printer in `src/ast/ast-printer.ts` to avoid code duplication. Or is there a reason to have a special pretty printer for the fuzzer?

fuzzer/src/generators/expression.ts
```diff
@@ -0,0 +1,990 @@
...
+    constValue?: string,
+): fc.Arbitrary<string> {
+    return constValue === undefined
+        ? nonEmpty
+            ? fc.string({ minLength: 1 })
+            : fc.string()
+        : fc.constantFrom(constValue);
+}
+
+export function generateSimplifiedString(
```
jeshecdom: This method is going to be deleted once we merge with main. `SimplifiedString` was recently deleted from ASTs.

fuzzer/src/generators/expression.ts
```diff
@@ -0,0 +1,990 @@
...
+            kind: fc.constant("static_call"),
+            function: fc.constantFrom(this.name),
+            args: packArbitraries(this.args),
+            id: fc.constant(this.idx),
+            loc: fc.constant(dummySrcInfoPrintable),
+        });
+    }
+}
+
+// export namespace OpUnary {
```
jeshecdom: Since these generators are commented out, the generated expressions do not contain binary and unary operators, nor the ternary operator `? _ : _`. Do you know the reason for commenting this code out? Probably because it is problematic to generate well-typed expressions when those operators are added?

jeshecdom (commented):
Overall, it looks good. There are some minor details like the use of optional arguments in functions and the use of classes, but I think this is acceptable, since this code is for testing. 

What I am more interested in is how good the generators are; for example, the ones for expressions. What is the distribution of the generated expressions? To tackle this question, for the moment let us focus on the expression generators.

We need to add a statistics function, probably in the `expression.spec.ts` file. The statistics function should count the frequencies of structurally different trees produced by the generators. By that I mean the following:

- Given an AST `T` generated by the expression generator, traverse `T` in preorder and only keep the `kind` field of each node (for function calls also keep the name of the function). For example, if the AST is:
```
      {kind: "static_call", function: "do_something", ......}  
                        |                                 |    
The children are -->    |                                 |    
its arguments           |                                 |                                   
                        |                                 |
         {kind: "number", value: 5n, ...}       {kind: "string", value: "foo", ...}
```
Then, after keeping only the `kind` field and function names, and traversing in preorder, we get the list:

"static_call_do_something", "number", "string"

- Join the string list into a single string (use a special symbol as separator, like @):

"static_call_do_something@number@string"

- Keep a map with the counts for the produced strings.

- Additionally, register the size of each produced tree (i.e., how many nodes it has), and its height (i.e., maximum distance from the root of the tree to a leaf node).

- After producing, say 50,000 trees (or more). Dump all those counts into a file, so that we can then produce plots from the data. It would also be nice if we can produce the plots automatically, but let us postpone that for later, what matters right now is the counts.

fuzzer/src/generators/expression.ts
Mell0r: This is exactly my next task on this fuzzer - to add generation of non trivial arithmetic and logic expressions and write additional tests to them, hopefully to be able to reproduce [this](https://github.com/tact-lang/tact/issues/820) bug.

jeshecdom:
After meeting with @Mell0r, we need to add the extensions described in #2721.

PR #2341, opened 2025-03-10
docs(WIP / DRAFT!): web versions of some of the whitepapers

### LINKED ISSUE #2331: Web version of whitepapers
It's nice to have documents defining everything we're working with in PDF, but it'd be nicer to send links to particular parts of them instead of screenshots.

https://ton-blockchain.github.io/docs/#/docs

https://github.com/ton-blockchain/ton/tree/0439613bff2cc8c4cc6f02bb43c5592d39a1c6e2/doc

Add `tvm.pdf` and `tblkch.pdf` as separate pages of Tact documentation.

novusnota:
This is still pretty much a draft, because many headers need to be placed correctly, footnotes fixed, final math issues resolved, etc.

PDFs converted:
* tvm.pdf
* tblkch.pdf

Initial approach used `.mdx`, now trying `.html` outputs instead.

## Issue

Closes #2331.

### DIFF
--- a/dev-docs/CHANGELOG.md
+++ b/dev-docs/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Docs
 
-- Add descriptions for `&&=`, `||=`, `>>=` and `<<=` augmented assignment operators: PR [#2328](https://github.com/tact-lang/tact/pull/2328)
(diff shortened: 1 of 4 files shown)


novusnota:
self-ping

anton-trunov:
ping from myself this time

novusnota:
Approach applied in this PR (pandoc to markdown) has to be adjusted or changed. Otherwise, it requires too much manual work to make it somewhat presentable.

Alternative solution, which might work better, is to convert to an `.html` and then use the Starlight/Astro's feature of special pages. That had mixed results in local testing, but looked rather promising, so I'll might do just that.

Issue #2044, opened 2025-02-26
Document `BasechainAddress` struct and related functions


novusnota:
Introduced in https://github.com/tact-lang/tact/pull/2035. It had some doc comments, but not the documentation itself.

- [ ] struct `BasechainAddress` (available since 1.6.0)
- [ ] `emptyBasechainAddress` (available since 1.6.0)
- [ ] `newBasechainAddress` (available since 1.6.0)
- [ ] `contractBasechainAddress` (available since 1.6.0)
- [ ] `storeBasechainAddress` (available since 1.6.0)
- [ ] add the new functions to CHANGELOG

novusnota:
Add docs for #2097 too.

novusnota:
And a new suggestion on the gas best practices page (#2342)

novusnota:
Needs #2391. Until then, I'll be completing #2375.

UPD: Ok, this is unblocked, doing it quickly.
UPD: I'll send PR in the morning — turns out we didn't add the respective load/skip functions for Slices when introducing the Builder.storeBasechainAddress function. That needs to change, which is why I've added this to the 1.6.4 milestone.

anton-trunov:
No new functionaly in PRs fixing things please

PR #2342, opened 2025-03-10
docs: gas best practices page

### LINKED ISSUE #1868: Gas best practices page
### General considerations

- If you don't need lazy initialization, don't use `init()` and contract storage variables, but prefer to define the contract storage via contract parameters. If a contract uses lazy initialization with `init()` it cannot benefit from the contract storage write optimization, so any receiver, even if it does not directly modify the contract storage, would always save the contract data at the end of its execution
- Use `BasechainAddress` instead of `Address` to manipulate temporary (runtime-only) addresses, since basechain addresses can be handled more efficiently (introduced in #2035). See the Jetton contracts for a use case. A frequent use-case is checking sender addresses for validity.
- Do not use `Deployable` trait unless your custom solution would be as gas-expensive.
- Try adding `inline` annotation to some functions. It might result in contract code size increase, so it needs benchmarking to decide the usefulness of it on a case-by-case basis.
- Tact docs annotate gas-expensive functions with a special badge, check if using an expensive function is unavoidable. 
- Using `asm`-functions can sometimes save you a lot of gas. 
- Prefer not to manipulate strings on-chain.
- Prefer not to use branches if possible, use arithmetic. Do `1 + sign(msg.forwardTonAmount)` instead of  `msg.forwardTonAmount > 0 ? 2 : 1` Note: This trick will work only if the value you compare with zero is non-negative

### Receiving messages

- Prefer binary receivers over text ones (#1867).
- Avoid using contract methods if those are not necessary. If a contract method does not access any contract variable, move it outside the contract and make a module-level function. Passing the contract storage variables makes contract methods more gas expensive and also make the intent less obvious in some cases.
- Prefer `sender()` to `context().sender`.
- Prefer `throwUnless()` to `require()`. This is subject to change in the future, when `require` is optimized.
 
### Sending messages

- Prefer the `deploy` function (introduced in #1832) to `send` if your contract sends a message that can deploy a contract.
- Prefer the `message` function to `send` if your message does *not* deploy a contract.
- Do not use `self.forward`, `self.reply`, `self.notify` functions as those are more gas expensive than the simple `send` function with the corresponding parameters.

### Examples

You can find some examples with those recommendations applied in our benchmarks:
- [Jetton Wallet](https://github.com/tact-lang/tact/blob/main/src/test/benchmarks/contracts/jetton_wallet.tact) contract
- [Jetton Minter Discoverable](https://github.com/tact-lang/tact/blob/main/src/test/benchmarks/contracts/jetton_minter_discoverable.tact) contract
- [Escrow](https://github.com/tact-lang/tact/blob/main/src/test/benchmarks/contracts/escrow.tact) contract

Related issue: #1137

novusnota:
## Issue

Closes #1868.

## Preview (initial)

image


dev-docs/CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Docs
 
-- Add descriptions for `&&=`, `||=`, `>>=` and `<<=` augmented assignment operators: PR [#2328](https://github.com/tact-lang/tact/pull/2328)
+- Added descriptions for `&&=`, `||=`, `>>=` and `<<=` augmented assignment operators: PR [#2328](https://github.com/tact-lang/tact/pull/2328)
+- Added gas best practices page: PR [#TBD](https://github.com/tact-lang/tact/pull/TBD)
```
```suggestion
- Added gas best practices page: PR [#2342](https://github.com/tact-lang/tact/pull/2342)
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+### Contract parameters vs. `init()` and contract fields
+
+If you require some on-chain deployment logic that runs just once after the contract's deployment is completed, then use the lazy initialization strategy provided by the [`init(){:tact}` function](/book/contracts#init-function). It uses an extra bit in the contract's persistent state, runs extra checks upon receiving any message, and disables storage write optimizations of Tact compiler.
+
+However, most contracts do not require lazy initialization and define their initial data only when the contract is first deployed, with everything for it prepared off-chain. Therefore, prefer using [contract parameters syntax](/book/contracts#parameters) to define the data of the contract's persistent storage that way.
+
+```tact
+contract ContractParams(
+    // Persistent state variables declared via the contract parameters syntax
+    val1: Int, // Int, implicitly serialized to a 257-bit signed
```
anton-trunov:
```suggestion
    val1: Int as uint64, // `as`-annotations are supported
```

novusnota:
The `BasechainAddress` part was temporarily omitted and will be brought back in a separate PR as per #2044

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+import { ContractParams } from '../wrappers/ContractParams';
+
+// Various utility libraries
+import { NetworkProvider } from '@ton/blueprint';
+import { toNano } from '@ton/core';
+
+// Default deployment function / script of the Blueprint
+export async function run(provider: NetworkProvider) {
+  const contract = provider.open(await ContractParams.fromInit(
+    420,                 // `val1` in the contract
```
anton-trunov:
```suggestion
    42,                   // `val1` in the contract
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+import { ContractParams } from '../wrappers/ContractParams';
+
+// Various utility libraries
+import { NetworkProvider } from '@ton/blueprint';
+import { toNano } from '@ton/core';
+
+// Default deployment function / script of the Blueprint
+export async function run(provider: NetworkProvider) {
+  const contract = provider.open(await ContractParams.fromInit(
+    42,                   // `val1` in the contract
```
novusnota:
```suggestion
    42,                  // `val1` in the contract
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+:::note
+
+  Tact compiler is continuously evolving, so some points below may become outdated and removed or replaced. Please check this page periodically for updates.
+
+:::
+
+## General considerations
+
+### Contract parameters vs. `init()` and contract fields
```
anton-trunov:
```suggestion
### Prefer contract parameters to `init()` and contract fields
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+    provider.sender(),
+    { value: toNano('0.05') },
+    null, // because there's a `null` message body
+          // `receive()` function in the contract
+  );
+  await provider.waitForDeploy(playground.address);
+}
+```
+
+### Deployments with `Deployable`
```
anton-trunov:
```suggestion
### Do not deploy contracts with `Deployable` trait
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+    receive() { cashback(sender()) } // expects a `null` body
+}
+
+contract Scrooge {
+    // This is when you don't want to send excesses back
+    receive() {} // expects a `null` body
+}
+```
+
+### Gas-expensive badge
```
anton-trunov:
```suggestion
### Pay attention to "Gas-expensive" badge
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+Some functions in the Tact documentation are annotated with a special badge, "Gas-expensive", which marks the functions that use 500 gas units or more. It is placed right under the function name heading and looks like this: <Badge text="Gas-expensive" title="Uses 500 gas units or more" variant="danger" size="medium"/>
+
+If you use one of those functions, consider finding cheaper alternatives.
+
+### Code inlining
```
anton-trunov:
```suggestion
### Inline functions that are rarely called
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+Some [kinds of functions](/book/functions) allow their code to be inlined by adding an `inline` annotation. If the function is used often, this might result in a significant increase in contract code size, but generally, it allows to save gas on redundant function calls.
+
+Furthermore, you might reach for the [experimental `inline` field](/book/config#experimental-inline) in `tact.config.json,` which enables the inlining of all functions that can be inlined.
+
+This advice needs benchmarks to decide its usefulness on a case-by-case basis.
+
+### Off-chain string manipulations
```
anton-trunov:
```suggestion
### Prefer manipulating strings off-chain
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+Strings on-chain are represented as [slices][slice], which are expensive for handling Unicode strings and quite costly even for ASCII ones. Prefer not to manipulate strings on-chain.
+
+### Arithmetic vs. branching
```
anton-trunov:
```suggestion
### Prefer arithmetic to branching operators
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+On the other hand, if you know the layout and boundaries of your data, the generic choice of underlying instructions might be suboptimal in terms of gas usage or code size.
+
+In both cases, you can use assembly functions (or `asm` functions for short) to manually describe the logic by writing a series of [TVM instructions][tvm-instruction]. If you know what you're doing, `asm` functions can offer you the smallest possible gas usage and the most control over [TVM][tvm] execution.
+
+Read more about them on their dedicated page: [Assembly functions](/book/assembly-functions).
+
+## Receiving messages
+
+### String vs. binary receivers
```
anton-trunov:
```suggestion
### Prefer binary receivers to text receivers
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+fun example() {
+    throwUnless(SOMETHING_BAD_1, now() > 1000);
+    throwUnless(SOMETHING_BAD_2, now() > 1000000);
+}
+```
+
+## Sending messages
+
+### Do not use internal functions provided by the `BaseTrait`
```
anton-trunov:
```suggestion
### Prefer `message` and `cashback` to `self.forward`, `self.reply`, and `self.notify`
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+```
+
+## Applied examples of best gas practices
+
+Tact has a growing set of contracts benchmarked against their reference FunC implementations. We fine-tune each Tact contract following the gas-preserving approaches discussed on this page while staying true to the original code and without doing precompilation or excessive ASM usage.
+
+See those examples with recommendations applied:
+
+- [Jetton Wallet contract](https://github.com/tact-lang/tact/blob/main/src/test/benchmarks/contracts/jetton_wallet.tact)
+- [Jetton Minter Discoverable contract](https://github.com/tact-lang/tact/blob/main/src/test/benchmarks/contracts/jetton_minter_discoverable.tact)
```
anton-trunov:
```suggestion
- [Jetton Minter Discoverable contract](https://github.com/tact-lang/jetton/blob/c3be995c67ed0602d59fbbd783ae44ea77d71124/sources/jetton_minter_discoverable.tact)
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+});
+```
+
+## Applied examples of best gas practices
+
+Tact has a growing set of contracts benchmarked against their reference FunC implementations. We fine-tune each Tact contract following the gas-preserving approaches discussed on this page while staying true to the original code and without doing precompilation or excessive ASM usage.
+
+See those examples with recommendations applied:
+
+- [Jetton Wallet contract](https://github.com/tact-lang/tact/blob/main/src/test/benchmarks/contracts/jetton_wallet.tact)
```
anton-trunov:
```suggestion
- [Jetton Wallet contract](https://github.com/tact-lang/jetton/blob/c3be995c67ed0602d59fbbd783ae44ea77d71124/sources/jetton_wallet.tact)
```

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+- [Jetton Minter Discoverable contract](https://github.com/tact-lang/tact/blob/main/src/test/benchmarks/contracts/jetton_minter_discoverable.tact)
+- [Escrow contract](https://github.com/tact-lang/tact/blob/main/src/test/benchmarks/contracts/escrow.tact)
```
anton-trunov:
```suggestion
- [Escrow contract](https://github.com/tact-lang/tact/blob/73fb52b2c8b4e8b2309e0aae4dcc0f8cb35117ea/src/benchmarks/contracts/escrow.tact)
```

docs/src/content/docs/book/gas-best-practices.mdx
novusnota: Well, there's also `send()` and even `deploy()` for some cases of `self.forward`. Plus, none of the `BaseTrait` functions are recommended, so the former heading seems better.

Perhaps, it can be narrowed to "Do not use `self.forward`, `self.reply` and `self.notify`"

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -0,0 +1,275 @@
...
+    to: sender(),
+    value: ton("0.05"),
+});
+```
+
+### Use `throwUnless()` over `require()`
+
+The [`require(){:tact}`](/ref/core-debug#require) function is convenient for stating assumptions in code, especially in debug environments. Granted, currently, it generates exit codes outside the range $0-2^{11}$, making it a bit expensive compared to alternatives.
+
+If you're ready for production and are willing to sacrifice some convenience for gas, use [`throwUnless(){:tact}`](/ref/core-debug#throwunless) function, keep track of your exit codes by declaring them as [constants](/book/constants), and keep exit codes within the inclusive range of $0-2^{11}$.
```
anton-trunov: we need to exclude the 0-255 range here and also link the exit codes page

docs/src/content/docs/book/gas-best-practices.mdx
anton-trunov: No, a negative piece of advice is not useful at all. You always need to suggest alternative

novusnota: > No, a negative piece of advice is not useful at all. You always need to suggest alternative

But the description tells that and more. Why cram stuff into the heading when it can bring attention first, discuss details later?

docs/src/content/docs/book/gas-best-practices.mdx
novusnota: Done

anton-trunov (approved):

PR #2346, opened 2025-03-10
docs: empty contract parameters

### LINKED ISSUE #2324: Document semantics of empty contract parameters
If a contract does not have any state variables, there can be two possibilities:
 
- `contract Foo() { }` does not have the `init()` function
- `contract Foo { }` has implicit `init()` function, meaning the lazy deployment bit gets checked at every incoming message and the contract storage save optimization does not get applied

novusnota:
## Issue

Closes #2324.

### DIFF
--- a/docs/src/content/docs/book/contracts.mdx
+++ b/docs/src/content/docs/book/contracts.mdx
@@ -175,7 +175,7 @@ contract ContractParameters(
 
 State variables declared in this way have **no** default value. Instead, they get their initial data only when the contract is deployed, which is much cheaper compared to deployments with an extra initialization step in the [`init(){:tact}`](#init-function) function.
 
-The use of contract parameters syntax conflicts with using an [`init(){:tact}`](#init-function) function or declaring persistent state variables via the [contract fields syntax](#variables). Whenever possible, prefer using contract parameters unless you need specific on-chain deployment logic that must be run exactly once, i.e. via the [`init(){:tact}`](#init-function) function.
+The use of contract parameters syntax conflicts with using an [`init(){:tact}`](#init-function) function or declaring persistent state variables via the [contract fields syntax](#variables). Whenever possible, prefer using contract parameters unless you need specific on-chain deployment logic that must be run exactly once, i.e. via the `init(){:tact}` function.
 
 ```tact
 contract ParamParamParam(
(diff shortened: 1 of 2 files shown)


docs/src/content/docs/book/contracts.mdx
```diff
@@ -175,7 +175,7 @@ contract ContractParameters(
 
 State variables declared in this way have **no** default value. Instead, they get their initial data only when the contract is deployed, which is much cheaper compared to deployments with an extra initialization step in the [`init(){:tact}`](#init-function) function.
 
-The use of contract parameters syntax conflicts with using an [`init(){:tact}`](#init-function) function or declaring persistent state variables via the [contract fields syntax](#variables). Whenever possible, prefer using contract parameters unless you need specific on-chain deployment logic that must be run exactly once, i.e. via the [`init(){:tact}`](#init-function) function.
+The use of contract parameters syntax conflicts with using an [`init(){:tact}`](#init-function) function or declaring persistent state variables via the [contract fields syntax](#variables). Whenever possible, prefer using contract parameters unless you need specific on-chain deployment logic that must be run exactly once, i.e. via the `init(){:tact}` function.
```
anton-trunov:
```suggestion
The use of contract parameters syntax conflicts with using an `init(){:tact}` function or declaring persistent state variables via the [contract fields syntax](#variables). Whenever possible, prefer using contract parameters unless you need specific on-chain deployment logic that must be run exactly once, i.e. via the `init(){:tact}` function.
```

anton-trunov: you just linked it in the paragraph above

docs/src/content/docs/book/contracts.mdx
```diff
@@ -187,6 +187,21 @@ contract ParamParamParam(
 }
 ```
 
+An empty parameter list is still a parameter list, meaning that the contract won't have an implicit or explicit [`init(){:tact}`](#init-function) function and will enjoy storage write optimizations. Removing the empty list makes a contract use the delayed initialization with an implicit `init(){:tact}` function.
```
anton-trunov:
```suggestion
An empty parameter list is still a parameter list, meaning that the contract won't have an implicit or explicit `init(){:tact}` function and will enjoy storage write optimizations. Removing the empty list makes a contract use the delayed initialization with an implicit `init(){:tact}` function.
```

anton-trunov (changes requested):

PR #2358, opened 2025-03-11
feat(codegen): optimize throw-only fallback receiver

### LINKED ISSUE #1946: Jetton contracts should throw `0xffff` exit code if unrecognized message is received
By default, Tact contracts throw `130` in this case.

Something like this should work:

```tact
receive(_: Slice) { throw(0xffff) }
```

### LINKED ISSUE #2357: Optimize throw-only fallback receiver
To copy the exact behaviour of some FunC standard implementation developers could write something like

```tact
receive(_: Slice) { throw(0xffff); }
```

To address this and enable binary-only optimization we should optimize this to just change our exit code 130 on invalid message to `0xffff`

This shouldn't break any existing tests or interfaces

Kaladin13:
## Issue

Closes #1946
Closes #2357

## Checklist


anton-trunov:
merge conflicts in the benchmark results

novusnota:
Hmm, we use 0xffff because of the Fift selector hack that uses 65535, right? What if, instead of throwing 0xffff, we throw 130 and then modify the selector hack to also use 130?

`@Kaladin13 @Gusarich @anton-trunov`

UPD: Nevermind

Gusarich:
> Hmm, we use 0xffff because of the Fift selector hack that uses 65535, right? What if, instead of throwing 0xffff, we throw 130 and then modify the selector hack to also use 130?
> 
> 
> 
> @Kaladin13 @Gusarich @anton-trunov 

How is this related to selector hack?

novusnota:
> > Hmm, we use 0xffff because of the Fift selector hack that uses 65535, right? What if, instead of throwing 0xffff, we throw 130 and then modify the selector hack to also use 130?
> 
> How is this related to selector hack?

I just thought since they both use the same number it could be related. But I guess not, so please don't mind it :)

anton-trunov:
Superseded by #2396

PR #1258, opened 2024-12-25
refactor: new parser

verytactical:
## Issue

Related to #1128.
Closes #1247.
Closes #1265.

## Checklist

- [x] I have updated CHANGELOG.md
- [x] I have run all the tests locally and no test failure was reported
- [x] I have run the linter, formatter and spellchecker
- [x] I did not do unrelated and/or undiscussed refactorings

PR #2361, opened 2025-03-11
fix(docs/schema): adjust the `parser` config option

### LINKED ISSUE #2359: Invalid sample in parser option tact.config.json
https://docs.tact-lang.org/book/config/#options-parser

novusnota:
## Issue

Closes #2359. A long forgotten follow-up to https://github.com/tact-lang/tact/pull/1258

I've thought about placing the `parser` option into `experimental` options, but it won't make much sense since users of 1.6.0, 1.6.1 and 1.6.2 won't have it there. And we're using the new parser since 1.6.0 anyways, so once the old parser is completely removed we'll simply mark this config option as deprecated.

### DIFF
--- a/docs/src/content/docs/book/config.mdx
+++ b/docs/src/content/docs/book/config.mdx
@@ -256,9 +256,9 @@ If set to `true{:json}`, enables the generation of a [getter](/book/contracts#ge
 
 `"new"{:json}` by default.
 
-If set to `new{:json}`, Tact will compile using the new language parser.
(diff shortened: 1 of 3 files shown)


anton-trunov (dismissed):

verytactical (approved):

verytactical:
This is mostly pointless, because we're going to drop the whole thing soon, but it's nice it was documented at least in some point of time :)

Issue #1692, opened 2025-02-04
Move cryptographic functions from core-math into its own page and add more relevant examples to the Cookbook


anton-trunov:
Some pointers:
- `sha256`, `checkSignature`, `checkDataSignature` should be moved to the new page.
- Let's also mention `.hash()` methods there.
- Improve the examples for signature functions and mention those can be used for mutlisig, for instance.
- Some examples of using `asm` functions with the cryptographic primitives of TVM, especially the ones that Ethereum uses.

P.S.: `core-crypto`

PR #2391, opened 2025-03-14
feat(docs/stdlib): split and reorder Core libraries

### LINKED ISSUE #1830: Reorder and split Core libraries in the Reference
For example, most of the things from core-advanced should be split elsewhere. But not all.

The core-common should be split into contextual, addressing and communication parts.
The core-math should extract the cryptographic parts.

### LINKED ISSUE #1940: Descriptions of `Slice.asString()` and `String.asSlice()` are wrong and very outdated
Nothing happens, both functions are `NOP`s.

Also, we should rewrite their `native` + FunC `asm` implementation in Tact `asm` functions instead once #1807 and/or #1805 would get resolved. Or just remove those now since fixing linked issues should be trivial and in time for 1.6.0.

But the code rewrite is out of the scope of this particular doc issue, since there are many specific issues already, see https://github.com/tact-lang/tact/issues/1761

### LINKED ISSUE #2392: `message` function docs should mention `data`, not `body`
https://docs.tact-lang.org/ref/core-common/#message

> The MessageParameters [struct](https://docs.tact-lang.org/book/structs-and-messages#structs) is similar to the [SendParameters](https://docs.tact-lang.org/book/send) [struct](https://docs.tact-lang.org/book/structs-and-messages#structs), but without the body and code fields.

This should be `data` and `code` fields.

novusnota:
## Issue

Closes #1830.
Closes #1940 (minor fix of two descriptions)
Closes #2392
Towards #1692 (only the move from core-math to core-crypto)

This PR removes core-common and core-advanced, and distributes their contents over old and new pages, such as:
* core-send - message-sending functions, except for those in core-base
* core-crypto - moved sha256 and stuff from core-math
* core-gas - most gas-related things from core-advanced
* core-addresses - most address-related things
* core-contextstate - context and state-related things. I thought about making two disjoint pages, but those are very deeply connected things so it was better to keep them together and organize some subsections.

image


novusnota:
Approves needed :shipit:

UPD: I've found a way to validate links in std/internal without waiting for the weekly link check CI, hence the new commits that introduce new checks.

docs/src/content/docs/ref/core-cells.mdx
```diff
@@ -1653,6 +1653,47 @@ let buzz: Cell = beginCell().storeSlice(s).endCell();
...
+
+```tact
+let s: String = "Keep your Slices close, but your Strings closer.";
+let fizz: String = s;
+let buzz: String = s.asSlice().asString();
+
+fizz == buzz; // true
+```
+
+### Slice.fromBase64
```
anton-trunov, line 1676: Should we mark this as gas expensive?

docs/src/content/docs/ref/core-contextstate.mdx
```diff
@@ -0,0 +1,373 @@
...
+```
+
+:::note[Useful links:]
+
+  [Gas in TON Docs][gas]\
+  [Other fees and gas-related functions in the Core library](/ref/core-gas)
+
+:::
+
+### nativeReserve
```
anton-trunov, line 255: this arguably should be moved to the `send`-page

docs/src/content/docs/ref/stdlib-ownable.mdx
```diff
@@ -72,7 +72,7 @@ contract ExampleContract with Ownable {
 
 ### OwnableTransferable
 
-`OwnableTransferable{:tact}` is an extension of [`Ownable{:tact}`](#ownable) that allows transferring ownership of a contract to another address. It provides a secure handler for the [Message](/book/structs-and-messages#messages) [`ChangeOwner{:tact}`](#changeowner) that can be called by the owner to transfer ownership.
+`OwnableTransferable{:tact}` is an extension of [`Ownable{:tact}`](#ownable) that allows transferring ownership of a contract to another address. It provides a secure handler for the [message struct](/book/structs-and-messages#messages) [`ChangeOwner{:tact}`](#changeowner). which can only be processed when sent by the current owner.
```
anton-trunov: extra `.`

docs/src/content/docs/ref/core-contextstate.mdx
novusnota: Hmm, not sure, since there are other actions like SETCODE and library change, plus the `myBalance()` and `gasConsumed()` are on the same page.

docs/src/content/docs/ref/core-cells.mdx
novusnota: Yes, we should

docs/src/content/docs/ref/core-contextstate.mdx
novusnota: I'll add more cross-links for visibility though in a moment

novusnota: Done!

anton-trunov (approved):

PR #2395, opened 2025-03-17
feat(stdlib): `Slice.loadBasechainAddress()` and `Slice.skipBasechainAddress()`

### LINKED ISSUE #2044: Document `BasechainAddress` struct and related functions

### LINKED ISSUE #2412: Add utility functions for dealing with BasechainAddresses for a Slice
The `BasechainAddress` was introduced in https://github.com/tact-lang/tact/pull/2035, which also added the `Builder.storeBasechainAddress()` extension function, but did not add the respective load/skip functions to the stdlib.

novusnota:
## Issue

Closes #2412


src/stdlib/stdlib/std/internal/address.tact
```diff
@@ -323,6 +326,76 @@ extends fun storeBasechainAddress(self: Builder, address: BasechainAddress): Bui
...
+/// * https://docs.tact-lang.org/book/exit-codes
+///
+/// [exit code 8]: https://docs.tact-lang.org/book/exit-codes#8
+/// [exit code 9]: https://docs.tact-lang.org/book/exit-codes#9
+///
+extends mutates fun loadBasechainAddress(self: Slice): BasechainAddress {
+    if (self.loadUint(2) == 0) {
+        return BasechainAddress{ hash: null };
+    } else {
+        // Skip the remaining prefix
```
anton-trunov: what if the prefix is not zero?

anton-trunov (changes requested):
Let's remove all the new functions and keep just the docs changes

anton-trunov:
I don't understand the purpose of this PR now, if we remove the new functions. If you want to fix the links, please open the new one

novusnota:
> I don't understand the purpose of this PR now, if we remove the new functions. If you want to fix the links, please open the new one

(which is why i don't like posting incomplete drafts)

1. this pr is now with functions
2. and with important fixes
3. but not all, since those were meant to be done for basechain address too, but now its docs are in another pr #2411

novusnota:
Now this needs #2411 because of the reference of BasechainAddress, which was supposed to be in core-addresses :)

anton-trunov:
> (which is why i don't like posting incomplete drafts)

there is a difference between drafts and mixing new features and fixes together

docs/src/content/docs/ref/core-cells.mdx
```diff
@@ -1194,6 +1194,59 @@ s1.skipAddress();
...
+extends mutates fun loadBasechainAddress(self: Slice): BasechainAddress;
+```
+
+Extension mutation function for the [`Slice{:tact}`][slice] type.
+
+Loads and returns a [`BasechainAddress{:tact}`][base-addr] [struct][struct] from the [`Slice{:tact}`][slice].
+
+Attempts to load such [`BasechainAddress{:tact}`][base-addr] struct when [`Slice{:tact}`][slice] doesn't contain it throw an exception with [exit code 8](/book/exit-codes#8): `Cell overflow`.
+
+Attempts to load more data than [`Slice{:tact}`][slice] contains throw an exception with [exit code 9](/book/exit-codes#9): `Cell underflow`.
```
anton-trunov, line 1211: What happens when a slice that is not a valid basechain addess is passed?

novusnota: Exit code 9.

Actually, I may have found a bunch of long-standing typos — the exit code 8 doesn't seem right for the .load functions.

Double-checking those now.

UPD: They're incorrect

anton-trunov: This should be documented

src/stdlib/stdlib/std/internal/address.tact
```diff
@@ -323,6 +326,81 @@ extends fun storeBasechainAddress(self: Builder, address: BasechainAddress): Bui
...
+/// [exit code 9]: https://docs.tact-lang.org/book/exit-codes#9
+///
+extends mutates fun loadBasechainAddress(self: Slice): BasechainAddress {
+    // addr_none
+    if (self.loadUint(2) == 0) {
+        return BasechainAddress{ hash: null };
+    }
+
+    // addr_std, skip the remaining prefix bits
+    self.skipBits(9);
```
anton-trunov, line 360: what happens when we have workchains other than basechain?

anton-trunov (changes requested):

novusnota:
Agree with moving it to 1.6.5, we can be more specific when attempting to load the BasechainAddress from a Slice. Like, to use exit codes 138 (not a basechain address), for example.

Plus, the exit code 8 doesn't seem achievable with the .load instructions, so that has to be fixed (in progress)

novusnota:
Converting to a draft until learnXinY is ready for review. Then, I'll return to this.

Here, I'll:
1. improve the contents of two new load/skip functions and make them use the basechain-related exit codes
2. make all .store functions mention the exit code 8 and all .load/.skip functions not mention it

novusnota:
This PR has already rotted a bit. I was trying to fix multiple problems at once and am now splitting the main one into its own issue. 

UPD: #2510

PR #2402, opened 2025-03-17
refactor: use own types for abi

### LINKED ISSUE #1429: Don't use `ContractABI` from `@ton/core`
`ContractABI` from `@ton/core` is incorrect and doesn't have some of the required fields (`interfaces: string[]`, see `src/generator/createABI.ts`).

verytactical:
## Issue

Closes #1429.


verytactical:
We decided to find an alternative approach to ABI that doesn't use these types at all.

PR #2403, opened 2025-03-17
chore: faster tests

verytactical:


novusnota:
Try commenting the `build:fast` for a while, to see if the regular `build` will fail

PR #2410, opened 2025-03-18
refactor(codegen): apply parameters rearrangement only for methods with single parameter

### LINKED ISSUE #2390: Confusing semantics of asm method calls when arguments are mutating
This fails with exit code 5:

```tact
fun example() {
    let s = beginCell()
        .storeBool(true)
        .storeInt(0, 257)
        .asSlice();
    
    let c = beginCell()
        .storeTwo(s.loadBool(), s.loadInt(257)) // the rearrangement optimization does not take the
                                                // immediate mutating operations into account
        .endCell();
    dump(c); // just to keep the code above
}
```

But this does not because the mutating operations are done before the optimization of the rearrangement:

```tact
fun example() {
    let s = beginCell()
        .storeBool(true)
        .storeInt(0, 257)
        .asSlice();

    let a = s.loadBool();
    let b = s.loadInt(257);

    let c = beginCell()
        .storeTwo(a, b)
        .endCell();
    dump(c); // just to keep the code above
}
```

In both cases, the `.storeTwo()` function is defined the same:

```tact
// first stores Bool, then stores Int
asm(val2 val1 self) extends fun storeTwo(self: Builder, val1: Bool, val2: Int): Builder {
    DUMPSTK
    1 STI   // val1 into self
    256 STI // val2 into self
}
```

The intended behavior [was documented](https://docs.tact-lang.org/book/assembly-functions/#arrangements) as follows:

> Considering arrangements, the evaluation flow of the assembly function can be thought of in these steps:
>
> 1. The function takes arguments in the order specified by the parameters.
> 2. If an argument arrangement is present, arguments are reordered before being pushed onto the stack.
> 3. The function body, consisting of TVM instructions and primitives, is executed.
> ...

Related to https://github.com/tact-lang/tact/pull/1702 and https://github.com/tact-lang/tact/issues/2388

i582:
## Issue

Closes #2390.

## Checklist

### DIFF
--- a/dev-docs/CHANGELOG.md
+++ b/dev-docs/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Language features
+
+- [fix] Apply parameters rearrangement only for ASM methods with single parameter: PR [#2410](https://github.com/tact-lang/tact/pull/2410)
+
 ### Internal infrastructure
 
 - Removed `postinstall` from `package.json` to not run scripts with dev dependencies on the user side: PR [#2382](https://github.com/tact-lang/tact/pull/2382)
--- a/dev-docs/CHANGELOG.md
+++ b/dev-docs/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Language features
 
-- [fix] Apply parameters rearrangement only for ASM methods with single parameter: PR [#2410](https://github.com/tact-lang/tact/pull/2410)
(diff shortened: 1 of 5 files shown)


Kaladin13 (dismissed):

novusnota (dismissed):

dev-docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Language features
+
+- [fix] Apply parameters rearrangement only for ASM methods with single parameter: PR [#2410](https://github.com/tact-lang/tact/pull/2410)
```
```suggestion
- Apply parameters rearrangement only for ASM methods with a single parameter to avoid confusion: PR [#2410](https://github.com/tact-lang/tact/pull/2410)
```

anton-trunov:
Let's do it to avoid confusion for our users, but technically this is not a fix since the order of function arguments evaluation is not specified. So marking this as `refactor`.

PR #2427, opened 2025-03-18
feat(codegen): optimize Context().sender to use sender()

### LINKED ISSUE #1557: Interpret `Context().sender` as `sender()`
Doing it directly may break backward compatibility.

However we may interpret Context().sender as sender() - temporary solution. Which will be better in most cases

Gusarich:
This optimization replaces Context().sender field access with a direct call to sender() function for better gas efficiency, as recommended in the documentation.

## Issue

Closes #1557.

### DIFF
--- a/src/generator/writers/writeExpression.ts
+++ b/src/generator/writers/writeExpression.ts
@@ -438,6 +438,19 @@ export function writeExpression(
     //
 
     if (f.kind === "field_access") {
+        // Optimize Context().sender to sender()
+        // This is a special case to improve gas efficiency
+        if (
+            f.aggregate.kind === "static_call" &&
+            f.aggregate.function.text === "context" &&
+            f.aggregate.args.length === 0 &&
+            f.field.text === "sender"
+        ) {
+            // Use sender() directly instead of context().sender
+            wCtx.used("__tact_context_get_sender");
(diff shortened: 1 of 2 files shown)


novusnota (dismissed):

src/generator/writers/writeExpression.ts
anton-trunov, line 1: In principle, this should be solved with a data-flow analysis. It's fine to have some special cases for now, but we need to demonstrate that this improves gas consumption. For instance, our benchmark contracts routinely use something like this:

```tact
let ctx: Context = context();
throwUnless(401, ctx.value == self.dealAmount);

self.isFunded = true;
self.buyerAddress = ctx.sender;
```

anton-trunov (changes requested):

src/generator/writers/writeExpression.ts
Gusarich: in this specific case this optimization shouldn't apply - gas consumption becomes worse if you try to replace any `(x: Context).sender` expression with `sender()` when there other field accesses too, because all `Context` values are on stack already and just taking it from there is cheaper in most cases

it makes sense to implement data-flow analysis and catch all cases when you only ever access the `sender` field from `Context`, but it'll require much more changes

anton-trunov (approved):

PR #2432, opened 2025-03-20
fix(codegen): ternary operation with struct and null

### LINKED ISSUE #2430: Fix tuple parameters in case of ternary expression
```tact
struct Info {
    owner: Address;
    minter: Address;
}

contract Example(
    owner: Address,
    minter: Address,
) {
    receive() {
        let foo = true ? Info{
                    owner: self.owner,
                    minter: self.minter } : null;
    }
}
```

Fails with 
```
💼 Compiling project Jetton ...
   > Example: tact compiler
   > Example: func compiler
FunC compilation error: /home/kaladin/IdeaProjects/work/jetton/src/output/Jetton_Example.fc:91:97: error: the two variants in a conditional expression have different types (slice, slice) and ??51 : cannot unify type ??51 with (slice, slice): cannot unify types of widths 1 and 2
          tuple $foo = (true ? $Info$_constructor_owner_minter($self'owner, $self'minter) : null());

error Command failed with exit code 30.
```

Gusarich:
## Issue

Closes #2430.

## Checklist

### DIFF
--- a/src/generator/writers/writeExpression.ts
+++ b/src/generator/writers/writeExpression.ts
@@ -14,6 +14,7 @@ import {
 } from "../../types/resolveDescriptors";
 import type { FieldDescription, TypeDescription } from "../../types/types";
 import { printTypeRef } from "../../types/types";
+import type { TypeRef } from "../../types/types";
 import type { WriterContext } from "../Writer";
 import { resolveFuncTypeUnpack } from "./resolveFuncTypeUnpack";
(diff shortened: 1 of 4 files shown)


src/generator/writers/writeExpression.ts
anton-trunov, line 1: Let's refactor the common code into a helper function

Gusarich: @anton-trunov done

PR #2480, opened 2025-03-25
fix: range intersection typo

### LINKED ISSUE #2447: Internal compiler error on incomplete return statement
The following minimal Tact snippet triggers an internal compiler error:

```tact
fun f(): Int {
    return 
```

**Observed behavior:**

Compiling the snippet above results in the following internal compiler error:

```
Tact compilation failed
Error: [INTERNAL COMPILER ERROR]: Interval [26, 26[ is empty or out of source bounds (26)
```

---

verytactical:
## Issue

Closes #2447.


anton-trunov (changes requested):
Needs a test. No reason not to use the test case provided by the LLM Agent

PR #2482, opened 2025-03-25
fix(codegen): save contract state on early return in receivers

### LINKED ISSUE #2461: Early `return` statements in receivers must invoke contract state saves
The following example should increment `x` after calling the empty receiver, but, in fact, the contract state never changes:

```tact
contract Foo(x: Int) {
  receive() {
    if (true) {
      x += 1;
      return;
    }
  }
}
```

anton-trunov:
## Issue

Closes #2461.

## Checklist


src/generator/writers/writeRouter.ts
```diff
@@ -673,18 +669,43 @@ function writeReceiverBody(
...
+    );
     wCtx.append("return ();");
 }
 
+function storeContractVariablesConditionally(
+    rcvEffects: ReadonlySet<Effect>,
+    contract: TypeDescription,
+    wCtx: WriterContext,
+): string {
+    return enabledAlwaysSaveContractData(wCtx.ctx) ||
```
Kaladin13, line 689: I understand that this is hack for now, but maybe let's break this 3-conditional ternary operator into few variables or add comments? 

Really hard to grasp what's happening here, especially with empty string as stub

Kaladin13 (dismissed):
Everything good, comment on ternary operator

i582 (approved):

src/generator/writers/writeRouter.ts
anton-trunov: added a comment

PR #2484, opened 2025-03-26
fix: show error for assembly function with `get` attribute

### LINKED ISSUE #2446: Compiler incorrectly allows `get` attribute on assembly functions
According to the documentation (e.g., in docs/book/assembly-functions.md), assembly functions cannot be declared with the 'get' attribute. However, the test snippet below using 'asm get fun illegalGetter(): Int { DEPTH }' compiled successfully, indicating that the compiler accepts assembly functions as getters. This constitutes a clear and confirmed discrepancy between the documented language restrictions and the actual compiler behavior.

```tact
asm get fun illegalGetter(): Int {
    DEPTH
}
```

At the same time, the following snippet results in compilation error, as expected:
```tact
get fun illegalGetter2(): Int {
    return 123;
}
```

---

i582:
## Issue

Closes #2446.

## Checklist


dev-docs/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Optimized `Context().sender` to use `sender()` function for better gas efficiency: PR [#2427](https://github.com/tact-lang/tact/pull/2427)
 - [fix] Ternary operator with struct and null: PR [#2432](https://github.com/tact-lang/tact/pull/2432)
+- [fix] Show error for assembly function with get attribute: PR [#2484](https://github.com/tact-lang/tact/pull/2484)
```
```suggestion
- [fix] Show error for assembly functions with the `get` attribute: PR [#2484](https://github.com/tact-lang/tact/pull/2484)
```

dev-docs/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Optimized `Context().sender` to use `sender()` function for better gas efficiency: PR [#2427](https://github.com/tact-lang/tact/pull/2427)
 - [fix] Ternary operator with struct and null: PR [#2432](https://github.com/tact-lang/tact/pull/2432)
+- [fix] Show error for assembly functions with the `get` attribute: PR [#2484](https://github.com/tact-lang/tact/pull/2484)
```
```suggestion
- [fix] Show an error message for assembly functions with the `get` attribute: PR [#2484](https://github.com/tact-lang/tact/pull/2484)
```

src/types/resolveDescriptors.ts
```diff
@@ -862,6 +862,13 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
             }
         }
 
+        if (a.kind === "asm_function_def" && isGetter) {
+            throwCompilationError(
+                "Assembly functions cannot be getters",
```
anton-trunov: This is a confusing error message: any module-level function cannot be a getter, since it's not a contract function

anton-trunov (changes requested):

src/types/resolveDescriptors.ts
i582: Then let's just reuse the `Getters must be defined within a contract` error message that we give for getters outside contracts

anton-trunov: And also, let's add a test with an `asm` getter inside a contract

i582: Added a test, it seems the current parser can't parse the asm getter correctly. AFAIK, we already have an issue about this

anton-trunov: Could you link that issue here, please?

anton-trunov: Maybe this one #1619?

i582: > Maybe this one #1619?

yep

PR #2497, opened 2025-03-26
fix(docs): description of the `contractAddressExt()` function

### LINKED ISSUE #2452: Documentation incorrectly mentions `contractAddressExt`'s chain ID param is checked at compile-time
According to Tact documentation, the function `contractAddressExt` should **reject invalid chain IDs** (any values other than `-1` or `0`) at compile time.

However, compiling the following minimal snippet with an invalid chain ID (`1`) succeeds without any errors:

```tact
contract WrongChain {
    init() {
        let initPkg = initOf WrongChain();
        // Using invalid chain ID: 1 (only -1 and 0 are valid)
        let wrongAddr = contractAddressExt(1, initPkg.code, initPkg.data);
    }
    
    receive() {}
}
```

### Expected Behavior:
Compilation should fail explicitly, disallowing chain IDs other than `-1` or `0`.

### Actual Behavior:
The compiler incorrectly accepts invalid chain IDs, contradicting documented specifications.

---

novusnota:
## Issue

Closes #2452, plus did very minor formatting adjustments for the `@stdlib/deploy`


docs/src/content/docs/ref/core-addresses.mdx
```diff
@@ -77,10 +77,6 @@ fun contractAddressExt(chain: Int, code: Cell, data: Cell): Address;
 
 Computes the smart contract's [`Address{:tact}`][p] in the given `chain` ID using the contract's `code` and its initial state `data`. Use the [`initOf{:tact}`][initpkg] expression to obtain the initial `code` and initial `data` of a given contract.
 
-This function tries to resolve constant values at [compile-time](/ref/core-comptime) whenever possible.
-
-Attempts to specify an invalid `chain` ID (not -1 or 0) that can be detected at [compile-time](/ref/core-comptime) will result in a compilation error.
-
```
anton-trunov: but what happens at runtime?

anton-trunov (changes requested):
let's explain what happens at runtime if the chain id is incorrect

docs/src/content/docs/ref/core-addresses.mdx
novusnota: Nothing much, contract's address is created in the specified workchain even if it's not -1 or 0. Same with newAddress, but that one throws at compile-time because of corresponding logic in `interpreter.ts`

novusnota: Explicitly stated so.

novusnota:
> let's explain what happens at runtime if the chain id is incorrect

1. not incorrect, but uncommon
2. explicitly stated that everything works still
3. and clarified what "stopped" contract means, because this notion is made by us and such contract state doesn't exist (there are those: active, frozen, uninit)

Furthermore, we can remove that comptime check from the newAddress in the interpreter.ts. But I've thought not to :)

PR #2513, opened 2025-03-27
refactor(stdlib): explicitly assign opcodes to all message structs

### LINKED ISSUE #2508: Explicitly assign opcodes to all message structs in the standard library
Instead of `message Name` there should be `message(number) Name`, where `number` should be taken as is from the compilation report / similar sources to prevent incompatible changes.

novusnota:
## Issue

Closes #2508.


src/stdlib/stdlib/libs/deploy.tact
```diff
@@ -2,7 +2,7 @@
 ///
 /// See: https://docs.tact-lang.org/ref/stdlib-deploy#deploy
 ///
-message Deploy {
+message(2490013878) Deploy {
```
anton-trunov: let's maybe use hexadecimal numbers here; it can be mapped to 32-bit opcodes easily

anton-trunov (approved):

PR #2521, opened 2025-03-28
fix(docs): descriptions of `getOriginalFwdFee()` and `Context.readForwardFee()` functions

### LINKED ISSUE #1869: readForwardFee misdescription
https://docs.tact-lang.org/ref/core-advanced/#contextreadforwardfee

It says that it reads forward fee from incoming message. But more accurate description could be 
"It reads forwardFee from the incoming message and estimates forward + action fee" or smth like that.
(Because it actually multiplies forwardFee on 3/2 approx.)

novusnota:
## Issue

Closes #1869.

Also corrected links for skip functions (all the other links elsewhere are good) and added a note about address-related functions of the core-cells page on top of the core-address page.


docs/src/content/docs/ref/core-gas.mdx
```diff
@@ -184,17 +184,26 @@ let lumpPrice = fee - feeNoLump;
...
+The result is computed using the [`first_frac`][param-24-25] value, which is obtained from [config param 24 for the masterchain and config param 25 for the basechain][param-24-25] of TON Blockchain. Due to the current value of `first_frac` for all workchains, this function performs a cheaper equivalent calculation of `fwdFee * 3 / 2`. This ratio might change, so it is better not to hardcode it and use this function instead.
+
+This function can be useful when the outgoing message depends heavily on the structure of the incoming message, so much so that you cannot fully predict the fee using [`getForwardFee(){:tact}`](#getforwardfee) alone. Even if you could, calculating the exact fee with [nanoToncoin][nano]-level precision can be very expensive, so the approximation given by this function is often good enough.
```
Kaladin13:
```suggestion
This function can be useful when the outgoing message depends heavily on the structure of the incoming message, so you can try to approximate the forward fee for your outgoing message based on the fee that previous sender paid. Calculating the exact fee with [nanoToncoin][nano]-level precision can be very expensive, so the approximation given by this function is often good enough.
```

Shvandre: > So much so that you cannot fully predict ... Even if you could ...


You can actually predict any forward fee, It is just very expensive, due to the need of opwning and traversing all cells. So originalFwdFee is used, when the outcoming message is less or equal to incoming one, just to cheaply estimate the fee.

(Maybe let's add a warning, that estimating fee like this, when outcoming message may become bigger than incoming is a potential issue/vulnerability)

docs/src/content/docs/ref/core-gas.mdx
```diff
@@ -184,17 +184,26 @@ let lumpPrice = fee - feeNoLump;
 fun getOriginalFwdFee(fwdFee: Int, isMasterchain: Bool): Int;
 ```
 
-Calculates and returns the so-called _original_ [forward fee][forward-fee] in [nanoToncoin][nano] [`Int{:tact}`][int] for an outgoing message based on the `fwdFee` obtained from the incoming message. If both the source and the destination addresses are in the [basechain][basechain], specify `isMasterchain` as `false{:tact}`. Otherwise, specify `true{:tact}`.
+Calculates and returns the so-called _original_ [forward fee][forward-fee] in [nanoToncoin][nano] [`Int{:tact}`][int] for an outgoing message based on the given `fwdFee`, which can be obtained from the incoming message. If both the source and the destination addresses are in the [basechain][basechain], specify `isMasterchain` as `false{:tact}`. Otherwise, specify `true{:tact}`.
```
Kaladin13: > Calculates and returns the so-called _original_ [forward fee][forward-fee] in [nanoToncoin][nano] [`Int{:tact}`][int] for an outgoing message

I think we should make it clear that at its core this function doesn't have anything to do with outgoing message, it's just happens we can use it this way. In my opinion current phrasing could be misleading

src/stdlib/stdlib/std/internal/context.tact
```diff
@@ -113,7 +113,9 @@ asm extends fun readForwardFee(self: Context): Int {
     6 1 BLKDROP2 // drop the loaded values as well as the first 3 fields of Context
 
     ZERO              // not masterchain
-    GETORIGINALFWDFEE // (fwd_fee * 2^16) / first_frac, where
+    GETORIGINALFWDFEE // floor(fwd_fee * 2^16 / (2^16 - first_frac)), where
```
Shvandre, line 116: this (2^16 - first_frac) doesn't look correct. I am currently looking for needed line in transaction.cpp

Shvandre: Found.

https://github.com/ton-blockchain/ton/blob/7151ff26279fef6dcfa1f47fc0c5b63677ae2458/crypto/block/transaction.cpp#L2614

https://github.com/ton-blockchain/ton/blob/7151ff26279fef6dcfa1f47fc0c5b63677ae2458/crypto/block/transaction.cpp#L2136

Shvandre: I was mistaken :(

Everything is correct
https://github.com/ton-blockchain/ton/blob/303e92b7750dc443ae6c282fb478d2114079d216/crypto/vm/tonops.cpp#L333

Shvandre (approved):

PR #2527, opened 2025-03-28
feat(stdlib): exit code constants

### LINKED ISSUE #2092: Use constants for all Tact-specific exit codes
For example, in `@stdlib/ownable` the throwUnless should use the constant of 132, not the raw number.

https://docs.tact-lang.org/book/exit-codes/#132

skywardboundd:
Closes #2092


src/stdlib/stdlib/std/internal/exit-codes.tact
```diff
@@ -0,0 +1,34 @@
+/// [Null reference exception] — Thrown when a null reference is encountered during the compute phase.
```
i582: No need to prefix documentation with constant name

novusnota: I.e., remove the `[ ... ] —` here and elsewhere and that'd be fine :)

src/stdlib/stdlib/std/internal/exit-codes.tact
```diff
@@ -0,0 +1,34 @@
...
+
+/// [Invalid standard address] — Thrown when an address does not conform to the expected standard format.
+const InvalidStandardAddress: Int = 136;
+
+/// [Masterchain not enabled] — Thrown when masterchain support is required but not enabled for this contract.
+/// Removed since Tact 1.6.
+const MasterchainNotEnabled: Int = 137;
+
+/// [Not a basechain address] — Thrown when an address does not belong to the basechain.
+const NotABasechainAddress: Int = 138;
```
novusnota: Available since Tact 1.6.3: https://docs.tact-lang.org/book/exit-codes/#138

skywardboundd: was not written in https://docs.tact-lang.org/book/exit-codes/#table ( may be open issue )

skywardboundd:
Closes #2092

src/stdlib/stdlib/std/internal/exit-codes.tact
novusnota: > was not written in https://docs.tact-lang.org/book/exit-codes/#table ( may be open issue )

Oh, then we can add that right now, because it's mentioned at the bottom already. One moment, I'll fix that and exit-codes.tact

novusnota:
We might want to add all Tact-specific constants even if they're unused in our stdlib yet, simply because all of them leak into user contracts. Also, they should be mentioned on the exit codes page then, in respective sections.

Additionally, I suggest adding some sort of a prefix to them, akin to `SendX` and `ReserveY` modes. For example, we can have `TactExitCodeAccessDenied` instead of simple `AccessDenied`. Or `TECAccessDenied` for short :)

@anton-trunov

anton-trunov:
Sure, let's add all of them, prefixed with `TactExitCode`, and use them in the exit code tests: https://github.com/tact-lang/tact/tree/main/src/test/exit-codes.

i582 (dismissed):
LGTM

novusnota (commented):
"I concur", but we'll need to resolve the `change-case` packaging issues and the fact that `camel-case` was deprecated on npm.

There are some ESM vs. CommonJS conflicts there, and we ONLY import `change-case` in two places: src/types/resolveSignatures.ts and src/bindings/writeTypescript.ts.

novusnota (changes requested):
Please add yourself to the list of contributors and create a separate CHANGELOG entry since the stdlib contains new constants that will be accessible in the user's code.

novusnota (approved):

PR #2551, opened 2025-03-30
draft: sha256 without ton/core

### LINKED ISSUE #1434: Move `@ton/crypto` to devDependencies

verytactical:
## Issue

Closes #1434.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,12 +81,12 @@
   "dependencies": {
     "@tact-lang/opcode": "^0.3.0",
     "@ton/core": "0.60.1",
-    "@ton/crypto": "^3.2.0",
     "@tonstudio/parser-runtime": "^0.0.1",
     "blockstore-core": "1.0.5",
     "change-case": "^4.1.2",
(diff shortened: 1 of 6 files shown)


verytactical:
Need more polish until this can be merged
- typed arrays instead of buffers
- tests

PR #2591, opened 2025-04-01
chore: make `yarn to-absolute` communicate what has happened

### LINKED ISSUE #2590: Make `yarn to-absolute` communicate what has happened
Now it is unclear what happens when this command fails. We need to add log when it fails

Image

Kaladin13:
## Issue

Closes #2590

### DIFF
--- a/src/to-absolute.build.ts
+++ b/src/to-absolute.build.ts
@@ -36,6 +36,7 @@ const main = async () => {
         }
     }
     if (hadChanges) {
+        console.error("Fixed relative imports, stage changes");
         process.exit(30);
     }
 };


src/to-absolute.build.ts
```diff
@@ -36,6 +36,7 @@ const main = async () => {
         }
     }
     if (hadChanges) {
+        console.error("Fixed relative imports, stage changes");
```
novusnota, line 39:
```suggestion
        console.error("Fixed relative imports — please, stage changes");
```

just a nit

novusnota: Actually, there's often nothing to stage — those fixes are applied to some files in .gitignore: the generated tests and such. Therefore, instead of this `console.error` we might need to look at what `git` says and only then exit with code 30

cc @verytactical

verytactical:
I would rather report which files were changed, not just the fact they did.

I know for a fact it's formatting some files it shouldn't, such as wrappers in `/output/` folders, but it's mostly invisible at the moment, and it's not obvious at the moment how to fix.

anton-trunov (changes requested):
> I would rather report which files were changed, not just the fact they did.

let's address this comment

anton-trunov:
ping

anton-trunov:
@Kaladin13 lets finish with this PR asap

Kaladin13:
Actually we have communicated with @verytactical that he will finish that PR, I have reassigned this issue to him

PR #2612, opened 2025-04-03
fix(docs): jetton minter link and hello world example

### LINKED ISSUE #2611: learnTactInY: make the hello world world example deployable and fix the link to jetton minter

novusnota:
## Issue

Closes #2611.


i582 (dismissed):

anton-trunov (changes requested):
let's do some fixes along the way: please remove the string receivers and add binary messages

novusnota:
> let's do some fixes along the way: please remove the string receivers and add binary messages

The thing is, I deliberately kept those in and added a note block right after, saying that those aren't meant to be used in production and stuff.

image

Otherwise empty message structs would be easily confused with empty receiver, because the notion of opcodes isn't introduced yet

anton-trunov:
> The thing is, I deliberately kept those in and added a note block right after, saying that those aren't meant to be used in production and stuff.

let's teach someone a thing and then tell them not to use it

makes sense? to me, it does not

anton-trunov:
we should not promote manipulating strings on-chain at all

docs/src/content/docs/book/learn-tact-in-y-minutes.mdx
```diff
@@ -26,37 +26,42 @@ You can try Tact without installing anything locally using the [Web IDE](https:/
...
-```
-
-:::note
-
-  The text receivers shown above are expensive because they need to hash their contents to differentiate message bodies, and that is a costly operation.
 
-  Later, see examples of cheaper and more controlled binary receivers that expect message structs: [Functions](#functions).
-
-:::
+// Empty message structs with autogenerated 32-bit integer prefix.
```
anton-trunov: let's not use auto-generated opcodes

docs/src/content/docs/book/learn-tact-in-y-minutes.mdx
```diff
@@ -26,37 +26,42 @@ You can try Tact without installing anything locally using the [Web IDE](https:/
 ```tact
 // Defining a contract
 contract HelloWorld {
-    // Listens to incoming "ping" messages
-    receive("ping") {
+    // Listens to incoming Ping messages
+    receive(_: Ping) {
         // Sends a reply: "pong"
         reply("pong");
```
anton-trunov: let's not use strings here too

anton-trunov (approved):

PR #2655, opened 2025-04-05
refactor(codegen): zero comparison optimization

### LINKED ISSUE #2644: Optimize `x != 0`
In `src/benchmarks/contracts/wallet-v4.tact`

```
while (payload.refs() > 0)
```

is not optimized into FunC code

```
while (cs.slice_refs())
```

because `Int` is not considered a `Bool`.

And even if we change the condition to `!= 0`, it won't be optimized.

By manually modifying the code to

```
asm fun cast(x: Int): Bool {
    //
}

while (cast(payload.refs()))
```

we can see an improvement in gas usage.

Add optimization for `x != 0` for `x: Int` that does not generate the condition in FunC.

anton-trunov:
## Issue

Closes #2644.


i582 (dismissed):
Nice optimization!

i582:
Maybe we can add more tests :)

i582 (approved):
Awesome tests!

PR #2658, opened 2025-04-05
feat(ast-printer): support contract parameters

### LINKED ISSUE #2657: AST Printer Drops Contract Parameters
The AST printer currently omits contract parameters entirely when printing the syntax tree, even though they are valid and present in the source.

```tact
contract Test(x: Int) {...}
```

When passed through the AST printer, the output shows the definition as:
```tact
contract Test {...}
```

### What’s Wrong

- The parameter `x: Int` is completely missing in the printed output.
- This breaks round-trip expectations and makes the AST printer unreliable for tools relying on accurate structural output.

### Expected Behavior

The output should preserve contract parameters exactly as they appear in the source:

```tact
contract Test(x: Int) {...}
```

### Conclusion

This is a bug in the **AST printer**, not the compiler itself. It must be updated to correctly handle and print contract parameters as part of the contract definition.

---

Gusarich:
## Issue

Closes #2657.

## Checklist

### DIFF
--- a/src/ast/ast-printer.ts
+++ b/src/ast/ast-printer.ts
@@ -456,16 +456,33 @@ export const ppAstContract: Printer<Ast.Contract> =
             .map(({ name: { value } }) => `@interface("${value}") `)
             .join("");
         const traitsCode = traits.map((trait) => trait.text).join(", ");
-        const paramsCode = params
-            ? `(${params
-                  .map((param) => {
-                      const asAlias = param.as
-                          ? ` as ${ppAstId(param.as)}`
-                          : "";
-                      return `${ppAstOptionalId(param.name)}: ${ppAstType(param.type)}${asAlias}`;
-                  })
-                  .join(", ")})`
-            : "";
+
+        let paramsCode = "";
+        if (params) {
+            if (params.length <= 1) {
+                // Single parameter or empty params stays on the same line
+                paramsCode = `(${params
+                    .map((param) => {
+                        const asAlias = param.as
+                            ? ` as ${ppAstId(param.as)}`
+                            : "";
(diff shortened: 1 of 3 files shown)


src/ast/ast-printer.ts
```diff
@@ -450,15 +450,42 @@ export const ppAstStruct: Printer<Ast.StructDecl> =
...
     (c) => {
         const attrsCode = attributes
             .map(({ name: { value } }) => `@interface("${value}") `)
             .join("");
         const traitsCode = traits.map((trait) => trait.text).join(", ");
+
+        let paramsCode = "";
+        if (params) {
+            if (params.length <= 1) {
+                // Single parameter or empty params stays on the same line
```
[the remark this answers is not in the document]
Gusarich: `()` should be kept as `()`, it's intended behaviour

src/ast/ast-printer.ts
```diff
@@ -450,15 +450,42 @@ export const ppAstStruct: Printer<Ast.StructDecl> =
...
 export const ppAstContract: Printer<Ast.Contract> =
-    ({ name, traits, declarations, attributes }) =>
+    ({ name, traits, declarations, attributes, params }) =>
     (c) => {
         const attrsCode = attributes
             .map(({ name: { value } }) => `@interface("${value}") `)
             .join("");
         const traitsCode = traits.map((trait) => trait.text).join(", ");
+
+        let paramsCode = "";
```
verytactical: Let's not use mutable variables where there is no need.

You can always just `return` that paramsCode from a separate function.

verytactical (changes requested):

src/ast/ast-printer.ts
Gusarich: Have a look now please!

i582 (approved):

PR #2681, opened 2025-04-08
fix: add check for `ton()` with empty or blank string

### LINKED ISSUE #2538: `ton("")` Successfully Compiles to `0`
The `ton(string): Int` compile-time function compiles successfully when given an empty string:

```tact
contract TestTonEmpty {
    value: Int = ton("");
    receive() {}
}
```

The documentation shows usage examples like `"1"`, `"0.1"`, and `"0.000000001"`, but does not indicate that empty strings are valid. It is unclear whether `ton("") → 0` is intentional or a silent fallback.

### Expected Behavior

The compiler should reject `ton("")` with a clear compile-time error, since an empty string is not a valid human-readable Toncoin value.

### Actual Behavior

`ton("")` compiles without error and appears to return `0`.

### Suggested Resolution

Either:
- Disallow empty strings and emit a compile-time error, **or**
- Update the documentation to explicitly state that empty strings are accepted and produce `0`.

This is a **documentation mismatch** or potentially a **compiler bug**, and may lead to confusion or unintended behavior in financial logic.

---

i582:
## Issue

Closes #2538.

## Checklist

### DIFF
--- a/dev-docs/CHANGELOG.md
+++ b/dev-docs/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Language features
 
 - Optimized message deserialization with native loading of `Maybe Cell` fields: PR [#2661](https://github.com/tact-lang/tact/pull/2661)
+- [fix] Compiler now disallows `ton()` with empty or blank string: PR [#2681](https://github.com/tact-lang/tact/pull/2681)
 
 ### Standard Library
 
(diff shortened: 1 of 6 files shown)


Gusarich (approved):
LGTM!

anton-trunov (changes requested):
Let's also add some weird strings like `ton("asdf")` and numbers with weird symbols and/or arithmetic expressions: `ton("45,67")`, `ton("40+2")`

Issue #2675, opened 2025-04-08
Internal Compiler Error When Using Optional Trait Field Inside Struct


Gusarich:
Using an optional trait field inside a struct and assigning it `null` causes an **internal compiler error** instead of a proper semantic rejection.

---

### Minimal Reproducible Example

```tact
trait T {
  foo: Int;
}

struct Holder {
  t: T?;
}

contract TestContract {
  field: Holder;

  init() {
    self.field = Holder{t: null};
  }
}
```

---

### Actual Behavior

The compiler crashes with:

```
[INTERNAL COMPILER ERROR]: Unsupported type: T
Please report at https://github.com/tact-lang/tact/issues
```

---

### Expected Behavior

The compiler should reject this with a **clear, user-facing error** such as:

```
Struct field 't' has unsupported type: trait T
```

because, as per documentation, trait types are **not allowed** in storage or as struct fields.

---

### Why This Is a Problem

- Traits are not meant to be stored or passed around, including as `T?` inside structs.
- The compiler fails to catch this at semantic analysis and crashes during type resolution or backend generation.
- Internal compiler crashes must never occur due to user code — even invalid user code.

---

### Documentation Support

From the **Type System Overview**:

> While [contracts] and [traits] are also considered a part of the Tact type system, one cannot pass them around like [Structs and Messages].

From the **Contracts** section:

> Contracts in Tact are similar to classes ... they can't be passed around like Structs and Messages.

This confirms that trait types — including `T?` — are invalid in this context, but must be **handled gracefully**, not cause an ICE.

---

### Conclusion

This is a **confirmed compiler bug**. Optional trait fields used inside structs (even with `null`) currently crash the compiler instead of being properly rejected. The compiler must detect and reject this with a proper diagnostic.

---

LLM Fuzzing discovery (see #2490)

PR #2683, opened 2025-04-08
fix(codegen): fix optional struct fields in contracts with default value

### LINKED ISSUE #2541: Getter Function with Optional Struct Fails When Default Value Is Provided
A getter function returning an optional struct fails to compile if the struct field is initialized with a value.

### 🔁 Reproducible Example (Fails to Compile)

```tact
struct S {
    a: Int;
    b: Int;
}

contract TestGetterOptional {
    s: S? = S{ a: 1, b: 2 };

    receive() {}

    get fun getState(): S? {
        return self.s;
    }
}
```

**Compiler Error:**
```
FunC compilation error: ... error: previous function return type tuple cannot be unified with return statement expression type (int, int): cannot unify type (int, int) with tuple
```

### ✅ Works If Field Is Uninitialized

This variation compiles successfully:

```tact
struct S {
    a: Int;
    b: Int;
}

contract TestGetterOptional {
    s: S?;

    receive() {}

    get fun s(): S? {
        return self.s;
    }
}
```

### 🔎 Analysis

The compiler fails to unify the type of the initialized optional struct (`S? = S{...}`) with the expected return type of the getter. However, if the struct field is **not initialized**, the same getter compiles correctly.

This suggests a **type unification bug** affecting optional struct types with default values in get-methods.

### ✅ Expected Behavior

The following pattern should compile successfully:

```tact
get fun s(): S? {
    return self.s;
}
```

Even when `self.s` is initialized with `S{ ... }`.

---

### LINKED ISSUE #2676: FunC Crash When Assigning Non-Null Struct to Optional Persistent Field
Initializing an optional struct field in contract storage with a non-null value causes a **FunC compiler crash**.

---

```tact
struct Inner {
  x: Int;
}

contract TestContract {
  nestedOpt: Inner? = Inner{x: 99};
}
```

---

### Expected Behavior

This should compile. The documentation allows:

- Optional struct types in contract storage
- Inline initialization with either `null` or a struct instance

---

### Actual Behavior

Compilation crashes in the FunC backend:

```
[FunC compilation error: ... cannot unify type int with tuple](error: previous function return type tuple cannot be unified with return statement expression type int: cannot unify type int with tuple)
return ($self'nestedOpt);
```

---

### Why This Is a Bug

- `Inner? = null` works as expected
- `Inner? = Inner{...}` crashes
- Docs permit both — no error is shown in Tact phase
- The compiler generates invalid code, leading to a backend crash

---

### Conclusion

This is a **compiler bug**: assigning a struct value to an optional struct storage field should work but instead leads to an invalid FunC output and backend failure.

---

### LINKED ISSUE #2680: FunC Crash for Optional Struct Field Initialized with Struct Literal

```tact
struct SomeS {
    x: Int;
    y: Bool;
}

contract OptDefault {
    s: SomeS? = SomeS{ x: 42, y: true };

    init() {
        self.s = null;
    }
}
```

---

### Expected Behavior

- Optional struct fields should support default values of either `null` or a concrete struct.
- This pattern is valid: `SomeS? = SomeS{ ... }`.
- The compiler should accept and generate valid code for this.

---

### Actual Behavior

The compiler crashes in the FunC backend with:

```
FunC compilation error: ...
cannot assign an expression of type ??42 to a variable or pattern of type (int, int): cannot unify type ??42 with (int, int): cannot unify types of widths 1 and 2
```

- No error is produced at the Tact level.
- The backend fails due to optional-struct width mismatch in generated FunC.

---

### Why This Is a Bug

- The documentation explicitly allows `Struct?` as a type.
- Assigning a non-null struct literal to an optional field is consistent and expected behavior.
- The Tact compiler accepts it, but the backend fails to generate code for it.
- This is a classic case of a backend compiler bug: **valid code is accepted but fails during codegen**.

---

### Conclusion

This is a **confirmed compiler bug**: initializing an optional struct persistent field with a non-null struct value causes a backend crash. The compiler must either support this correctly or reject it at the Tact level with a proper diagnostic — but it must not crash.

---

i582:
## Issue

Closes #2541.
Closes #2680.
Closes #2676.
Towards #2675.

## Checklist


src/generator/writers/writeContract.ts
```diff
@@ -160,7 +160,13 @@ export function writeInit(
             t.fields.forEach((tField) => {
                 let init = "null()";
                 if (tField.default !== undefined) {
-                    init = writeValue(tField.default!, ctx);
+                    init = writeValue(
```
anton-trunov, line 163: perhaps `writeValue` should take in the type of value?

i582: It looks like we need to use slightly modified logic for this case:

https://github.com/tact-lang/tact/blob/e8d041f8af1a69971380e496103cbc035b59392e/src/generator/writers/writeExpression.ts#L165-L181

So passing only type is not easy here

anton-trunov:
merge conflicts in changelog :(

i582:
Oops

anton-trunov:
merge conflicts in changelog :(

i582:
> merge conflicts in changelog :(

Fixed

PR #2684, opened 2025-04-08
fix: add check for ton() with invalid number value or negative numbers

### LINKED ISSUE #2536: Compile-Time `ton()` Function Accepts Negative Values
`ton()` is intended to convert a human-readable Toncoin value (e.g. `"1"`, `"0.1"`, `"0.000000001"`) into its nanoTon `Int` format. All examples in the documentation show **non-negative values**, and negative coin amounts are generally **nonsensical** in the context of the TON blockchain.

However, the following test case **compiles without error**:

```tact
contract Test {
    neg: Int = ton("-1");
    receive() {}
}
```

### ✅ Expected Behavior

The compiler should reject negative input strings to `ton()` with a **compile-time error**, as negative Toncoin values are invalid.

### ❌ Actual Behavior

The compiler **accepts negative values** such as `ton("-1")` without error.

---

i582:
## Issue

Closes #2536.

## Checklist

### DIFF
--- a/src/types/__snapshots__/resolveDescriptors.spec.ts.snap
+++ b/src/types/__snapshots__/resolveDescriptors.spec.ts.snap
@@ -951,6 +951,33 @@ exports[`resolveDescriptors should fail descriptors for ton-with-empty-string 1`
 "
 `;
 
+exports[`resolveDescriptors should fail descriptors for ton-with-invalid-string 1`] = `
+"<unknown>:3:22: ton() function requires a valid number with no more than 10 digits after the decimal point
+  2 | 
+> 3 | const FOO: Int = ton("45,67");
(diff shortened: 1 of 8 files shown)


src/types/__snapshots__/resolveDescriptors.spec.ts.snap
anton-trunov, line 1: need a test with more than 10 digits after the decimal point

anton-trunov (changes requested):

src/types/__snapshots__/resolveDescriptors.spec.ts.snap
i582: added

anton-trunov (approved):

PR #2687, opened 2025-04-08
fix(interpreter): correctly process nested structs in interpreter

### LINKED ISSUE #2673: Internal Compiler Error When Using Nested Struct with Default Field Values

```tact
struct NestedDefaults {
    foo: Int = 1;
    inner: Inner = Inner{};
}

struct Inner {
    bar: Bool = false;
    text: String = "";
}

contract TestStorage {
    myVal: NestedDefaults;

    init() {
        self.myVal = NestedDefaults{};
    }
}
```

### Expected Behavior

- Structs with default field values should allow omitted fields when constructing.
- Since **all fields in `Inner` have defaults**, `Inner{}` should be valid.
- Since `inner` has a default value (`Inner{}`), `NestedDefaults{}` should also be valid.

This is supported by documentation, which explicitly says:

> "Structs can also contain default fields... if you have a lot of fields but don't want to keep specifying common values..."

There is no mention that defaults don’t apply recursively or cannot be used across nested struct instances.

### Actual Behavior

The compiler crashes with an internal error:

```
[INTERNAL COMPILER ERROR]: Struct value is missing a field: bar
```

### Why This Is Incorrect

- The behavior violates the documented default value rules.
- The compiler produces a hard **internal error** instead of a graceful rejection or type error.
- The bug only occurs due to **nested default values**, which should be fully supported or explicitly disallowed (with a proper message).

### Summary

This is a confirmed **compiler bug**. The compiler crashes when attempting to construct a struct that includes a nested struct with only default values. This violates documentation and breaks expected struct behavior. Nested default values should be supported or rejected explicitly — not crash the compiler.

---

i582:
## Issue

Closes #2673.

## Checklist

### DIFF
--- a/dev-docs/CHANGELOG.md
+++ b/dev-docs/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Optimized message deserialization with native loading of `Maybe Cell` fields: PR [#2661](https://github.com/tact-lang/tact/pull/2661)
 - [fix] Compiler now disallows `ton()` with empty or blank string: PR [#2681](https://github.com/tact-lang/tact/pull/2681)
 - [fix] Compiler now disallows `ton()` with invalid number value or negative numbers: PR [#2684](https://github.com/tact-lang/tact/pull/2684)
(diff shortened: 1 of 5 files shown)


anton-trunov:
merge conflicts in changelog :)

i582:
> merge conflicts in changelog :)

Fixed

novusnota (approved):
approved by a :trollface:

anton-trunov (approved):

PR #2694, opened 2025-04-09
fix: better error message for `Slice? as remaining`

### LINKED ISSUE #2674: Internal Compiler Error When Using `Slice? as remaining` in Struct
Declaring a struct with an optional `Slice` field marked `as remaining` causes an **internal compiler error**, despite no documentation indicating that this usage is invalid.

---

```tact
struct OptRem {
    tail: Slice? as remaining;
}
```

---

### Actual Behavior

The compiler crashes with:

```
[INTERNAL COMPILER ERROR]: "as remaining" cell cannot be optional
```

This is a hard crash, not a user-facing compile-time error.

---

### Expected Behavior

- The documentation explains that a `remaining` field must be the **last field** in a struct and only one is allowed.
- It does **not state** that `remaining` fields **must not be optional**.
- Optionals are generally allowed in structs and for `Slice` types.
- The compiler should either support this or reject it cleanly with a proper error message.

---

### Why This Is a Problem

- There is **no documented restriction** against using `?` on a `remaining` field.
- The placement is correct (last field, single occurrence).
- An internal compiler error indicates a missing safety check in the compiler.
- This breaks the language's guarantees about robustness and developer feedback.

---

### Conclusion

This is a confirmed **internal compiler bug and documentation gap**. If `Slice? as remaining` is unsupported, the compiler should provide a proper error — not crash. If it's intended to be supported, it must be fixed to compile successfully. Either way, the current behavior is incorrect.

---

i582:
## Issue

Closes #2674.

## Checklist

### DIFF
--- a/src/types/resolveSignatures.ts
+++ b/src/types/resolveSignatures.ts
@@ -217,6 +217,17 @@ export function resolveSignatures(ctx: CompilerContext, Ast: FactoryAst) {
             }
         }
 
+        const lastField = t.fields.at(-1);
+        if (lastField?.as === "remaining") {
+            const type = lastField.type;
+            if (type.kind === "ref" && type.optional) {
+                throwCompilationError(
+                    `The "as remaining" field cannot have optional type`,
+                    lastField.ast.type.loc,
+                );
+            }
+        }
+
         const fields = t.fields.map((v) => createTLBField(v.abi));
 
         // Calculate signature and method id
--- a/src/types/resolveSignatures.ts
+++ b/src/types/resolveSignatures.ts
@@ -217,14 +217,15 @@ export function resolveSignatures(ctx: CompilerContext, Ast: FactoryAst) {
             }
         }
 
-        const lastField = t.fields.at(-1);
-        if (lastField?.as === "remaining") {
-            const type = lastField.type;
-            if (type.kind === "ref" && type.optional) {
-                throwCompilationError(
-                    `The "as remaining" field cannot have optional type`,
-                    lastField.ast.type.loc,
-                );
+        for (const field of t.fields) {
+            if (field.as === "remaining") {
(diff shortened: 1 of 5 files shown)


anton-trunov (approved):

src/types/resolveSignatures.ts
```diff
@@ -217,6 +217,17 @@ export function resolveSignatures(ctx: CompilerContext, Ast: FactoryAst) {
             }
         }
 
+        const lastField = t.fields.at(-1);
```
anton-trunov: Wait, let's actually ban it for any field, not just the last one

anton-trunov (changes requested):

src/types/resolveSignatures.ts
i582: If you write `Slice? as remaining` not in the last field, compiler will give an error already

i582:
```
Error: test/main.tact:75:5: The "as remaining" field can only be the last field of the message
  74 | message Name {
> 75 |     bar: Slice? as remaining;
           ^~~~~~~~~~~~~~~~~~~~~~~~
  76 |     foo: Int;
```

i582: So, it is enough to check only the last field :)

anton-trunov: > If you write Slice? as remaining not in the last field, compiler will give an error already

We might re-enable using `as remaining` anywhere, it's actually can be quite useful

anton-trunov: Also, this makes this check orthogonal which is good from the design prospective

i582: Done

anton-trunov (approved):

PR #2695, opened 2025-04-09
fix: better error message for trait typed fields

### LINKED ISSUE #2675: Internal Compiler Error When Using Optional Trait Field Inside Struct

i582:
## Issue

Closes #2675.

## Checklist

### DIFF
--- a/src/types/test-failed/message-field-with-optiona-trait-type.tact
+++ b/src/types/test-failed/message-field-with-optiona-trait-type.tact
@@ -1,0 +1,7 @@
+trait BaseTrait {}
+
+trait Test {}
+
+message TakeEscrowData {
+    data: Test?;
+}
--- a/dev-docs/CHANGELOG.md
+++ b/dev-docs/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [fix] Compiler now correctly processes nested structs with default values in the interpreter: PR [#2687](https://github.com/tact-lang/tact/pull/2687)
 - [fix] Compiler now correctly compiles contracts with optional struct fields with default values: PR [#2683](https://github.com/tact-lang/tact/pull/2683)
 - [fix] Compiler now shows a more informative error message for unsupported assembly functions inside traits and contracts: PR [#2689](https://github.com/tact-lang/tact/pull/2689)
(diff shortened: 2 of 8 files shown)


src/types/test-failed/message-field-with-optiona-trait-type.tact
```diff
@@ -0,0 +1,7 @@
+trait BaseTrait {}
```
novusnota: File name:

```
with-optiona-trait
//         ^
```

i582: fixed :)

novusnota (dismissed):

anton-trunov (changes requested):
Let's add tests with structures too

i582:
> Let's add tests with structures too

Added

anton-trunov (dismissed):

anton-trunov (approved):

verytactical (approved):

PR #2702, opened 2025-04-09
feat: add raise() function

### LINKED ISSUE #2701: Throw analogue, that uses text, not integer exit codes
I think good name for it is `raise("Error")`

Shvandre:
## Issue

Closes #2701.

### DIFF
--- a/src/abi/global.ts
+++ b/src/abi/global.ts
@@ -111,6 +111,45 @@ export const GlobalFunctions: Map<string, AbiFunction> = new Map([
         },
     ],
     [
+        "raise",
+        {
+            name: "raise",
+            resolve: (_ctx, args, ref) => {
+                if (args.length !== 1) {
+                    throwCompilationError(
+                        "raise() expects single string argument",
+                        ref,
+                    );
+                }
+                const arg0 = args[0]!;
+                if (arg0.kind !== "ref") {
+                    throwCompilationError(
+                        "raise() expects single string argument",
+                        ref,
+                    );
+                }
+                if (arg0.name !== "String") {
+                    throwCompilationError(
+                        "raise() expects single string argument",
+                        ref,
+                    );
+                }
+                return { kind: "void" };
+            },
+            generate: (ctx, _args, resolved, ref) => {
+                if (resolved.length !== 1) {
+                    throwCompilationError(
+                        "raise() expects single string argument",
+                        ref,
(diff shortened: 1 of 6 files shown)


src/types/resolveErrors.ts
```diff
@@ -1,7 +1,8 @@
 import type { CompilerContext } from "@/context/context";
 import { createContextStore } from "@/context/context";
 import type { AstNode } from "@/ast/ast";
-import type { FactoryAst } from "@/ast/ast-helpers";
+import type {FactoryAst } from "@/ast/ast-helpers";
```
verytactical: `yarn fmt`

Shvandre: Already done

src/test/e2e-emulated/builtins/intrinsics.tact
```diff
@@ -1,3 +1,5 @@
+message TestRaise{}
```
verytactical: .

novusnota:
:boar: docs incoming

UPD: Done!

anton-trunov (changes requested):
1. needs a CHANGELOG entry marked as a **breaking change**
2. needs a test what happens if the user already defined a function like `fun raise(s: String) { throw 42 }`

src/abi/global.ts
```diff
@@ -110,6 +110,45 @@ export const GlobalFunctions: Map<string, AbiFunction> = new Map([
...
+                return { kind: "void" };
+            },
+            generate: (ctx, _args, resolved, ref) => {
+                if (resolved.length !== 1) {
+                    throwCompilationError(
+                        "raise() expects single string argument",
+                        ref,
+                    );
+                }
+                const resolved0 = resolved[0]!;
```
i582, line 146: `arg`

src/abi/global.ts
```diff
@@ -110,6 +110,45 @@ export const GlobalFunctions: Map<string, AbiFunction> = new Map([
             },
         },
     ],
+    [
+        "raise",
+        {
+            name: "raise",
+            resolve: (_ctx, args, ref) => {
+                if (args.length !== 1) {
+                    throwCompilationError(
```
i582, line 119: we need a test for each new throw

anton-trunov:
will solve this with macros

PR #2706, opened 2025-04-09
fix(wrappers): export all tuple parsers/load/store functions from TypeScript wrappers

### LINKED ISSUE #2559: Export helper functions for serialization/deserialization from TS wrappers
Image

i582:
## Issue

Closes #2559.

## Checklist


verytactical (approved):
~~Maybe one of the `load` functions is extraneous though. I couldn't find what it's used for.~~

A very specific case of getter return value serialization is handled by TVM/emulator itself, so there long tuples are not split into small parts, but in rest of the code they are.

novusnota (approved):

Issue #1546, opened 2025-01-23
Build refactor


verytactical:
@byakuren-hijiri: Document and refactor the frontend pipeline. Currently, the way the compiler prepares to calling the `precompile` function is quite convoluted. All the arguments of the functions used before `precompile` should be documented, we should extract small methods when possible, and document them, to make that API more useful for third-parties.

@byakuren-hijiri: Extract methods from the `build` function (`src/pipeline/build.ts`) to make it more modular. We need to separate the build functionality from CLI parsing and use different methods to create context, compile, and precompile. This is important to implement in order to enable third-party tools to hook into the compilation pipeline in the most flexible way. Perhaps, the best way to achieve this functionality is to create the `Builder` class with public methods defining the pipeline.

novusnota:
Great suggestion from the past, and it's already addressed [here](https://github.com/tact-lang/tact/pull/1363)!

verytactical:
Thanks, I forgot to link it.

PR #2717, opened 2025-04-10
refactor: simplify `precompile()` and `build()` signatures

i582:
## Issue

Towards #1546


novusnota:
This is a dangerous move, since `build()` function is used in: Misti, Web IDE, Verifier, etc.

I'd advise to make a separate, non-public `build()` and `precompile()` functions instead.

UPD: Since we don't give any guarantees on internal APIs, that's ok, I guess

i582:
> This is a dangerous move, since `build()` function is used in: Misti, Web IDE, Verifier, etc.
> 
> I'd advise to make a separate, non-public `build()` and `precompile()` functions instead.

And they should update its usage accordingly :)

anton-trunov (approved):

PR #2729, opened 2025-04-10
refactor: better test setup

verytactical:


src/test/utils/cache-state.ts
anton-trunov, line 1: merged this file into `main`

anton-trunov:
already started adapting this approach

PR #2730, opened 2025-04-11
fix(codegen): `dump` for optionals

### LINKED ISSUE #2639: Attemp to call `dump()` on an optional which is `null` at runtime throws exit code 7: Type check error
```tact
// Empty map
let fizz: map<Int, Int> = emptyMap();

// Setting a couple of values under different keys
fizz.set(7, 70);
fizz.set(42, 42);

// Overriding one of the existing key-value pairs
let oldVal1 = fizz.replaceGet(7, 68); // key 7 now points to value 68
dump(oldVal1); // 70

// Trying to replace the value of a non-existing key-value pair will do nothing
let oldVal2 = fizz.replaceGet(8, 68); // no key 8, so nothing was altered,
                                      // and oldVal2 is null
dump(oldVal2); // exit code 7
```

Notice that if we use the non-null assertion operator `!!` on `oldVal2`, we get an exit code 128. That is, the `oldVal2` is indeed `null` and the issue is, presumably, somewhere in the `dump()`.

Related to #2628

anton-trunov:
## Issue

Closes #2639.


src/generator/writers/writeStdlib.ts
```diff
@@ -229,6 +234,40 @@ export function writeStdlib(ctx: WriterContext): void {
...
+    ctx.fun("__tact_debug_int", () => {
+        ctx.signature(
+            `() __tact_debug_int(int number, slice debug_print_1, slice debug_print_2)`,
+        );
+        ctx.flag("impure");
+        ctx.context("stdlib");
+        ctx.body(() => {
+            ctx.write(`
+                if (null?(number)) {
+                    ${ctx.used("__tact_debug_str")}("null", debug_print_1, debug_print_2);
```
i582: Let's extract this logic since it is used many times

dev-docs/CONTRIBUTING.md
```diff
@@ -128,6 +128,17 @@ We use [Jest](https://jestjs.io) as our testing framework. Jest supports a combi
...
 Some tests are put in the same folder as the implementation and can be located in the `*.spec.ts` files; other tests are grouped into categories in the [src/test](../src/test) folder. The [project map section](#project-map) has more information on tests relevant to each compiler component.
 
+To rebuild the compiler and run end-to-end tests, the usual workflow is as follows:
+
+```shell
+# build the compiler, but skip typechecking of the test files
+yarn build:fast
+# build the test contracts and benchmarks
+yarn gen:fast
+# run all tests, expect the slow property-based map tests
```
novusnota:
```suggestion
# run all tests except the slow property-based map tests
```

On another note, do we want a `yarn all:fast`? Like: 

`"all:fast": "yarn build:fast && yarn gen:fast && yarn test:fast && yarn lint:all"`

novusnota (changes requested):

dev-docs/CONTRIBUTING.md
anton-trunov: maybe in a different PR, I only tried to streamline the workflow that was needed for this PR

i582 (dismissed):

package.json
```diff
@@ -30,16 +31,16 @@
     "gen:contracts:benchmarks:tact": "ts-node src/benchmarks/bench.build.ts",
     "gen:contracts:benchmarks:func": "ts-node src/benchmarks/func.build.ts",
     "gen:contracts:benchmarks": "yarn gen:contracts:benchmarks:tact && yarn gen:contracts:benchmarks:func",
+    "gen:contracts:fast": "yarn gen:contracts:test && yarn gen:contracts:benchmarks",
     "gen:contracts:test:map": "ts-node ./src/test/e2e-slow/map-property-tests/generate.ts",
-    "gen:contracts:all": "yarn gen:contracts:test && yarn gen:contracts:benchmarks && yarn gen:contracts:test:map",
+    "gen:contracts:all": "yarn gen:contracts:fast && yarn gen:contracts:test:map",
     "gen": "yarn gen:grammar && yarn gen:stdlib && yarn gen:func-js && yarn gen:contracts:all",
```
novusnota, line 37: There's no `gen:fast` command, which contradicts the text in `CONTRIBUTING.md`:

image

novusnota (changes requested):

PR #2735, opened 2025-04-11
fix(benches): tlb types for jettons

### LINKED ISSUE #2734: Fix `JettonBurnNotification` and related things in benchmarks
Currently, addresses in `JettonBurn` and `JettonBurnNotification` are represented as `Address` whereas they should be `Address?` to support the `MsgAddress` from the Jetton standards TL-B.

This fix is needed for #2509.

Kaladin13:
## Issue

Closes #2734


anton-trunov (changes requested):
If this is a fix, where is the corresponding test?

Kaladin13:
Added test

anton-trunov:
looks like the new test is failing

src/test/tep/jettons.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+        };
+
+        snapshot = blockchain.snapshot();
+    });
+
+    beforeEach(async () => {
+        await blockchain.loadFrom(snapshot);
+    });
+
+    it("Should correctly handle response destination as addr_none", async () => {
```
anton-trunov: What is the correct way to handle it? One should be able to understand if the code below actually tests it

anton-trunov (changes requested):

src/test/tep/jettons.spec.ts
Kaladin13: Added descriptive naming

skywardboundd:
If we’re testing the TEP-74 (describe) standard here, I think we should make sure all aspects of it are explicitly covered in the tests to ensure full compliance with the spec.

PR #2739, opened 2025-04-11
fix(stdlib): doc comments and parameter names

### LINKED ISSUE #2664: Find and correct all parameter name-type mismatches
For example, we have a `storeBuilder(self: Builder, cell: Builder)`, where `cell` should be named `builder`.

Since those parameters are positional and not named, we can change the names without breaking the backwards compatibility with any code that uses the respective functions.

anton-trunov:
## Issue

Closes #2664.


i582 (dismissed):

docs/src/content/docs/ref/core-cells.mdx
```diff
@@ -283,19 +283,22 @@ let buzz: Builder = b.storeBit(false); // writes 0
 ### Builder.storeBuilder
 
 ```tact
-extends fun storeBuilder(self: Builder, cell: Builder): Builder;
+extends fun storeBuilder(self: Builder, other: Builder): Builder;
 ```
 
-Extension function for the [`Builder{:tact}`][builder] type.
+Extension function for the `Builder` type.
```
novusnota:
```suggestion
Extension function for the [`Builder{:tact}`][builder] type.
```

Let's keep the first link in.

novusnota (changes requested):

docs/src/content/docs/ref/core-send.mdx
```diff
@@ -192,10 +192,11 @@ Various niche, dangerous, or unstable features which can produce unexpected resu
 <Badge text="Gas-expensive" title="Uses 500 gas units or more" variant="danger" size="medium"/><p/>
 
 ```tact
-fun nativeSendMessage(cell: Cell, mode: Int);
+fun nativeSendMessage(msg: Cell, mode: Int);
 ```
 
-[Queues the message](/book/send#outbound-message-processing) to be sent by specifying the complete `cell` and the [message `mode`](/book/message-mode).
+[Queues the `msg` message](/book/send#outbound-message-processing) to be sent by specifying the complete message cell.
```
novusnota: Hmm, why this change? We don't give raw links in the docs whenever possible

docs/src/content/docs/ref/core-send.mdx
```diff
@@ -192,10 +192,11 @@ Various niche, dangerous, or unstable features which can produce unexpected resu
...
+The `mode` parameter configures how to send a message, see: https://docs.tact-lang.org/book/message-mode.
```
novusnota:
```suggestion
[Queues the message](/book/send#outbound-message-processing) to be sent by specifying the complete `msg` cell and the [message `mode`](/book/message-mode).
```

src/stdlib/stdlib/std/internal/send.tact
```diff
@@ -476,9 +476,10 @@ asm fun cashback(to: Address) {
 
 /// Global function.
 ///
-/// Queues the message to be sent by specifying the complete `cell` and the message `mode`.
+/// Queues the `msg` message to be sent by specifying the complete message cell.
+/// The `mode` parameter configures how to send a message, see: https://docs.tact-lang.org/book/message-mode.
```
novusnota:
```suggestion
/// Queues the message to be sent by specifying the complete `msg` cell and the message `mode`.
```

novusnota (changes requested):

docs/src/content/docs/ref/core-send.mdx
anton-trunov: we just should generate the reference docs from the source code, it's just mind-numbing to do the same work twice

docs/src/content/docs/ref/core-send.mdx
anton-trunov: nope, I'm not doing it here, once @i582 implements the generator, this will go away

novusnota (approved):
Nice work, thanks!

docs/src/content/docs/ref/core-send.mdx
novusnota: > nope, I'm not doing it here, once i582 implements the generator, this will go away

But I will. Also, we'll need more talks regarding generator to preserve:
1. links in the notes
2. other stuff (will discuss later anyways, this msg would instantly get lost)

i582 (approved):

PR #2740, opened 2025-04-11
feat(ci): do not download the published npm version of Tact in CI

### LINKED ISSUE #1810: Don't download the published version of Tact in CI
Instead, we need to use the `main` branch and always link with it. This is important for reducing the npm download count from the usage of tact-template and Blueprint in the test suite.

That said, the command for creating a Blueprint project might have to be changed, otherwise it'll still download the Tact compiler there no matter what we do. **This part is a blocker.**

novusnota:
## Issue

Closes #1810.


novusnota:
fixin'
UPD: done

.github/actions/test/action.yml
```diff
@@ -32,7 +32,7 @@ runs:
       shell: bash
       run: |
         yarn clean
-        ${{ runner.os != 'Windows' && 'time ' || '' }} yarn cross-env BUILD_THREADS=4 yarn gen
```
novusnota, line 35: we no longer need to time the build, hence this removal

novusnota:
image

Windows CI ran long enough to encounter the storage fees.

anton-trunov:
I tried, but I have no idea what's going on here. We need vibe reviewing from @copilot or @i582

novusnota:
We can also unite the main `tact` job with its Windows counterpart in the same manner as I did unite workflows of Node.js package managers in this PR.

At the discretion of @xpyctumo

xpyctumo:
> We can also unite the main tact job with its Windows counterpart in the same manner as I did unite workflows of Node.js package managers in this PR.

Yes, please. Use it from my PR. I fixed it [here](https://github.com/tact-lang/tact/blob/85d8145bc3e5e82785514d864d40042fa1075651/.github/workflows/tact.yml).

In my PR it would be unnecessary right now

PR #2742, opened 2025-04-11
fix(docs): clarify that one cannot declare local variables without immediate initialization

### LINKED ISSUE #2557: Cannot Declare Local Variable Without Initializer — Contradicts Documentation
The Tact documentation shows that variables can be declared without immediate initialization using syntax like:

```tact
let value: Int; // declaration only
```

However, in practice, this form fails to compile — even in simple cases.

### 🔁 Minimal Reproducible Example

```tact
contract TestVarDecl {
    receive() {
        let value: Int; // declaration
        value = 5;      // assignment
    }
}
```

### ❌ Actual Behavior

The compiler returns a syntax error:

```
Error: Expected "<", "=", "?", or "as" at the variable declaration line.
```

This suggests that uninitialized declarations are not supported — despite being documented.

### ✅ Suggested Fix

Update the documentation to clearly state that all local variables must be initialized on declaration.

---

novusnota:
## Issue

Closes #2557.


anton-trunov (dismissed):

docs/src/content/docs/book/learn-tact-in-y-minutes.mdx
```diff
@@ -583,9 +583,12 @@ Read more: [Expressions](/book/expressions).
 
 ```tact
 fun showcase() {
-    // let statement to define new variables, as we've seen above.
-    let theAnswer = 42;                // type ascription is optional for everything,
-    let m: map<Int, Int> = emptyMap(); // except for optionals and maps
+    // As we've seen above, the let statement defines new variables.
+    // You must always provide an initial value, but type ascriptions
+    // aren't mandatory for most cases, except for optionals and maps.
```
i582: hmm, not "optionals", only for `null` AFAIK

i582:
```tact
let a = getMaybeInt()
```
is ok

docs/src/content/docs/book/statements.mdx
```diff
@@ -11,11 +11,11 @@ The following statements can appear anywhere in a [function](/book/functions) bo
 
 The `let{:tact}` statement allows local and [block](#block)-scoped variable declaration.
 
-In Tact, declaring a local variable always requires an initial value. However, the type ascription can be omitted, and Tact will try to infer it from the initial value:
+In Tact, declaring a local variable always **requires** an initial value. However, the type ascription can be omitted, and Tact will try to infer it from the initial value:
```
i582, line 14: maybe add info about maps and `nulls`?

novusnota: It's right below, in the next example :)

docs/src/content/docs/book/learn-tact-in-y-minutes.mdx
novusnota: Huh, good point

novusnota: Fixed!

anton-trunov (approved):

PR #2748, opened 2025-04-12
feat(docs): `as bytes32` and `as bytes64` storage annotations

anton-trunov:
## Issue

Closes #1178.

PR #2750, opened 2025-04-12
feat(docs): document "unreachable code"

### LINKED ISSUE #2535: "Unreachable Code" Compiler Error is Not Documented
Example of code that produces unreachable code error:

```tact
fun test3(): Int {
    return 10;
    return 20;
}
```

This is expected, but documentation does not state anywhere that this can happen. It only says "The return statement ends function execution and specifies a value to be returned to the function caller.".

---

novusnota:
## Issue

Closes #2535


docs/src/content/docs/book/cells.mdx
```diff
@@ -236,11 +236,24 @@ Also, note that specifying `remaining{:tact}` for a `Cell{:tact}` type used as a
 
 The `as bytes32{:tact}` serialization annotation can only be applied to slices and is used to read 32 bytes (256 bits) of data from a cell. The `as uint256{:tact}` annotation achieves a similar result, but it can only be applied to integers.
 
+```tact
+struct Example {
+    b32: Slice as bytes32; // 256 bits with arbitrary contents
+    u256: Int as uint256;  // also 256 bits, but viewed as a single integer
+}
+```
+
```
anton-trunov: I don't understand this example. What practical problem does it solve?

docs/src/content/docs/book/cells.mdx
```diff
@@ -236,11 +236,24 @@ Also, note that specifying `remaining{:tact}` for a `Cell{:tact}` type used as a
...
 The `as bytes64{:tact}` serialization annotation can only be applied to slices and is used to read 64 bytes (512 bits) of data from a cell. It can be useful for serializing and deserializing cryptographic signatures.
 
-For a usage example, see the [`SignedBundle`](/ref/core-crypto#signedbundle) [struct][struct].
+```tact
+struct Example {
+    b64: Slice as bytes64; // 512 bits with arbitrary contents
+}
+```
+
+For an advanced usage example, see the [`SignedBundle`](/ref/core-crypto#signedbundle) [struct][struct].
```
anton-trunov: Same here, this does not explain anything. It's better to just click the link and see a _real practical_ example

anton-trunov (changes requested):

anton-trunov:
> And a small follow-up on https://github.com/tact-lang/tact/pull/2748

Please no follow-ups without an issue. Please focus on the existing issues

docs/src/content/docs/book/cells.mdx
novusnota: > I don't understand this example. What practical problem does it solve?

Gives visual input to the text before

docs/src/content/docs/book/statements.mdx
```diff
@@ -53,6 +53,29 @@ fun getTimeFromNow(offset: Int): Int {
...
+If the function does not have an explicit return type, it has an implicit return type of `void`. As such, the `return{:tact}` statement must be empty.
+
+```tact
+extends mutates fun correct(self: Int) {
+    if (self == 42) {
+        return;
+    }
+
+    self = 42;
+}
```
anton-trunov: need a better example. try to look in for examples in the benchmarks or at least in the end-to-end tests

anton-trunov: A better example would include an `if` statement with two reachable returns in its branches vs just one reachable return 

also, `throw` should be showcased here

anton-trunov: It needs to be said the `return`-reachability analysis is not almighty. And show an example where for a human it is more or less obvious a function always returns but the analysis rejects it

docs/src/content/docs/book/statements.mdx
```diff
@@ -53,6 +53,29 @@ fun getTimeFromNow(offset: Int): Int {
...
+```
+
+All statements after the `return{:tact}` statement are unreachable and will not be executed. Such statements are detected by the Tact compiler, producing an "Unreachable code" error.
+
+This is done to help avoid potential logical errors in the code.
+
+```tact
+fun doubleTrouble(): Int {
+    return 10;
+    return 20; // COMPILATION ERROR! Unreachable statement
```
anton-trunov: This example is too obvious and definitely not practical

docs/src/content/docs/book/cells.mdx
anton-trunov: > Gives visual input to the text before

It's either an example that can be adopted by our user or no example at all. This is going to be our policy for any docs update from now on

novusnota: ...although I don't fully agree here, I might've found a better example. Trying to add it :)

anton-trunov (approved):

PR #2770, opened 2025-04-14
fix: throw an error if non-optional method is called on an optional type

### LINKED ISSUE #2766: Forbid calling an extension function for type T on an expression with type T?
Example:
```tact
contract Test {
    addr: Address?;

    receive() {
        if (self.addr.isNotNone()) {
            throw(10);
        }
    }
}

inline extends fun isNotNone(self: Address): Bool { return self.asSlice().preloadUint(2) != 0 }
```

i582:
## Issue

Closes #2766.

## Checklist

### DIFF
--- a/src/types/__snapshots__/resolveStatements.spec.ts.snap
+++ b/src/types/__snapshots__/resolveStatements.spec.ts.snap
@@ -82,7 +82,7 @@ exports[`resolveStatements should fail statements for bounced-type-is-smaller 1`
 `;
 
 exports[`resolveStatements should fail statements for call-non-optional-method-on-optional 1`] = `
-"<unknown>:10:13: Cannot call method "isNotNone" on expression with type "Address?" without unwrapping, add "!!" to unwrap optional
+"<unknown>:10:13: Cannot call method "isNotNone" on an expression of type "Address?" (optional Address) without unwrapping it first. Add "!!" before the method call to unwrap an optional value
    9 |     receive() {
 > 10 |         if (self.addr.isNotNone()) {
                    ^~~~~~~~~~~~~~~~~~~~~
--- a/src/types/__snapshots__/resolveStatements.spec.ts.snap
+++ b/src/types/__snapshots__/resolveStatements.spec.ts.snap
@@ -81,6 +81,15 @@ exports[`resolveStatements should fail statements for bounced-type-is-smaller 1`
 "
 `;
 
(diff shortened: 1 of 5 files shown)


src/types/resolveExpression.ts
```diff
@@ -643,6 +643,18 @@ function resolveCall(
                 checkParameterType(exp.args[i]!, a, ctx);
             }
 
+            if (f.self) {
+                const selfIsOptional =
+                    f.self.kind === "ref" ? f.self.optional : false;
+
+                if (src.optional && !selfIsOptional) {
+                    throwCompilationError(
+                        `Cannot call method ${idTextErr(exp.method)} on an expression of type "${src.name}?" (optional ${src.name}) without unwrapping it first. Add "!!" before the method call to unwrap an optional value`,
```
novusnota:
```suggestion
                        `Cannot call method ${idTextErr(exp.method)} on an expression of type "${src.name}?" (optional ${src.name}) without unwrapping it with "!!" first`,
```

src/types/__snapshots__/resolveStatements.spec.ts.snap
```diff
@@ -81,6 +81,15 @@ exports[`resolveStatements should fail statements for bounced-type-is-smaller 1`
 "
 `;
 
+exports[`resolveStatements should fail statements for call-non-optional-method-on-optional 1`] = `
+"<unknown>:10:13: Cannot call method "isNotNone" on an expression of type "Address?" (optional Address) without unwrapping it first. Add "!!" before the method call to unwrap an optional value
```
novusnota:
```suggestion
"<unknown>:10:13: Cannot call method "isNotNone" on an expression of type "Address?" (optional Address) without unwrapping it unwrapping it with "!!" first
```

novusnota (changes requested):
simplified the error message

src/types/__snapshots__/resolveStatements.spec.ts.snap
i582: Done

src/types/resolveExpression.ts
i582: Done

anton-trunov (dismissed):

novusnota (dismissed):

anton-trunov (approved):

PR #2773, opened 2025-04-14
fix: improved error when inheriting from two traits with the same method name

### LINKED ISSUE #2629: Misleading Error Message for Conflicting Trait Functions
When two traits define a function with the same name and a contract implements both, the compiler correctly identifies the conflict — but the **error message is misleading**.

Given the following code:

```tact
trait TraitA {
    fun f(): Int {
        return 1;
    }
}

trait TraitB {
    fun f(): Int {
        return 2;
    }
}

contract TestContract() with TraitA, TraitB {
    // nothing
}
```

The compiler outputs:

```
Error: test.tact:2:5: Function "f" is already defined in trait "TraitB"
  1 | trait TraitA {
> 2 |     fun f(): Int {
          ^~~~~~~~~~~~~~
  3 |         return 1;
```

### What’s Wrong

- The message says `Function "f" is already defined in trait "TraitB"` — but it's pointing to the definition in `TraitA`, which came first.
- It should say the function is **already defined in `TraitA`**, and the error is occurring **in `TraitB`**, since that’s the conflicting definition.

This creates confusion by reporting the right problem but pointing to the wrong location.

### ✅ Expected

The error should say:

```
Error: test.tact:8:5: Function "f" is already defined in trait "TraitA"
  7  | trait TraitB {
> 9  |     fun f(): Int {
          ^~~~~~~~~~~~~~
  10 |         return 2;
```

This would make the source of the problem and the compiler’s interpretation much clearer.

---

i582:
## Issue

Closes #2629.


anton-trunov (changes requested):
Let's add more tests with two traits that have the same function identifier, but with `abstract` and `virtual` attributes in various combinations

PR #2774, opened 2025-04-14
feat(stdlib): add `ecrecover()` and `normalizeECDSARecoveryId()` functions

### LINKED ISSUE #2733: Add stdlib function for `ECRECOVER`
This instruction does not work as expected, so we can add function to the stdlib with a correct implementation and describe it in the documentation. This way, users won't get confused when trying to use this instruction and will be able to find a ready-made working solution in stdlib.

We can also update our fork of tvm-spec.

See https://github.com/ton-blockchain/ton/issues/1126

i582:
## Issue

Closes #2733.

## Checklist


novusnota:
Do we want tests for this? :)

anton-trunov (changes requested):
Tests needed :)

docs/src/content/docs/ref/core-crypto.mdx
```diff
@@ -180,6 +180,56 @@ sha256(someVariableElsewhere); // will try to resolve at compile-time,
 
 :::
 
+## normalizeECDSARecoveryId
```
verytactical, line 183: Abbreviations in camelcase must be capitalized as regular words, i.e. `normalizeEcdsaRecoveryId`

src/stdlib/stdlib/std/internal/crypto.tact
```diff
@@ -235,3 +235,82 @@ struct SignedBundle {
...
+///
+/// * `hash`: 256-bit message hash
+/// * `v`: recovery ID (must be `0` or `1`, use `normalizeECDSARecoveryId()` to convert from `27`/`28`)
+/// * `r`, `s`: 256-bit signature values
+///
+/// Returns `null` on failure or an `EcrecoverKey` structure on success.
+///
+/// Note: This function uses the TVM `ECRECOVER` instruction and therefore requires `v` to
+/// be already normalized with [`normalizeECDSARecoveryId()`](#normalizeecdsarecoveryid).
+asm fun ecrecover(hash: Int, v: Int, r: Int, s: Int): EcrecoverKey? {
```
verytactical, line 306: We shouldn't name functions in standard library after assembly instructions. `ECRECOVER` is a misnomer, as we're not recovering elliptic cryptography with it.

novusnota: There's a builtin [`ecrecover()`](https://docs.soliditylang.org/en/latest/units-and-global-variables.html#mathematical-and-cryptographic-functions) function in Solidity, and people know that `ecrecover` and reference it elsewhere.

image

But I agree that we might need to change the name since ECRECOVER on TON does a slightly different thing

Issue #2514, opened 2025-03-27
Add keccak512, sha512 and blake2b functions, and implement runtime logic of `sha256()` in one asm function


novusnota:
As we need some pre-processing for the `sha256()` function to support passing a Slice or a String, we'll probably have to do the same for the remaining functions.

The `sha256()` could look like this in FunC:

```func
int onchain_sha256(slice data) asm """
    ONE
    <{
        OVER SREFS 0 NEQINT
    }> PUSHCONT
    <{
        OVER LDREF s0 POP CTOS s0 s1 XCHG INC
    }> PUSHCONT
    WHILE
    HASHEXT_SHA256
""";
```

Alternatively, we can simply make all those functions operate on Slices only, since converting from a String to Slice is a no-op and always accessible. That way, those functions can be implemented via asm functions in Tact.

For example:

```tact
asm fun keccak256(data: Slice): Int {
    ONE
    <{
        OVER SREFS 0 NEQINT
    }> PUSHCONT
    <{
        OVER LDREF s0 POP CTOS s0 s1 XCHG INC
    }> PUSHCONT
    WHILE
    HASHEXT_KECCAK256
}
```

novusnota:
The 512 versions that produce two integers require either introducing a `struct Pair { a: Int; b: Int }` or depending on #1032 and #1116

PR #2775, opened 2025-04-14
feat(stdlib): optimize `sha256()` and add `keccak256()`

### LINKED ISSUE #2767: Optimize onchain logic of `sha256()` and introduce the `keccak256()` function
```tact
asm fun sha256(data: Slice): Int {
    <{
        <{
           DUP SREFS 
        }> PUSHCONT
        <{
           LDREFRTOS 
        }> PUSHCONT
        WHILE
        DEPTH
        HASHEXT_SHA256 // replace this with HASHEXT_KECCAK256, profit!
    }>CONT 1 1 CALLXARGS
}
```

Wrapping the code in `<{ ... }>CONT 1 1 CALLXARGS` is needed for the `DEPTH` instruction to calculate only the depth of the given continuation, not the whole stack.

_Inspired by: https://github.com/pyAndr3w/stdlib.func/blob/main/crypto.func_

novusnota:
## Issue

Closes #2767.

## Checklist


anton-trunov (changes requested):
let's not change the FunC benchmarks in this PR

novusnota:
> let's not change the FunC benchmarks in this PR

The FunC benchmarks do not use the `sha256()` in any way, but they're compiled with `stdlib.fc` and `stdlib_ex.fc` that Tact use and that somehow affects them. I've opened a separate issue for that #2776.

If I revert the changes to benchmarked JSONs then CI won't pass :)

UPD: Looking for the source of trouble. As @anton-trunov pointed out, the reason might be that FunC's dead code elimination is quirky.

anton-trunov:
> UPD: Looking for the source of trouble. As @anton-trunov pointed out, the reason might be that FunC's dead code elimination is quirky.

Sorry, my bad. We just had a debugging session with @i582 and turns out the issues is the code dict internal structure is changed because of the changing method_ids

Shvandre:
🐗

src/test/e2e-emulated/stdlib/crypto.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+        blockchain,
+        treasury,
+        contract,
+    };
+};
+
+describe("crypto.tact: keccak256() function", () => {
+    const state = cached(setup);
+
+    it("should hash values correctly", async () => {
```
skywardboundd, line 46: Let add some random data for test

Shvandre (dismissed):

src/test/e2e-emulated/stdlib/crypto.spec.ts
novusnota: So, I went other way around — I've added a check against an actual audited Keccak-256 implementation to the tests. Seems working fine, plus there are other nice hashes in the same lib, such as an implementation of the [BLAKE2b](https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE2b_algorithm), which we can use later when implementing #2514.

Shvandre (approved):

PR #2777, opened 2025-04-14
compile: use native func-fift compiler for faster (2-3x times) compilation of FunC code

i582:


src/func/funcCompile.ts
```diff
@@ -60,6 +62,73 @@ export async function funcCompile(args: {
     entries: string[];
     sources: { path: string; content: string }[];
     logger: ILogger;
+}): Promise<FuncCompilationResult> {
+    const USE_NATIVE = process.env.USE_NATIVE === "true";
```
verytactical: "USE_NATIVE" is not specific to Tact, and can clash with other usages of the name.
image

src/func/funcCompile.ts
```diff
@@ -60,6 +62,73 @@ export async function funcCompile(args: {
...
+    return funcCompileWasm(args);
+}
+
+export function funcCompileNative(args: {
+    entries: string[];
+    sources: { path: string; content: string }[];
+    logger: ILogger;
+}): FuncCompilationResult {
+    const TACT_PATH = process.env.TACT_PATH;
+    if (typeof TACT_PATH === "undefined") {
```
verytactical: Isn't TACT_PATH already known by a code that is running out of TACT_PATH?

src/func/funcCompile.ts
i582: Fixed

src/func/funcCompile.ts
i582: done (somehow)

src/func/funcCompile.ts
```diff
@@ -60,6 +62,70 @@ export async function funcCompile(args: {
...
+    }
+    const FUNC_FIFT_COMPILER_PATH = process.env.FUNC_FIFT_COMPILER_PATH;
+    if (typeof FUNC_FIFT_COMPILER_PATH === "undefined") {
+        throw new Error("FUNC_FIFT_COMPILER_PATH is not set");
+    }
+
+    const files: string[] = args.entries;
+    const configStr = JSON.stringify({
+        sources: files.map((f) =>
+            f.replace("@stdlib/", path.join(TACT_PATH, "src/stdlib/stdlib/")),
```
jeshecdom, line 91: In our fuzzer, we use a custom FunC stdlib. So, instead of using `TACT_PATH`, I generalize it to an arbitrary FunC path, and replace this line like this:
```
f.replace("@stdlib/", FC_STDLIB_PATH),
```
Or if you prefer to use relative to `TACT_PATH`:
```
f.replace("@stdlib/", path.join(TACT_PATH, FC_STDLIB_PATH)),
```

anton-trunov:
This is not going to be useful for us in the long run. Let's just keep this branch for now as it might be useful for fuzzing but close this PR.

PR #2778, opened 2025-04-14
test: integrate Allure reporting for test results

### LINKED ISSUE #2741: Make Allure example
Make a small example, where

- at least one test is exported to [Allure](https://allurereport.org)
- there is a `step()` function that reports steps into test log
- `yarn allure` can be used instead of `yarn test` to run tests and open HTML report in browser

xpyctumo:
## Issue

Allure Implementation

Closes #2741 

## Checklist
- [x] I have documented my contribution in docs/ and made the build locally

## Allure metadata checklist

To ensure that the tests are rich, I suggest dividing the list into folders with `*.spec.ts` files, inside which we will need to add Allure data, for example, `step()` or `parameter()`

- [ ] pipeline
- [ ] vfs
- [ ] grammar
- [x] types
- [ ] test
- [ ] optimizer
- [ ] fmt
- [ ] server
- [ ] utils
- [ ] imports
- [ ] cli
- [ ] storage
- [ ] generator
- [ ] benchmarks
- [ ] ast
- [ ] error
- [ ] func


.github/workflows/tact.yml
```diff
@@ -74,54 +74,52 @@ jobs:
...
       - name: Test compatibility with tact-template
         run: |
-          # !!!!!!! Don't forget to update test-windows below !!!!!!!
           git clone https://github.com/tact-lang/tact-template.git
           cd tact-template
-          # Hijack the dependency on Tact to avoid redundant npm downloads
-          jq 'del(.dependencies."@tact-lang/compiler")' package.json -M > temp.json
-          mv temp.json package.json
+          # Modify package.json to hijack the dependency on Tact
+          if [[ "${{ matrix.os }}" == "windows-latest" ]]; then
```
novusnota: to use bash syntax you need bash shell for all platforms to be set explicitly via `shell: bash` (Git Bash on Windows), which will fix the issues with this `if`, but will break its body.

xpyctumo: Yes, we can, but i'm not sure about `jq` in Windows

xpyctumo: Okay, let's test it!

.github/workflows/tact.yml
```diff
@@ -102,11 +102,14 @@ jobs:
         shell: bash
         run: |
           yarn global add allure-commandline
```
novusnota: Why not `npm i -g allure-commandline`? It's usually better to install global CLIs via `npm`, regardless of the package manager used in the project IMHO

xpyctumo: Good point. I was focused on the neighboring commands used in the step. Using `npm` is a good suggestion, let's do it!

.github/workflows/tact.yml
```diff
@@ -78,54 +78,45 @@ jobs:
...
-          cd tact-template
-          # Hijack the dependency on Tact to avoid redundant npm downloads
-          jq 'del(.dependencies."@tact-lang/compiler")' package.json -M | Set-Content temp.json
-          Move-Item temp.json package.json -Force
           yarn install
           yarn link @tact-lang/compiler
           yarn build
           yarn test
-
+      - name: Install Allure
```
xpyctumo: Drop CI

jest.config.js
```diff
@@ -6,10 +7,26 @@ module.exports = {
     moduleNameMapper: {
         "^@/(.*)$": "<rootDir>/src/$1",
     },
-    testEnvironment: "node",
+    testEnvironment: "allure-jest/node",
```
xpyctumo, line 10: Test allure reporting for `-fast` config

xpyctumo: ALL FAST config

jest.config.js
```diff
@@ -6,10 +7,26 @@ module.exports = {
...
+            os_platform: os.platform(),
+            os_release: os.release(),
+            os_version: os.version(),
+            node_version: process.version,
+        },
+    },
     testPathIgnorePatterns: ["/node_modules/", "/dist/"],
     globalSetup: "./jest.globalSetup.js",
     setupFiles: ["./jest.setup.js"],
+    setupFilesAfterEnv: ["./src/test/allure/jest.allure.ts"],
```
xpyctumo: Move this to top level, not `src`

package.json
```diff
@@ -43,6 +43,7 @@
     "to-relative": "ts-node src/to-relative.build.ts",
     "test": "jest",
     "test:fast": "jest --config=./jest-fast.config.js",
+    "test:allure": "rm -rf ./allure-results && yarn test && allure serve allure-results --single-file",
```
xpyctumo: `rm -rf` no need here, how about Windows users?

https://www.npmjs.com/package/rimraf

src/test/allure/jest.allure.ts
```diff
@@ -0,0 +1,66 @@
...
+import * as allure from "allure-js-commons";
+import { Blockchain } from "@ton/sandbox";
+
+declare global {
+    interface BigInt {
+        toJSON(): number;
+    }
+}
+
+BigInt.prototype.toJSON = function () {
```
xpyctumo: No prototype modifications

src/test/allure/jest.allure.ts
```diff
@@ -0,0 +1,66 @@
...
+    return Number(this);
+};
+const originalCreate: typeof Blockchain.create = Blockchain.create;
```
xpyctumo: as const

src/test/allure/jest.allure.ts
```diff
@@ -0,0 +1,66 @@
...
+
+Blockchain.create = function proxy(
+    this: typeof Blockchain,
+    ...args: Parameters<typeof Blockchain.create>
+) {
+    return allure.step("Blockchain.create()", () =>
+        originalCreate.apply(this, args),
+    );
+} as typeof Blockchain.create;
```
xpyctumo: is `as` needed here?

src/test/allure/jest.allure.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const originalTreasury: typeof Blockchain.prototype.treasury =
+    Blockchain.prototype.treasury;
```
xpyctumo:
```suggestion
const originalTreasury = Blockchain.prototype.treasury;
```

src/test/utils/write-vm-log.ts
```diff
@@ -16,31 +17,33 @@ export const writeLog = ({ blockchain, path }: WriteLogParams) => {
...
-            console.log = (message: unknown, ...rest: unknown[]) => {
-                if (typeof message !== "string" || rest.length > 0) {
-                    throw new Error("Unexpected log");
-                }
-                const lines = message
-                    .split("\n")
-                    .map((s) => "    " + s)
-                    .join("\n");
-                const entry = `  - |\n${lines}\n`;
+        return await allure.step(name, async () => {
```
xpyctumo: Revert this and create step in separate file

src/types/effects.spec.ts
```diff
@@ -8,8 +8,10 @@ import { getParser } from "@/grammar";
 import { getAstFactory } from "@/ast/ast-helpers";
 import { computeReceiversEffects } from "@/types/effects";
 import type { Source } from "@/imports/source";
+import * as allure from "allure-js-commons";
 
 describe("effects", () => {
+    allure.tags("Effects");
```
xpyctumo: revert this

src/test/utils/write-vm-log.ts
xpyctumo: And make step async

jest.config.js
xpyctumo: UPD: It's working with -fast config

And Allure report will be empty if `testEnvironment: "node"`

jest.config.js
xpyctumo: This file no more need

.github/workflows/tact.yml
```diff
@@ -78,54 +78,29 @@ jobs:
...
       matrix:
         node-version: [22]
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/test
       - name: Test compatibility with tact-template
+        shell: bash
```
verytactical: Not everyone uses `bash` on windows. If we have CI this way, we lose checks that everything works in powershell.

xpyctumo: That file was reverted

verytactical (dismissed):

verytactical (approved):

PR #2786, opened 2025-04-15
refactor(test): add sqrt benchmark

### LINKED ISSUE #2765: Add `sqrt` microbenchmarks
These microbenchmarks can be added to https://github.com/tact-lang/tact/blob/main/src/test/gas-consumption/gas.spec.ts.

See https://github.com/tact-lang/tact/issues/770#issuecomment-2481863626

#### TonoxDeFi/open-contracts/math.func

https://github.com/TonoxDeFi/open-contracts/blob/ebbb877e5ee6a9e506605c455c44b2b5223956f1/contracts/math/math.func#L26

```func
int math::sqrt(int x) inline {
  if (x == 0) { return x; }

  int r = 181;
  int xx = x;
  if (xx >= 0x100000000000000000000000000000000) {
    xx >>= 128;
    r <<= 64;
  }
  if (xx >= 0x10000000000000000) {
    xx >>= 64;
    r <<= 32;
  }
  if (xx >= 0x100000000) {
    xx >>= 32;
    r <<= 16;
  }
  if (xx >= 0x10000) {
    xx >>= 16;
    r <<= 8;
  }

  r = (r * (xx + 65536)) >> 18;

  repeat(7) {
    r = (r + x / r) >> 1;
  }

  int r1 = x / r;

  return (r < r1 ? r : r1);
}
```

#### mathlib.fc

https://github.com/ton-blockchain/ton/blob/ed4682066978f69ffa38dd98912ca77d4f660f66/crypto/smartcont/mathlib.fc#L205

```func
;; computes sqrt(a*b) exactly rounded to the nearest integer
;; for all 0 <= a, b <= 2^256-1
;; may be used with b=1 or b=scale of fixed-point numbers
int geom_mean(int a, int b) inline_ref {
  ifnot (min(a, b)) {
    return 0;
  }
  int s = log2_floor_p1(a);   ;; throws out of range error if a < 0 or b < 0
  int t = log2_floor_p1(b);
  ;; NB: (a-b)/2+b == (a+b)/2, but without overflow for large a and b
  int x = (s == t ? (a - b) / 2 + b : 1 << ((s + t) / 2));
  do {
    ;; if always used with b=2^const, may be optimized to "const LSHIFTDIVC#"
    ;; it is important to use `muldivc` here, not `muldiv` or `muldivr`
    int q = (muldivc(a, b, x) - x) / 2;
    x += q;
  } until (q == 0);
  return x;
}

;; integer square root, computes round(sqrt(a)) for all a>=0.
;; note: `inline` is better than `inline_ref` for such simple functions
int sqrt(int a) inline {
  return geom_mean(a, 1);
}
```

skywardboundd:
Closes #2765


src/test/gas-consumption/gas.spec.ts
```diff
@@ -294,4 +295,26 @@ describe("benchmarks", () => {
...
+    ];
+
+    it.each(variants)("benchmark %s", async (name, message) => {
+        const instance = blockchain.openContract(await Sqrt.fromInit());
+
+        const sendResult = await step(name, () =>
+            instance.send(
+                treasure.getSender(),
+                { value: toNano(1) },
+                { $$type: message, value: 100n },
```
anton-trunov: Let's also try a few different values: `0`, `MAX_INT`, typical Toncoin values, for instance

anton-trunov (changes requested):

Shvandre (dismissed):
LGTM

src/test/gas-consumption/contracts/sqrt.tact
```diff
@@ -0,0 +1,132 @@
+// func reference
+
+// int math::sqrt(int x) inline {
```
Kaladin13: Use multiline comment here /**/

src/test/gas-consumption/contracts/sqrt.tact
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    let r1: Int = x / r;
+
+    return (r < r1 ? r : r1);
+}
+
+asm fun log2_floor_p1(x: Int): Int { UBITSIZE }
+
+// ;; computes sqrt(a*b) exactly rounded to the nearest integer
```
Kaladin13: Same multiline here

Kaladin13 (changes requested):
Please refactor comemnts

PR #2799, opened 2025-04-15
refactor(benches): unify escrow

### LINKED ISSUE #2793: Unify FunC and Tact benchmarks: escrow

skywardboundd:
Closes #2793


src/benchmarks/escrow/escrow.spec.ts
```diff
@@ -62,90 +62,87 @@ describe("Escrow Gas Tests", () => {
...
-        );
-
-        const deployResult = await deployer.send({
-            to: escrowAddress,
-            value: toNano("0.1"),
-            init,
-            body: beginCell().endCell(),
-            sendMode: SendMode.PAY_GAS_SEPARATELY,
-        });
+        if (implementation === "func") {
```
verytactical: In order to prove our tact implementation is equivalent to func implementation, they must use absolutely identical tests.

There must be no `implementation === "func"` in tests

skywardboundd: Our contracts written on Tact use a different structure for storing data(c4), so this is necessary. And we load another code, not from wrappers

anton-trunov:
merge conflict

.vscode/extensions.json
```diff
@@ -1,8 +0,0 @@
-{
```
reverofevil, line 1: Don't remove it please

verytactical (approved):

PR #2797, opened 2025-04-15
feat(fmt): support directory formatting

i582:
## Issue

Closes #2787.

PR #2806, opened 2025-04-15
feat(fmt): `--check` flag to check formatting on CI

### LINKED ISSUE #2788: Add `--check` flag to formatter
With this flag, the formatter will only check formatting and exit with code 1 if some files are not formatted

i582:
## Issue

Closes #2788.


src/cli/tact-fmt/index.ts
```diff
@@ -84,17 +152,50 @@ const parseArgs = (Errors: FormatterErrors, Args: Args) => {
...
+        if (onlyCheck) {
+            console.log("Checking formatting...");
+        }
+
+        if (!fs.statSync(filePath).isFile()) {
+            const files = globSync(["**/*.tact"], {
+                cwd: filePath,
+            });
+
+            let wasError = false;
```
novusnota: `hasErrored`?

i582: `someFileCannotBeFormatted`
https://github.com/tact-lang/tact/pull/2797/commits/bc9e004796672566261701c864a4a6215cbb1f3e

src/cli/colors.ts
```diff
@@ -2,7 +2,7 @@ import { release } from "os";
 
 export const isColorSupported = () => {
     if (process.env.CI) {
-        return true;
+        return process.stdout.isTTY && process.env.TERM !== "dumb";
```
verytactical: Please remove `process.env.CI` check instead.

src/cli/tact-fmt/index.ts
```diff
@@ -47,6 +47,7 @@ const processArgs = (Errors: FormatterErrors, argv: string[]) => {
 const ArgSchema = (Parser: ArgParser) => {
     return Parser.tokenizer
         .add(Parser.boolean("write", "w"))
+        .add(Parser.boolean("check", "c"))
```
verytactical, line 50: It should be an opt-out, not an opt-in.

When a person calls `fmt` on their files, they expect the tool will only check the files. To fix them, they'll call it with `--fix`.

anton-trunov:
```
❯ yarn prettier --help
yarn run v1.22.22
$ /Users/anton-m1/ton/tact/node_modules/.bin/prettier --help
Usage: prettier [options] [file/dir/glob ...]

By default, output is written to stdout.
Stdin is read if it is piped to Prettier and no files are given.

Output options:

  -c, --check              Check if the given files are formatted, print a human-friendly summary
                           message and paths to unformatted files (see also --list-different).
  -l, --list-different     Print the names of files that are different from Prettier's formatting (see also --check).
  -w, --write              Edit files in-place. (Beware!)
```

src/cli/colors.ts
i582: Done

src/cli/tact-fmt/index.ts
```diff
@@ -80,7 +82,11 @@ type Args = ArgConsumer<GetParserResult<ReturnType<typeof ArgSchema>>>;
 
 const markup = getAnsiMarkup(Boolean(isColorSupported()));
 
-function formatFile(filepath: string, write: boolean): boolean | undefined {
+function formatFile(
+    filepath: string,
+    write: boolean,
+    onlyCheck: boolean,
```
anton-trunov: Looks like this should be one `mode` parameter instead of these two. For instance, `write` and `onlyCheck` both `true` do not make sense.

anton-trunov (changes requested):

src/cli/tact-fmt/index.ts
i582: fixed

PR #2810, opened 2025-04-16
fix: show full error message for map with optional value

### LINKED ISSUE #2534: Invalid contract field type leads to non-explaining error
Code:

```tact
contract TestContract {
    m: map<Int, Int?>; // Int? is invalid for values

    receive () {}
}
```

No explanations in error:

```
Tact compilation failed
Error: src/test/e2e-emulated/contracts/test.tact:2:8: 
  1 | contract TestContract {
> 2 |     m: map<Int, Int?>;
             ^~~~~~~~~~~~~~
  3 | 

    at /opt/homebrew/lib/node_modules/@tact-lang/compiler/dist/grammar/index.js:799:19
    ...
```

i582:
## Issue

Closes #2534.

## Checklist

### DIFF
--- a/src/grammar/__snapshots__/grammar.spec.ts.snap
+++ b/src/grammar/__snapshots__/grammar.spec.ts.snap
@@ -677,6 +677,15 @@ exports[`should fail literal-underscore-after-leading-zero.fail 1`] = `
 "
 `;
 
+exports[`should fail optional-map-value.fail 1`] = `
+"<unknown>:2:8: value cannot be optional
+  1 | contract TestContract {
+> 2 |     m: map<Int, Int?>; // Int? is invalid for values
+             ^~~~~~~~~~~~~~
+  3 | 
+"
+`;
+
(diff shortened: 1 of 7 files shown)


src/grammar/__snapshots__/grammar.spec.ts.snap
```diff
@@ -677,6 +677,15 @@ exports[`should fail literal-underscore-after-leading-zero.fail 1`] = `
 "
 `;
 
+exports[`should fail optional-map-value.fail 1`] = `
+"<unknown>:2:8: value cannot be optional
```
anton-trunov: Is it possible to point at the `Int?` here? And fix the error message to something like `map value types cannot be optional`

anton-trunov (changes requested):

src/grammar/__snapshots__/grammar.spec.ts.snap
i582: Done

PR #2844, opened 2025-04-18
feat: optimize the ` if` statement with `ifnot` if possible

### LINKED ISSUE #2795: Optimize if (a == 0)
By replacing 0 EQINT / IF with IFNOT, we can save 8 gas.

Additionally, this can be addressed at the FunC level by using the ifnot() helper directly in the generated code, which will emit the optimal TVM bytecode automatically.

i582:
## Issue

Closes #2795.


src/generator/writers/writeFunction.ts
```diff
@@ -557,6 +557,33 @@ export function writeStatement(
...
+const rewriteWithIfNot = (expr: Ast.Expression): [Ast.Expression, string] => {
+    if (expr.kind === "op_unary" && expr.op === "!") {
+        // `if (~ cond)` => `ifnot (cond)`
+        return [expr.operand, "ifnot"];
+    }
+    if (
+        expr.kind === "op_binary" &&
+        expr.op === "==" &&
+        expr.right.kind === "number" &&
+        expr.right.value === 0n
```
anton-trunov: - We should also take care of the symmetrical `0 == a` case.
- The optimization should work for any compile-time expression that evaluates to `0`, see how it is implemented in `writeExpressionInCondition`

src/generator/writers/writeFunction.ts
```diff
@@ -557,6 +557,33 @@ export function writeStatement(
...
+    ) {
+        // if (a == 0) => ifnot (a)
+        return [expr.left, "ifnot"];
+    }
+    if (
+        expr.kind === "op_binary" &&
+        expr.op === "!=" &&
+        expr.right.kind === "number" &&
+        expr.right.value === 0n
```
anton-trunov: Same as above

anton-trunov (changes requested):
In addition to generalizing the optimization a bit, let's also add some more tests, especially with nested `if`s with `== 0` and `!= 0` `else if`-branches.

src/generator/writers/writeFunction.ts
i582: Done

src/generator/writers/writeFunction.ts
i582: Done

anton-trunov (approved):

PR #2862, opened 2025-04-19
feat(stdlib): add stubs.tact file with definition of all built-in functions in Tact

### LINKED ISSUE #202: Eventually, docs in `stubs/` have to be moved to the compiler repo

### LINKED ISSUE #2861: Create a stdlib file containing definitions for all built-in functions in Tact
Currently Tact has many functions like `ton`, `address` or `Foo.toCell` that don't actually have a definition. This is a problem for the Language Server, as it needs an anchor for documentation or go-to-definition

We need to create a file with all these definitions for LS and other tools.

See https://github.com/tact-lang/tact-language-server/issues/202

i582:
## Issue

Closes #2861.
Closes https://github.com/tact-lang/tact-language-server/issues/202.


novusnota (approved):
This will allow the language server to display information relevant to the actual compiler version, although you'd need to keep the stubs.tact for Tact 1.6.6 and lower.

PR #2875, opened 2025-04-21
fix(benches): wallet v5 return instead of throw

### LINKED ISSUE #2843: Change v5 wallet Tact implementation in benchmarks to follow FunC logic
We need to:

- Return instead of throw in internal receiver
- Check for signature on a message from extension

Kaladin13:
## Issue

Closes #2843


Kaladin13:
image

It's sad that we lose 2.5% on external transfer bench, when the changed code don't even affect the tested data flow, e.g. it's because of final after-fift boc methods reordering

Kaladin13:
review: add tests for exist + return in /test/

anton-trunov (approved):

Issue #1241, opened 2024-12-24
Map expressions


anton-trunov:
```tact
// or map<K, V> {
map {  
  k1: v1,
  ...
  [kn]: vn, // <- optional trailing comma
}
```

where `k1`, ..., `kn` -- key expressions, `v1`, ..., `vn` -- value expressions. If a simple literals and Identifiers are used as key expressions, we the syntax allows using them directly and if those are complex expressions, then the user needs to wrap those in square brackets (see an example below). 

If all the key and value expressions can be computed at compile-time, it should translate into a constant cell primitive, a-la `cell(...)`, otherwise we desugar something like `let m: ... = map { [x + 1]: 42, [b]: y - 5, 24: 0 }` as follows:

```tact
let m: ... = emptyMap();
m.set(x + 1, 42);
m.set(b, y - 5);
m.set(24, 0);
```

verytactical:
[Reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#computed_property_names)

novusnota:
> [Reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#computed_property_names)

i.e. the current syntax of Struct and Message instantiations (instance expressions), but without the names.

To reduce ambiguity, I'd consider adding an explicit `map<K, V>` before, like:

```tact
let m = map<Int, Cell> {
    5: cell("..."),
    6: cell("..."),
};
```

Or maybe without `<K, V>`, since that can be inferred and because `map` is a reserved word:

```tact
let m: map<Int, Cell> = map {
    5: cell("..."),
    6: cell("..."),
};
```

And for the case of keys of `Address` type:

```tact
let m: <Address, Cell> = map {
    address("..."): cell("..."),
    address("..."): cell("..."),
    // or even this
    "EQ...": cell("..."),
};
```

Additionally, there's a question of whether this should still be considered a `map<K, V>` or `Cell` in the Tact's type system. I'd say this should be a `map<K, V>`, and if user wants to, they can add explicit `.asCell()` cast as such:

```tact
let c: Cell = map<Int, Cell> {
    5: cell("..."),
    6: cell("..."),
}.asCell();
```

anton-trunov:
@verytactical  Our map literals are only compile-time. This is a more general issue

PR #2881, opened 2025-04-21
feat: map literals

### LINKED ISSUE #2900: Compile-time map literals

verytactical:
Towards #1241.

Closes #2900.


i582:
Related to https://github.com/tact-lang/tact/issues/1241

src/generator/writers/writeExpression.ts
```diff
@@ -186,45 +195,27 @@ export function writeValue(
...
             return value;
         }
-        default:
-            throwInternalCompilerError("Unrecognized ast literal kind");
+        case "map_value": {
+            if (typeof val.bocHex === "undefined") {
+                return "null()";
+            }
+            const res = writeCell(
+                Cell.fromBoc(Buffer.from(val.bocHex, "hex"))[0]!,
```
i582:
```suggestion
                Cell.fromHex(val.bocHex),
```

src/generator/writers/writeExpression.ts
```diff
@@ -829,13 +814,143 @@ export function writeExpression(
...
+    };
+
+const writeMapLiteral =
+    (node: Ast.MapLiteral) =>
+    (_ctx: WriterContext): string => {
+        throwCompilationError(
+            "Only constant map literals are supported",
+            node.loc,
+        );
+        // const { fields, loc } = node;
```
i582, line 837: commented code

src/generator/writers/writeExpression.ts
```diff
@@ -829,13 +814,143 @@ export function writeExpression(
...
+): string {
+    // literals and constant expressions are covered here
+
+    // FIXME: Once optimization step is added, remove this try and replace it with this
+    // conditional:
+    // if (isLiteral(f)) {
+    //    return writeValue(f, wCtx);
+    // }
+    try {
+        const util = getAstUtil(getAstFactory());
```
i582, line 938: reuse `constEval()`

src/generator/writers/writeExpression.ts
verytactical: Intentionally left there.

src/generator/writers/writeExpression.ts
verytactical: That code was there before. We're not doing side refactoring here.

src/test/e2e-emulated/maps/map-literals.tact
```diff
@@ -0,0 +1,93 @@
...
+    get fun address(): map<Address, Int as coins> {
+        return map<Address, Int as coins> {
+            address("EQDtFpEwcFAEcRe5mLVh2N6C0x-_hJEM7W61_JLnSF74p4q2"): ton("1"),
+            address("EQAYqo4u7VF0fa4DPAebk4g9lBytj2VFny7pzXR0trjtXQaO"): ton("2"),
+        };
+    }
+
+    get fun large(): map<Int as int257, Int as int257> {
+        return map<Int as int257, Int as int257> {
+            // prettier-ignore
```
i582: `// fmt-ignore`

src/generator/writers/writeExpression.ts
i582: >  We're not doing side refactoring here.

What about splitting of `writeExpression` in this PR :D

src/generator/writers/writeExpression.ts
i582: add comment about this then

i582:
Also please support it in the formatter

novusnota (changes requested):
Docs are needed — I'll add them first thing in the morning

UPD: Almost done
UPD: Done

novusnota (commented):
What happens if maps are initialized with the same key? Would the first value be taken, the last one, or would there be an error?

There are no such tests yet, hence the questions :)

UPD: There would be no error and the last value would be taken:

```tact
const EXAMPLE: map<Int, Int> = map<Int, Int> {
    1: 2,
    1: 3,
};

// somewhere in a receiver/getter:
dump(EXAMPLE.get(1)!!); // 3
```

anton-trunov:
> UPD: There would be no error and the last value would be taken

correct

novusnota (commented):
Since map literals are resolved at compile-time and let statements declaring new maps do not require type ascriptions, we can (probably) allow omitting type ascriptions for constants too.

If that's too much of a hassle, then nevermind, but it'd be great to:

```tact
// Write this
const EXAMPLE = map<Int, Int> {
    1: 2,
    1: 3,
};

// And not this
const EXAMPLE: map<Int, Int> = map<Int, Int> {
    1: 2,
    1: 3,
};
```

Perhaps, removal of type ascriptions for constants can be fine-tuned to only allow the MapLiteral ast node as the right-side expression.

novusnota (commented):
I hereby approve the docs, but we might add more tests:
1. When there are same keys, only the last entry is used
2. Ternary expression with a void return value in either of branches

docs/src/content/docs/book/maps.mdx
```diff
@@ -180,7 +180,7 @@ fun getExchangesList(): map<Address, Cell> {
 }
 ```
 
-Finally, serialization annotations of key and value types are allowed with the [`as{:tact}` keyword](/book/integers#serialization).
+Because maps are represented as cells with binary data, `Int` and `Uint` types must be used with [storage annotations](/book/integers#serialization).
```
novusnota, line 176: UInt vibe coding I see.
And there's no "must"


UPD: https://github.com/tact-lang/tact/issues/2931

docs/src/content/docs/book/maps.mdx
```diff
@@ -127,18 +127,18 @@ Note that [persistent state variables](/book/contracts#variables) of type `map<K
 map<K, V> { key1: value1, key2: value2, /* ... */, keyN: valueN };
 ```
 
-Map literals (or initializers) are expressions that provide a concise way to create maps. First goes the `map<K, V>{:tact}` type expression, then a curly-brace enclosed comma-delimited list of zero or more predefined key-value pairs.
+Map literals are expressions that provide a concise way to create maps. First goes the `map<K, V>{:tact}` type expression, then a curly-brace enclosed comma-delimited list of zero or more predefined key-value pairs.
 
 ```tact
 // Declaring a map as a local variable via a map literal expression
-let myMap: map<Int, Int> = map<Int, Int> {
+let myMap: map<Int as uint8, Int as uint8> = map<Int as uint8, Int as uint8> {
```
novusnota, line 134: Hey, that's shown later

UPD: https://github.com/tact-lang/tact/issues/2931

novusnota:
:rocket:

PR #2886, opened 2025-04-21
feat: add an ` opcode` method for any message to obtain message opcode

### LINKED ISSUE #2230: Static field for message opcodes
So, instead of 

```tact
const TakeWalletAddressOpcode: Int = 0xd1735400;

message(TakeWalletAddressOpcode) TakeWalletAddress {
    queryId: Int as uint64;
    walletAddress: Address;
    ownerAddress: Cell?; //It is Maybe ^Address, just encoded it like this
}
```

the user should be able to use

```tact
TakeWalletAddress.opcode
```

i582:
```tact
message(0x777) TripleAxe {
    prize: Int as uint32;
}

contract Example {
    receive(msg: TripleAxe) {
        dump(TripleAxe.opcode()); // 0x777
   }
}
```

## Why not field

- The method is easier to implement, since we already have the `Message.fromSlice()` method and code for it
- The method is easier to support in the Language Server (just one change in the stubs!)
- The method is more consistent with other methods of any messages

## Issue

Closes #2230.

## Checklist


docs/src/content/docs/ref/core-cells.mdx
```diff
@@ -2025,6 +2025,30 @@ fun cautiousParse(payload: Slice): TripleAxe? {
 }
 ```
 
+## Message.opcode
+
```
novusnota, line 2029:
```suggestion

<Badge text="Available since Tact 1.7" variant="tip" size="medium"/><p/>

```

src/stdlib/stdlib/std/stubs.tact
```diff
@@ -394,6 +394,26 @@ fun AnyMessage_fromCell(cell: Cell): M;
 ///
 fun AnyMessage_fromSlice(slice: Slice): M;
 
+/// Extension function for any Message type.
```
novusnota:
```suggestion
/// Extension function for any Message type. Available since Tact 1.7.0.
```

novusnota (changes requested):

novusnota (dismissed):

anton-trunov (changes requested):
Let's add a few more things to this PR:
- negative tests, like using the `.opcode()` method on weird types, like `Int`, some structs, instances of messagges;
- `Message.opcode()` should be available at compile time, so we need some tests for `const`s;
- We should check things like the ones if `src/types/test/message-opcode-expr.tact` interact well with the new method

i582:
Added more tests :)

As we discussed earlier, to support the method at compile time we need to move the method ID calculation logic from codegen to a separate middle-end pass, this will be done in the next PR to avoid overcomplicating this one

anton-trunov (dismissed):

anton-trunov:
something wrong with snapshots?

anton-trunov (approved):

PR #2515, opened 2025-03-27
perf(stdlib): internal trait functions in `base.tact`

skywardboundd:
Closes #2511

PR #2918, opened 2025-04-23
perf(stdlib): base trait 

skywardboundd:
Follow-up to #2515


src/stdlib/stdlib/std/internal/base.tact
```diff
@@ -5,6 +5,116 @@
 /// and a constant `self.storageReserve` aimed at advanced users of Tact.
 ///
 /// See: https://docs.tact-lang.org/ref/core-base/
+/// 
+struct MessageParametersWithoutBounce {
```
novusnota, line 9: I'd say such stdlib changes aren't worth it — consider utilizing `writeStdlib.ts` and producing necessary functions in the generated FunC output instead

skywardboundd: It's just template, but thx for idea

PR #2932, opened 2025-04-24
fix: maximum size of useful data in bounced messages is 224 bits, not bytes

### LINKED ISSUE #1139: Update description of `bounced()` message receiver
It currently only describes how to accept the raw `Slice`  and the truncated to 224 bits `bounced<Message>`. But if the message occupies less than 224-bits, than it can be specified there directly, without the special type constructor `bounced<…>`.

Additionally, the whole incoming bounced message body has 256 bits, where 32 bits are commonly used by the message opcode, and the remaining 224 bits are usable for the rest of the data.

Image

`<T>` → `<M>`
Addition: "If the bounced message would fit into..."

Some prior discussion (and links to implementation in Tact 1.1.0!): https://github.com/tact-lang/tact/issues/29

Add more examples from: https://github.com/tact-lang/tact/issues/1675

### LINKED ISSUE #2911: Incorrect message: "Maximum size of bounced message is 224 bytes"
It should be `bits`, not `bytes`.

Clarify why 224 and not 256 (or fix it).

anton-trunov:
Closes #2911.
Closes #1139.

### DIFF
--- a/dev-docs/CHANGELOG.md
+++ b/dev-docs/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Language features
 
+- Fixed incorrect error message for bounced messages: PR [#2932](https://github.com/tact-lang/tact/pull/2932)
+- Added compile-time map literals: PR [#2881](https://github.com/tact-lang/tact/pull/2881)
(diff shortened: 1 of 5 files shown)


Gusarich:
isn't it 256 bits actually?

anton-trunov:
We don't count the first 32 bits which are 0xFFFFFFFF as useful, so it's not 256 bits in our docs and error messages

novusnota (approved):
:shipit:

PR #2954, opened 2025-04-27
fix(imports): prevent duplicate imports within the same file

### LINKED ISSUE #2939: Reject duplicate imports
```tact
// foo.tact
const X: Int = 2;
```

```tact
// test.tact
import "./foo";
import "./foo";
import "./foo.tact";

contract Bar {
    x: Int = X;
}
```

Since multiple imports of the same file should redefine the same things, those should be rejected.

murka:
Closes: #2939


src/imports/resolveImports.ts
```diff
@@ -45,6 +46,15 @@ export function resolveImports({
                 continue;
             }
 
+            // Check for duplicates within the current file
+            if (resolvedPathsInCurrentFile.has(resolved.path)) {
+                throwCompilationError(
+                    `Duplicate import of '${resolved.path}' in file '${sourceFrom.path}'`,
```
anton-trunov: Hey thanks for the fix! Here the error message would be a bit too long and contain the full path, let's make it shorter

anton-trunov (changes requested):

src/imports/resolveImports.ts
verytactical: Paths in error messages should be relative to `process.cwd()`

src/test/compilation-failed/import-errors.spec.ts
```diff
@@ -18,3 +18,16 @@ it("symlinks are not allowed", async () => {
...
+it("should fail on duplicate imports", async () => {
+    const result = await run({
+        config: createSingleFileConfig(`duplicate-import.tact`, "./output"),
+        logger: new Logger(LogLevel.NONE),
+        project: createNodeFileSystem(join(__dirname, "contracts")),
+        stdlib: createVirtualFileSystem("@stdlib", Stdlib.files),
+    });
+    expect(result.ok).toBe(false);
+    const message = result.error.map((err) => err.message).join("; ");
+    expect(message).toContain("Duplicate import of");
```
verytactical: These tests would be a bit annoying to update. We prefer using `toMatchSnapshot` tests in such situations.

For cases when we need to check that compilation of a contract fails with a specific message, there's a helper `itShouldNotCompile` (see `src/test/compilation-failed/const-eval-failed.spec.ts`). If interface of `run` changes, we'll otherwise have to update this code too.

PR #2961, opened 2025-04-28
Vitest implementation

### LINKED ISSUE #1102: Switch to a testing framework that is not Jest
Related to https://github.com/tact-lang/tact/issues/146

xpyctumo:
## Issue

Closes #1102.

🥇  No more Jest. Here [Vitest](https://vitest.dev/).

The available commands remain the same, but one new one has been added. About this below.

### Realtime test run in browser

Just `yarn test:ui` to run tests and watch it in browser. You can see the code of the tests in the browser. Once the tests are completed, a coverage report using `istanbul` will also be collected

image

### Displaying heap memory in tests

For each test, the heap memory it consumes after execution is now displayed. This makes it easier to find memory leaks in tests. 

image

### Perfomance

Group 1

At the moment it is supported to run both regular (without ignoring heavy tests) - `yarn test`, and lightweight - `yarn test:fast`.

You can see that their performance stays +- on the same level. Vitest can give more by adding the `isolate: false` parameter to the config, but this will add some flaky tests that were detected during local test runs. 

`isolate: false` in `yarn test:fast` gives about ~3 seconds gain and carries some flaky test (not always).

## Other 
- ~Right now, Jest is left to CI (since it collects coverage) using third-party tools like `@tact-lang/coverage`. Vitest doesn't really want to be friends with it.~ 
- `.toHaveTransaction` matcher from a third-party library is also supported
- Vitest offers its Snapshot, for that reason they have been updated in this PR as well


src/cli/logger.ts
```diff
@@ -5,7 +5,7 @@ import {
     showExpectedText,
     showTemplate,
 } from "@/error/string-util";
-import { cwd } from "process";
+import "process";
```
xpyctumo, line 8: remove this? Maybe it will work without this import

vitest.setup.ts
```diff
@@ -0,0 +1,121 @@
+import { afterAll, expect } from "vitest";
+import { compareTransactionForTest } from "@ton/test-utils/dist/test/transaction";
+import {
+    compareAddressForTest,
+    compareCellForTest,
+    compareSliceForTest,
+} from "@ton/test-utils/dist/test/comparisons";
+import coverage from "@tact-lang/coverage";
```
xpyctumo, line 8: no need

xpyctumo:
Postponed this until we find product need for Vitest

Issue #2940, opened 2025-04-25
Compiler accepts `map<Int, Int>` literals without storage annotations, contradicting documentation


Gusarich:
**Reproduction:**
```tact
contract C {
    receive() {
        // According to documentation, Int key/value in map literals must use storage annotations
        let m = map<Int, Int> { 1: 2 };
        cashback(sender());
    }
}
```

**Expected Behavior:**  
Compiler should reject this code with a clear error like:  
```
Error: map literal key and value of Int type must have storage annotations (e.g., `as uintX` or `as coins`)
```

**Actual Behavior:**  
The code compiles successfully without any warnings or errors.

**Why This Is Incorrect:**  
The Maps reference documentation explicitly states:  
> “Notice that keys and values of the Int type must be defined with storage annotations via `as` keyword.”

This behavior creates a mismatch between the documented language rules and the actual compiler enforcement, potentially misleading users and causing inconsistent serialization behavior.

---

LLM Fuzzing discovery (see #2490)

PR #2989, opened 2025-05-01
fix(docs): clarify that exactly `emptyMap` allows omitting type ascriptions

i582:
Towards #2940.

### DIFF
--- a/docs/src/content/docs/book/maps.mdx
+++ b/docs/src/content/docs/book/maps.mdx
@@ -142,7 +142,8 @@ let myMap: map<Int as uint8, Int as uint16> = map<Int as uint8, Int as uint16> {
 let mismatch: map<Address, Cell> = map<Address, Int> {}; // COMPILATION ERROR! Type mismatch
 ```
 
-Empty map literals are also supported. They produce the same value as the [`emptyMap(){:tact}`](#emptymap) function but allow omitting type ascriptions.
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/book/maps.mdx
```diff
@@ -142,7 +142,7 @@ let myMap: map<Int as uint8, Int as uint16> = map<Int as uint8, Int as uint16> {
 let mismatch: map<Address, Cell> = map<Address, Int> {}; // COMPILATION ERROR! Type mismatch
 ```
 
-Empty map literals are also supported. They produce the same value as the [`emptyMap(){:tact}`](#emptymap) function but allow omitting type ascriptions.
+Map literals with an empty body are also supported. They produce the same value as the [`emptyMap(){:tact}`](#emptymap) function and also require type ascriptions for the variable declaration.
```
anton-trunov, line 145: a code example is needed

anton-trunov (changes requested):

docs/src/content/docs/book/maps.mdx
novusnota: It's right below :)

```tact
let myMap: map<Int as uint8, Cell> = map<Int as uint8, Cell> {};
let myMap2: map<Int as uint8, Cell> = emptyMap();

myMap == myMap2; // true
```

anton-trunov (approved):

PR #2998, opened 2025-05-03
fix(stdlib): add missed link to exit code

i582:


anton-trunov (approved):

anton-trunov:
@i582 we should automate checks for links in stdlib

PR #3002, opened 2025-05-05
feat(docs): rework the optionals page

### LINKED ISSUE #2555: Nested Optionals (`T??`) Are Disallowed but Not Mentioned in the Docs
The compiler rejects nested optional types such as `Int??`, but the Tact documentation does **not** mention that this is disallowed.

The Optionals section states that "all primitive types, as well as Structs and Messages, can be made nullable and hold a special null value" — but it does **not** say that applying `?` more than once is invalid.

This leads to confusion for users who may reasonably expect `T??` to be valid or at least explicitly documented as unsupported.

### 🔁 Minimal Example

```tact
contract Test {
    fun nestedWait1(a: Int??) {
        // does nothing
    }

    receive() {
        self.nestedWait1(null);
    }
}
```

### 🧨 Compiler Output

```
Nested optional types are not allowed
```

### 📌 Why This Is a Problem

- The documentation says that **any** type can be made nullable — but doesn’t define or restrict how many times `?` can be applied.
- The compiler rejects a nested optional with a hard error and no prior warning in the docs.
- This is a clear **documentation mismatch** — the behavior is not reflected in the written specification.

### ✅ Suggested Fix

Add a note in the Optionals section explicitly stating that **nested optional types (e.g. `Int??`) are not allowed**, and only one layer of optionality is supported per type.

This will help prevent confusion and ensure the language’s limitations are clearly understood.

---

### LINKED ISSUE #2981: `Int? as coins` is accepted but undocumented
**Description:**  
The compiler accepts persistent fields like `Int? as coins`, but the documentation never mentions that serialization specifiers can be used with optional types.

**Minimal Repro:**  
```tact
contract OptSerTest {
    a: Int? as coins;

    init() {
        self.a = 10;
    }
}
```

**Why this is a problem:**  
The docs describe serialization (`as coins`, `as uint8`, etc.) only for non-optional primitives like `Int`, never for `Int?`. There are no examples, explanations, or serialization semantics for `null` values. This mismatch may confuse users and tooling about whether this syntax is valid, how it behaves, and whether it's safe to rely on. Either the docs should clarify this behavior or the compiler should reject it.

---

novusnota:
## Issue

Closes #2555 (Nested optionals `??` are disallowed)
Closes #2981 (`Int? as coins` and other serialization notes)

Rewrote the page and all descriptions such that they make sense and are easy to comprehend with examples following almost every paragraph.


docs/src/content/docs/book/optionals.mdx
```diff
@@ -2,41 +2,140 @@
...
+As mentioned in the [type system overview](/book/types#optionals), all [primitive types](/book/types#primitive-types), [structs][struct], and [message structs][message] can be made nullable. That is, [variables](/book/statements#let), [function parameters](/book/functions), [contract parameters](/book/contracts#parameters) and [structure fields][structure] of primitive or structural types can hold the special `null{:tact}` value that represents the intentional absence of any other value.
 
-[Variables](/book/statements#let) or fields of [Structs](/book/structs-and-messages#structs) and [Messages](/book/structs-and-messages#messages) that can hold `null{:tact}` are called "optionals". They're useful for reducing state size when the variable isn't necessarily used.
+Such data types that may or may not contain the `null{:tact}` value are called _optionals_. They are useful for [reducing the state size](#serialization) when the value is not necessarily used.
```
anton-trunov: That extra bit makes the stored value larger
```suggestion
Such data types that may or may not contain the `null{:tact}` value are called _optionals_.
```

anton-trunov (changes requested):

anton-trunov (dismissed):

docs/src/content/docs/book/optionals.mdx
```diff
@@ -2,41 +2,140 @@
 title: Optionals
```
novusnota, line 2:
```suggestion
title: Optionals
description: "Data types that can contain the null value in addition to values of their encapsulated primitive or structural types."
```

docs/src/content/docs/book/optionals.mdx
```diff
@@ -1,42 +1,142 @@
 ---
 title: Optionals
+description: "Data types that can contain the null value in addition to values of their encapsulated primitive or structural types."
```
novusnota:
```suggestion
description: "Data types that can contain the null value in addition to values of their encapsulated primitive or struct types."
```

docs/src/content/docs/book/optionals.mdx
```diff
@@ -1,42 +1,142 @@
...
 ---
 
-As mentioned in the [type system overview](/book/types#optionals), all [primitive types](/book/types#primitive-types), [structs](/book/structs-and-messages#structs), and [Messages](/book/structs-and-messages#messages) can be nullable. That is, they don't necessarily hold any value aside from `null{:tact}` — a special value that represents the intentional absence of any other value.
+As mentioned in the [type system overview](/book/types#optionals), all [primitive types](/book/types#primitive-types), [structs][struct], and [message structs][message] can be made nullable. That is, [variables](/book/statements#let), [function parameters](/book/functions), [contract parameters](/book/contracts#parameters) and [structure fields][structure] of primitive or structural types can hold the special `null{:tact}` value that represents the intentional absence of any other value.
```
novusnota:
```suggestion
As mentioned in the [type system overview](/book/types#optionals), all [primitive types](/book/types#primitive-types), [structs][struct], and [message structs][message] can be made nullable. That is, [variables](/book/statements#let), [function parameters](/book/functions), [contract parameters](/book/contracts#parameters) and [structure fields][structure] of primitive or struct types can hold the special `null{:tact}` value that represents the intentional absence of any other value.
```

docs/src/content/docs/book/optionals.mdx
```diff
@@ -1,42 +1,142 @@
...
-
-Optional variables or optional fields that are not defined hold the `null{:tact}` value by default. You cannot access them without checking for `null{:tact}` first. However, if you're certain they are not `null{:tact}` at a given moment, use the [non-null assertion operator `!!{:tact}`](/book/operators#unary-non-null-assert) to access their value.
-
-Attempts to access the value of an optional variable or an optional field without using [`!!{:tact}`](/book/operators#unary-non-null-assert) or without checking for `null{:tact}` beforehand will result in a compilation error if the compiler can track it, or, if it cannot, in an exception with [exit code 128](/book/exit-codes#128): `Null reference exception`.
-
-Example of optionals:
+You can make a primitive or structural type into an optional by adding a question mark `?{:tact}` after its type declaration.
```
novusnota:
```suggestion
You can make a primitive or struct type into an optional by adding a question mark `?{:tact}` after its type declaration.
```

docs/src/content/docs/book/optionals.mdx
```diff
@@ -1,42 +1,142 @@
...
+        // Explicit check of the message struct field
+        if (msg.opt != null) {
+            // Non-null assertion to work with its inner value
+            self.reset(msg.opt!!.opt);
         }
     }
 }
 ```
+
+Since [`map<K, V>{:tact}`](/book/maps) and [`bounced<Msg>{:tact}`](/book/bounced) are not primitive or structural types, they cannot be made optional. Furthermore, their inner key-value types (in the case of a map) and the inner [message struct][message] (in the case of a bounced constructor) cannot be optional too.
```
novusnota:
```suggestion
Since [`map<K, V>{:tact}`](/book/maps) and [`bounced<Msg>{:tact}`](/book/bounced) are not primitive or struct types, they cannot be made optional. Furthermore, their inner key-value types (in the case of a map) and the inner [message struct][message] (in the case of a bounced constructor) cannot be optional too.
```

docs/src/content/docs/book/optionals.mdx
```diff
@@ -1,42 +1,142 @@
...
+
+```tact
+// COMPILATION ERROR! Map key types cannot be optional
+let myMap: map<Int?, Int> = emptyMap();
+//             ~~~~
+```
+
+Creating a nested optional type by adding multiple question marks `?{:tact}` is not allowed, as optionals are neither a primitive nor a structural type.
```
novusnota:
```suggestion
Creating a nested optional type by adding multiple question marks `?{:tact}` is not allowed, as optionals are neither a primitive nor a struct type.
```

docs/src/content/docs/book/optionals.mdx
```diff
@@ -1,42 +1,142 @@
...
+Deserialization works inversely. First, a single bit is loaded from a [`Slice{:tact}`][slice]. If it is 0, the value is read as `null{:tact}`. If it is 1, then the value is loaded from the following data bits.
+
+```tact
+struct Wasp {
+    hasSting: Bool?, // 2 bits max: 1 for the optional, 0 or 1 for the boolean
+    stingLength: Int?, // 258 bits max: 1 for the optional, 0 or 257 for the integer
+}
+```
+
+Optionals can have serialization annotations provided after the `as{:tact}` keyword and have all the same serialization options as their encapsulated primitive or structural types. As such, the optional [`Int?{:tact}`](/book/integers) type has the most serialization formats available.
```
novusnota:
```suggestion
Optionals can have serialization annotations provided after the `as{:tact}` keyword and have all the same serialization options as their encapsulated primitive or struct types. As such, the optional [`Int?{:tact}`](/book/integers) type has the most serialization formats available.
```

anton-trunov (approved):

PR #3056, opened 2025-05-14
feat: add .send() methods to message parameter structs

### LINKED ISSUE #3055: We should add extends send function to SendParameters / DeployParameters / MessageParameters
It should look like this:
```
SendParameters{}.send()
DeployParameters{}.send()
MessageParameters{}.send()
```

skywardboundd:
Closes #3055


docs/src/content/docs/book/contracts.mdx
```diff
@@ -481,12 +481,12 @@ contract HelloWorld {
...
 
         // sending the message back to the sender
-        send(SendParameters {
+        SendParameters {
             to: sender(),
             value: 0,
             mode: SendRemainingValue | SendIgnoreErrors,
             body: CanBounce { counter: self.counter }.toCell(),
-        });
+        }.send();
```
novusnota: Cool, but I'm not sure we want to replace all existing usages with this new approach. To be frank, I'd prefer to wait for / implement `send {}` as a statement and stop with our workarounds upon workarounds :)

wdyt @anton-trunov @verytactical?

src/stdlib/stdlib/std/internal/send.tact
```diff
@@ -441,13 +721,13 @@ asm fun deploy(params: DeployParameters) {
...
 ///     // The cashback() function above is cheaper, but functionally
 ///     // equivalent to the following call to the message() function
-///     message(MessageParameters{
+///     MessageParameters{
 ///         mode: SendRemainingValue | SendIgnoreErrors,
 ///         body: null,
 ///         value: 0,
 ///         to: sender(),
 ///         bounce: false,
-///     });
```
Shvandre, line 450: Missing ; after substitution

src/stdlib/stdlib/std/internal/send.tact
```diff
@@ -129,11 +129,11 @@ struct MessageParameters {
 ///
 /// ```tact
 /// fun example() {
-///     message(MessageParameters{
+///     MessageParameters{
 ///         to: sender(),    // back to the sender,
 ///         value: ton("1"), // with 1 Toncoin (1_000_000_000 nanoToncoin),
 ///                          // and no message body
-///     });
+///     }.send()
```
novusnota:
```suggestion
///     }.send();
```

src/stdlib/stdlib/std/internal/send.tact
```diff
@@ -167,15 +167,61 @@ asm fun message(params: MessageParameters) {
     SENDRAWMSG
 }
 
+/// Available since Tact 1.7.0.
```
novusnota:
```suggestion
/// Extension function for the `MessageParameters` struct. Available since Tact 1.7.0.
```

src/stdlib/stdlib/std/internal/send.tact
```diff
@@ -281,6 +327,120 @@ asm fun send(params: SendParameters) {
     SENDRAWMSG // https://github.com/tact-lang/tact/issues/1558
 }
 
+/// Queues the message to be sent using a `SendParameters` struct.
```
novusnota, line 316:
```suggestion
/// Extension function for the `SendParameters` struct. Available since Tact 1.7.0.
///
/// Queues the message to be sent using a `SendParameters` struct.
```

src/stdlib/stdlib/std/internal/send.tact
```diff
@@ -427,6 +587,126 @@ asm fun deploy(params: DeployParameters) {
     SENDRAWMSG
 }
 
+/// Available since Tact 1.7.0.
```
novusnota:
```suggestion
/// Extension function for the `DeployParameters` struct. Available since Tact 1.7.0.
```

src/stdlib/stdlib/std/internal/send.tact
```diff
@@ -441,13 +721,13 @@ asm fun deploy(params: DeployParameters) {
...
 ///     // equivalent to the following call to the message() function
-///     message(MessageParameters{
+///     MessageParameters{
 ///         mode: SendRemainingValue | SendIgnoreErrors,
 ///         body: null,
 ///         value: 0,
 ///         to: sender(),
 ///         bounce: false,
-///     });
+///     }.send()
```
novusnota:
```suggestion
///     }.send();
```

src/stdlib/stdlib/std/internal/send.tact
```diff
@@ -322,12 +482,12 @@ struct DeployParameters {
...
 /// ```tact
 /// fun example() {
-///     deploy(DeployParameters{
+///     DeployParameters{
 ///         init: initOf SomeContract(), // with initial code and data of SomeContract
 ///                                      // and no additional message body
 ///         mode: SendIgnoreErrors,      // skip the message in case of errors
 ///         value: ton("1"),             // send 1 Toncoin (1_000_000_000 nanoToncoin)
-///     });
+///     }.send()
```
novusnota:
```suggestion
///     }.send();
```

novusnota (changes requested):
And the changes to `core-send.mdx` have to be reversed + 3 new send functions added there. One moment.

UPD: Done, but I'm still not in favor of changing every usage example in the docs.

novusnota (approved):

novusnota (changes requested):
Hold on, let's revert changes to the documentation in all places but the `core-send`.

I've previously vaguely expressed my concerns [here](https://github.com/tact-lang/tact/pull/3056#discussion_r2091063049), but now I am certain — we must not replace older examples with the new code that won't work for users until 1.7 is released, because docs here will be published immediately on merge, unlike Tact 1.7. Furthermore, not all users are on the latest version, and we still have many folks that use versions prior to 1.6.0.

novusnota (approved):
Sehr gut.

novusnota (commented):
To be actually semantically closer to what's going on with messages under the hood, we might want to rename the `.send()` to the `.queue()`.

And in the end, perhaps we're looking at having a single `msg {}` statement instead of all those functions. That statement will compose the message declaratively, and then we'll be able to call `.queue()` on the resulting message cell, specifying the mode of sending.

PR #3060, opened 2025-05-15
chore: export all assembly related definitions

i582:
At least for Jetton coverage

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -40,3 +40,10 @@ export * from "@/imports/path";
 export * from "@/grammar";
 export * from "@/grammar/src-info";
 export * from "@/optimizer/interpreter";
+
+export * from "@/asm/coverage";
+export * from "@/asm/helpers";
+export * from "@/asm/logs";
+export * from "@/asm/runtime";
+export * as asmText from "@/asm/text";
+export * from "@/asm/trace";


src/index.ts
```diff
@@ -40,3 +40,10 @@ export * from "@/imports/path";
 export * from "@/grammar";
 export * from "@/grammar/src-info";
 export * from "@/optimizer/interpreter";
+
+export * from "@/asm/coverage";
+export * from "@/asm/helpers";
+export * from "@/asm/logs";
+export * from "@/asm/runtime";
+export * as asmText from "@/asm/text";
```
reverofevil, line 48: why do we need to export it like this?

i582: both text and runtime export symbols like `PUSHNULL` with the same name, so without this we'll get a bunch of errors

i582:
```
src/index.ts:48:1 - error TS2308: Module "@/asm/runtime" has already exported a member named 'XLOADQ'. Consider explicitly re-exporting to resolve the ambiguity.

48 export * from "@/asm/text";
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/index.ts:48:1 - error TS2308: Module "@/asm/runtime" has already exported a member named 'XOR'. Consider explicitly re-exporting to resolve the ambiguity.

48 export * from "@/asm/text";
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

anton-trunov (approved):

PR #3067, opened 2025-05-15
feat: add hard deprecate error for VarAddress and parseVarAddress

### LINKED ISSUE #3048: Rework or remove everything related to `VarAddress` (`addr_var`) from the standard library
In [Version 10](https://github.com/ton-blockchain/ton/blob/master/doc/GlobalVersions.md#version-10) of the TON Blockchain, support of `addr_var` addresses was almost completely cut down. For example, the `addr_var` cannot be a destination address of external messages, and if it's used for a destination address of an internal message, it is rewritten as an `addr_std`.

Furthermore, these TVM instructions drop support for anycast addresses and `addr_var`: `LDMSGADDR(Q)`, `PARSEMSGADDR(Q)`, `REWRITESTDADDR(Q)`, `REWRITEVARADDR(Q)`. They throw an exit code (usually 9) whenever one attempts to apply them to anything, including `addr_var` addresses.

That hits our `parseVarAddress()` function and the respective `VarAddress` struct. We should either remove those entirely or deprecate/turn into a no-op.

i582:
## Issue

Closes #3048.


anton-trunov (changes requested):
Let's add some tests featuring the deprecation errors for each stdlib entry we are deprecating

i582:
> Let's add some tests featuring the deprecation errors for each stdlib entry we are deprecating

Added

anton-trunov (changes requested):
Let's add an entry to the changelog

i582:
> Let's add an entry to the changelog

Done

anton-trunov (approved):

PR #3076, opened 2025-05-16
feat(docs): revamp of the functions page

### LINKED ISSUE #888: Revamp of Functions page and related pages
They need:
* Better formatting
* More thorough description of attributes (`extends`, `mutates`, etc.)
* Mention of `receive()`, `external()`, `bounced()` with links to relevant pages
* Consistent and cohesive descriptions
* Better cross-linkage
* ...

Separate receivers and getters from all other functions — make two groups. Or three, to put getters in their own category.

### LINKED ISSUE #3175: Use banner for warning instead of quote
https://docs.tact-lang.org/book/functions/#extension-function

Image

Image

### LINKED ISSUE #3177: "Getter Functions" don't explain their purpose
https://docs.tact-lang.org/book/functions/#getter-functions

### LINKED ISSUE #3178: Make all items in a Function list links
A bit confusing right now

https://docs.tact-lang.org/book/functions/#getter-functions

Image

### LINKED ISSUE #3183: Link "Receiver functions" description with communication section
https://docs.tact-lang.org/book/functions/#receiver-functions

Image

novusnota:
## Issue

Closes #888. Also does some minor adjustments to related pages.
Closes #3175.
Closes #3177.
Closes #3183.
Closes #3178 ...and lots more.

This PR finally makes a distinction between the most important functions (such as receivers and getters) with the rest of them. There are follow-up issues for receivers and various function kinds, so this PR focuses on getting the essentials and the foundations done well.


docs/src/content/docs/book/functions.mdx
```diff
@@ -50,7 +44,189 @@ contract YouMakeMyHeartSing with WildThing {
...
+contract E() {
+    external("hello, it's me") {
+        // Can't be replied to as there's no sender!
+    }
+}
+```
+
+Read more about them on their dedicated page: [External messages](/book/external).
+
+### `get fun` — offchain getter functions {#get}
```
anton-trunov: let's also add:
- getters can modify the contract state variables, but the modification is not-persistent;
- need an example of a getter with parameters;
- getters do not consume gas from the contract's balance;
- there is a gas limit for getters (@Shvandre mentioned it depends on an API provider);
- `method_id` is not only useful for collision resolution but also to ensure the correct interface (the case of Jettons);
- use different examples: borrow from Jettons or our benchmarks;
- we can provide an example of returning the whole state of the contract without listing all the contract fields;
- getters cannot be `asm` functions;
- getters can be used as contract methods: for instance, receivers can call the methods of the contract they belong to;
- check issue #2573, if it has more things to cover.

anton-trunov (changes requested):

docs/src/content/docs/book/assembly-functions.mdx
anton-trunov, line 1: With all this URL renaming, would old links still work?

docs/src/content/docs/book/contracts.mdx
```diff
@@ -514,13 +514,13 @@ contract Example {
 }
 ```
 
-### Internal functions
+### Member functions {#internal-functions}
```
anton-trunov: These are private contract functions (or methods). You can define public contract function on the top-level.

anton-trunov (changes requested):
Please revert all the internal -> member renames: it makes it hard to review the essential changes.

docs/src/content/docs/book/functions.mdx
```diff
@@ -1,22 +1,16 @@
...
-* Global static functions
-* Extension functions
-* Mutation functions
-* Native functions
-* [Assembly functions](#asm)
-* [Internal functions](/book/contracts#internal-functions)
-* Receiver functions
-* Getter functions
+* [Primary](#primary), which constitute the backbone and the API of the contract
+* [Complementary](#complementary), which serve as helper building blocks for reusable and maintainable contracts.
```
anton-trunov: please do not make this classification one more level deep, let's keep the previous one

anton-trunov (changes requested):

docs/src/content/docs/book/functions.mdx
```diff
@@ -50,7 +44,189 @@ contract YouMakeMyHeartSing with WildThing {
...
+### `receive` — internal message receivers {#receive}
+
+<Badge text="Contract scope" variant="note" size="medium"/><p/>
+
+Receiver functions are special functions responsible for receiving messages in contracts and can be defined only within a contract or trait.
+
+```tact
+contract Counter(counter: Int) {
+    // This means that this contract can receive the comment "Increment", and this function would be called for such messages
+    receive("Increment") {
```
anton-trunov: do not use text receivers

docs/src/content/docs/book/functions.mdx
```diff
@@ -50,7 +44,189 @@ contract YouMakeMyHeartSing with WildThing {
...
+
+### `external` — external message receivers {#external}
+
+<Badge text="Contract scope" variant="note" size="medium"/><p/>
+
+The `external(){:tact}` is a special kind of receivers which handle external messages — they are sent from the off-chain world and do not have a sender address on the blockchain. Such messages are often sent to wallet contracts to process specific messages or simply to send funds to another wallet contract.
+
+```tact
+contract E() {
+    external("hello, it's me") {
```
anton-trunov: not text receivers

docs/src/content/docs/book/functions.mdx
```diff
@@ -64,33 +240,96 @@ fun customPow(a: Int, c: Int): Int {
...
-contract Filter with FilterTrait {
-    // Overriding the default behavior of the FilterTrait
-    override fun filterMessage(): Bool {
-        return true;
-    }
+They can only be called from [receivers](#receive), [getters](#get), and other member functions, but not from other contracts.
 
-    override fun specialFilter(): Bool {
-        return true;
+### `init` — constructor function {#init}
```
anton-trunov: To better explain `init`, add an example with contract parameters (including an `initialized: Bool` flag) and show how `init` can be simulated. This can be done using the Counter contract and its two equivalent implementations using contract fields and contract parameters.

docs/src/content/docs/book/contracts.mdx
```diff
@@ -423,17 +423,17 @@ contract LazyInit {
...
 
 :::
 
 ### Getter functions
 
-[Getter functions](/book/functions#getter-functions) are **not accessible from other contracts and are exported only to the off-chain world**.
+[Getter functions](/book/functions#get) are **not accessible** from other contracts and are exported only to the off-chain world.
 
-Additionally, **getters cannot modify the contract's state variables**; they can only read their values and use them in expressions.
+Additionally, getters **cannot modify the contract's state variables** — they can only read their values and use them in expressions.
```
Shvandre: They actually can. Moreover, we can call getters directly in Tact.
```tact
contract Test (var: Int) {
    receive() {
        self.modify();
    }

    get fun modify() {
        self.var = 1;
    }
}
```
Maybe it's better to say
> getters **cannot modify the contract's state variables** when called offchain?

still sounds confusing, idk

docs/src/content/docs/book/functions.mdx
```diff
@@ -50,7 +44,189 @@ contract YouMakeMyHeartSing with WildThing {
...
+    get(crc32("crc32") + 42 & 0x3ffff | 0x4000)
+    fun methodId2(): Int {
+        return crc32("crc32") + 42 & 0x3ffff | 0x4000;
+    }
+}
+```
+
+Unlike getters, method IDs for [member functions](/book/contracts#internal-functions) and some special functions are obtained sequentially: integers in the inclusive range from -4 to 0 are given to [certain message handlers](https://docs.ton.org/v3/documentation/smart-contracts/func/docs/functions#special-function-names), while member functions are numbered with method IDs starting at $1$ and going up to $2^{14} - 1$ inclusive.
+
+Since method IDs are 19-bit signed integers and some of them are reserved, only the inclusive ranges from $-2^{18}$ to $-5$ and from $2^{14}$ to $2^{18} - 1$ are free to be used by users. To avoid collisions, it is recommended to specify method IDs only in these ranges, avoiding the method IDs of Tact-specific getters mentioned above.
```
Shvandre: If I understand correctly, according to algorithm above, get methods ids can't be negative, so specifying negative ids explicitly looks kinda confusing

docs/src/content/docs/book/functions.mdx
```diff
@@ -121,82 +366,124 @@ extends mutates fun customPow(self: Int, c: Int) {
...
+<Badge text="Contract scope" variant="note" size="medium"/><p/>
+
+The `get{:tact}` attribute is a special attribute that cannot be combined with any other attribute. It is applied to [member functions], transforming them into [getter functions](#get).
+
+## Low-level representation
 
-Since method IDs are $19$-bit signed integers and some of them are reserved, only the inclusive ranges from $-2^{18}$ to $-5$ and from $2^{14}$ to $2^{18} - 1$ are free to be used by users. To avoid collisions, it is recommended to specify method IDs only in these ranges, avoiding the method IDs of Tact-specific getters mentioned above.
+On [TVM][tvm] level, contracts and their functions are often represented as a dictionary (map) of methods, from their numeric method IDs to their respective bodies composed of TVM instructions. Each function then is a method that transforms the stack and TVM registers.
+
+A function call is really a call into that dictionary under that functions ID, after which:
```
Shvandre: This sentence sounds overcomplicated, what about something like:

`A function call is essentially a call into the dictionary using the function’s ID`

Shvandre (dismissed):
Good job! 
All my comments are discussible, so here in an approval from me. Feel free to resolve my comments if you think it's better to leave it as it is

docs/src/content/docs/book/assembly-functions.mdx
novusnota: They'll lead to the same page, at the very least

docs/src/content/docs/book/functions.mdx
novusnota: Hmm, but that's kinda the point — to not mix all kinds of things together and have a clear primary-complementary separation.

image

Do you insist? :)

anton-trunov: I do. I don't get this separation at all

novusnota: Ok

docs/src/content/docs/book/contracts.mdx
novusnota: Yeah, they'll be explained on the functions page in full, so I'll remove this sentence entirely and only keep the brief note about off-chain and link directly to the functions page.

docs/src/content/docs/book/gas-best-practices.mdx
```diff
@@ -225,7 +225,7 @@ contract Example() {
 }
 ```
 
-### Avoid internal contract functions
+### Avoid contract internal functions
```
anton-trunov:
```suggestion
```

docs/src/content/docs/book/functions.mdx
novusnota: Method IDs can be negative if specified directly. This is a valid contract:

```tact
contract Example() {
    get(-100) fun antipositive(): Int { return -100 }
}
```

And in its respective Fift code there is the following declaration:

```fift
PROGRAM{
   // ...
  -100 DECLMETHOD %antipositive
  // ...
```

So yeah, perhaps a note about algorithm vs. manual is needed.

docs/src/content/docs/book/gas-best-practices.mdx
novusnota: I'll just restore the previous heading, `### Avoid internal contract functions`

novusnota:
Current state of affairs (pretty good):

image

Finishing the bulk of the work on getters and resolving a few remaining TODOs. Should be ready for review very soon.

docs/src/content/docs/book/functions.mdx
novusnota: Done! This PR is already huge, so if there are any getter-related spots that I've missed, let's cover them in separate, smaller issue-PR pairs.

novusnota:
Ready for review.

image

.github/workflows/tact-docs-test.yml
```diff
@@ -70,6 +70,8 @@ jobs:
           output: "/dev/stdout"
           fail: true
           failIfEmpty: false
+        env:
```
anton-trunov, line 73: why this change?

docs/src/content/docs/book/assembly-functions.mdx
```diff
@@ -398,17 +398,17 @@ asm fun myCode(): Cell { MYCODE }
 
 ### `mutates` consumes an extra value {#caveats-mutates}
 
-Specifying the [`mutates{:tact}`](/book/functions#mutation-functions) attribute, i.e. defining a mutation function, makes the assembly function consume one additional value deeper in the stack than the declared return values. Consider the following example:
+Specifying the [`mutates{:tact}`](/book/functions#attributes-mutates) attribute, i.e. defining a mutation function, makes the assembly function consume one additional value deeper in the stack than the declared return values. Consider the following example:
```
anton-trunov, line 401: mutating methods should not be mentioned in the attributes section, they should be mentioned in the same list as `get fun` together with non-mutating methods

anton-trunov (changes requested):
Let's fix the general structure first (add mutating and non-mutating methods in the right place of the table of contents) and then continue with reviewing

.github/workflows/tact-docs-test.yml
novusnota: To increase limits for lychee checks of GitHub links.

docs/src/content/docs/book/assembly-functions.mdx
novusnota: Done

docs/src/content/docs/book/functions.mdx
```diff
@@ -1,184 +1,463 @@
...
+These functions behave similarly to private methods in popular object-oriented languages — they are internal to contracts and can be called by prefixing their names with a special [identifier `self{:tact}`](/book/contracts#self). That is why internal (or member) functions are sometimes referred to as "contract methods".
+
+Internal functions can access the contract's [persistent state variables](/book/contracts#variables) and [constants](/book/contracts#constants).
+
+They can only be called from [receivers](#receivers), [getters](#get), and other internal functions, but not from other contracts.
+
+```tact
+contract InternalFun() {
+    fun add(a: Int, b: Int): Int {
+        return a + b;
```
anton-trunov: let's make this function use a contract parameter, so `self` actually makes sense below

docs/src/content/docs/book/functions.mdx
```diff
@@ -1,184 +1,463 @@
...
+<Badge text="Global scope" variant="note" size="medium"/><p/>
+
+Regular functions that can be defined only at the top (module) level are called global. They differ from [internal functions](#fun-internal) not only in the scope, but also in available [attributes](#attributes).
 
 ```tact
-trait FilterTrait with Ownable {
-    // Virtual functions can be overridden by users of this trait
-    virtual fun filterMessage(): Bool {
-        return sender() != self.owner;
+fun customPow(a: Int, c: Int): Int {
```
anton-trunov: let's provide a more domain specific example, or at least less naive

docs/src/content/docs/book/functions.mdx
```diff
@@ -1,184 +1,463 @@
...
-> The first argument MUST be named `self`, and the argument's type must be the type you are extending.
+## `extends` — extension functions {#extends}
+
+<Badge text="Global scope" variant="note" size="medium"/><p/>
+
+To define an extension function, add an `extends{:tact}` attribute to any [global function](#fun-global) and name its first parameter as `self`. The type of `self` is the type that is being extended by the extension function.
 
 ```tact
-extends fun customPow(self: Int, c: Int): Int {
+extends fun customPow(self: Int, exp: Int): Int {
```
anton-trunov: better example is needed

docs/src/content/docs/book/functions.mdx
```diff
@@ -1,184 +1,463 @@
...
-Mutation functions perform mutation on a value by replacing it with an execution result. To perform mutation, the function must change the `self` value.
+<Badge text="Global scope" variant="note" size="medium"/><p/>
+
+The extension _mutation_ functions replace the value of the extended type with the new one at the end of its execution. That value would change only upon the assignment to `self{:tact}` within the function body.
+
+To define an extension mutation function, add a `mutates{:tact}` attribute to the existing [extension function](#extends).
 
 ```tact
-extends mutates fun customPow(self: Int, c: Int) {
+extends mutates fun customPow(self: Int, exp: Int) {
```
anton-trunov: better example

docs/src/content/docs/book/functions.mdx
```diff
@@ -193,10 +472,544 @@ contract ManualMethodId {
...
+
+    get fun counter(): Int {
+        return self.counter;
+    }
+}
+
+contract CounterDIYInit(
+    // Persistent state variables
+    initialized: Bool, // set this field to `false` during deployment
+    counter: Int as uint32, // set this field to `0` during deployment
```
anton-trunov:
```suggestion
    counter: Int as uint32, // during deployment the value of this field does not matter
```

docs/src/content/docs/book/functions.mdx
```diff
@@ -193,10 +472,544 @@ contract ManualMethodId {
...
+) {
+    // Internal function, which will be used in place of a pseudo-constructor function
+    fun init() {
```
anton-trunov: let's use some other name: this should also fix syntax highlighting and make it more obvious it's not a builtin construction

docs/src/content/docs/book/functions.mdx
```diff
@@ -193,10 +472,544 @@ contract ManualMethodId {
...
+
+Native functions are Tact's [FFI](https://en.wikipedia.org/wiki/Foreign_function_interface) to FunC — they allow direct bindings to imported FunC functions. In order to declare them, you have to import a `.fc` or `.func` file first. Granted, the FunC's [standard library file, `stdlib.fc`](https://github.com/tact-lang/tact/blob/main/src/stdlib/stdlib/std/stdlib.fc), is always imported for you.
+
+Consider the following FunC function:
+
+```func title="import_me.fc"
+;; Notice the tensor return type (int, int),
+;; which means we would have to define at least one
+;; Tact structure to bind to this function.
+(int, int) minmax(int a, int b) {
```
anton-trunov: this is not a very good implementation of a 2-element sorting function: most likely, FunC won't optimize it to that single TVM which it corresponds to

i582:
Maybe we can already use the correct word here (see https://github.com/tact-lang/tact/issues/3229)

docs/src/content/docs/book/functions.mdx
anton-trunov: and since it can be implemented as an `asm` function in Tact, a better example would be to use some function from the `mathlib.fc` library (something about fixed-point arithmetic, for instance)

docs/src/content/docs/book/functions.mdx
```diff
@@ -193,10 +472,544 @@ contract ManualMethodId {
...
+* [`inline{:tact}`](#attributes-inline) — embeds the function contents at the call site.
+* [`extends{:tact}`](#attributes-extends) — makes it an [extension function](#extends).
+* [`mutates{:tact}`](#attributes-mutates), along with [`extends{:tact}`](#attributes-extends) — makes it an [extension mutation function](#mutates).
+
+These attributes _cannot_ be specified:
+
+* [`abstract{:tact}`](#attributes-abstract), [`virtual{:tact}`](#attributes-virtual) and [`override{:tact}`](#attributes-override) — native functions cannot be defined within a contract or a trait.
+* [`get{:tact}`](#attributes-get) — native functions cannot be [getters](#get).
+
+## `asm` — assembly functions {#asm}
```
anton-trunov, line 661: a simple example of an `asm` function is needed here, just to illustrate what it looks like

docs/src/content/docs/book/functions.mdx
```diff
@@ -193,10 +472,544 @@ contract ManualMethodId {
...
+        return true;
+    }
+
+    override fun specialFilter(): Bool {
+        return true;
+    }
+}
+```
+
+{/* TODO: inheritance table as per https://github.com/tact-lang/tact/issues/3030 */}
```
anton-trunov, line 771: this comment is visible in the rendered text

docs/src/content/docs/book/functions.mdx
```diff
@@ -193,10 +472,544 @@ contract ManualMethodId {
...
+```tact
+trait BlackSquare {
+    abstract fun paint();
+}
+
+trait CompositionVII {
+    abstract fun paintedWithOil(preliminary: Bool): Bool;
+}
+
+contract Abstractionism() with BlackSquare, CompositionVII {
```
anton-trunov: I appreciate the cultural references, but let's not do it here

we are building a DSL to handle finances, so let's keep our examples practical and definitely no string here (they are very gas expensive)

image

docs/src/content/docs/book/functions.mdx
novusnota: No, it's not — the `{/* ... */}` are `.mdx` comments, and we have a couple of those ones on some other pages

docs/src/content/docs/book/functions.mdx
```diff
@@ -193,10 +472,544 @@ contract ManualMethodId {
...
+
+Therefore, there are no modifications of the original values in different scopes, except for the [extension _mutation_ functions](#attributes-mutates). And even so, those functions do not mutate the original value denoted by the `self{:tact}` keyword — instead, they discard the old value and replace it with the new one obtained at the end of the function body.
+
+The above applies to all functions except for [receivers](#receivers), because they cannot be called directly and are only executed upon receiving specific messages.
+
+### Scope {#commonalities-scope}
+
+All kinds of functions vary in scope: some can only be defined at the top level, such as [global functions](#fun-global), while others can only be defined within contracts and traits, such as [internal functions](#fun-internal) or [receivers](#receivers).
+
+Functions are hoisted upon declaration, such that the ones declared earlier might call those declared later and vice versa.
```
anton-trunov, line 904: Two more clarifications are needed:
- static functions or internal contract functions cannot share identifiers
- extension functions can have the same identifiers, provided the types of the `self` parameters are different

docs/src/content/docs/book/functions.mdx
```diff
@@ -193,10 +472,544 @@ contract ManualMethodId {
...
+// and call to the first() function works as expected.
+fun second(flag: Bool): Bool {
+    if (flag) {
+        return true;
+    }
+    return first(); // always true, because of the prior condition
+}
+```
+
+### Return values {#commonalities-return}
```
anton-trunov, line 947: Link to the explanation that it should be obvious to the compiler that returns are always reachable

docs/src/content/docs/book/functions.mdx
novusnota: +1:
• names of internals and getters do not conflict with contract variables/fields. Or rather that they are in different namespaces, so "you" can name them the same for convenience, especially in the getters' case.

docs/src/content/docs/book/functions.mdx
```diff
@@ -193,10 +472,544 @@ contract ManualMethodId {
...
+fun example(someA: Int, someB: Int) {
+    // This sum will always be 2, regardless of values of `someA` and `someB`.
+    // The only exception is when their values are too big and there is not
+    // enough gas to finish the computations, in which case the transaction
+    // will fail with an exit code -14: Out of gas error.
+    produceOneFrom(someA) + alsoProduceOneFrom(someB); // 2
+}
+```
+
+### Trailing comma {#commonalities-comma}
```
anton-trunov, line 1008: and trailing semi-colon

novusnota: Oh, yeah

docs/src/content/docs/book/functions.mdx
novusnota: I've browsed through the `mathlib.fc`, and things written there can be done in Tact via regular or `asm` functions. However, there are things that are simply cannot be expressed in Tact yet, such as parametric polymorphism with `forall`.

```func
forall X -> (X, ()) ~impure_touch(X x) impure asm "NOP";
```

UPD: I totally forgot that the minmax example showed how to use structures. Searching for a suitable example...
UPD2: It's very tough to find a simple example that returns multiple values without resorting to asm. I actually want to keep the `minmax()` here, if possible.

Ok, I went with an asm function with HASHEXT_KECCAK512, just to show the mapping of Tact structures to FunC tuples.

anton-trunov (approved):

PR #3077, opened 2025-05-16
chore: use official WASM build v2025.04

i582:
Build from:
https://github.com/ton-blockchain/ton/releases/tag/v2025.04


anton-trunov (changes requested):
Let's add `SDBEGIN` / `SDBEGINQ` tests to make sure `Asm.fif` has been indeed fixed.

anton-trunov (approved):
It's not possible to control how BoCs are formed from Tact, so not adding tests

PR #3080, opened 2025-05-16
feat(docs): make low-level map representation more apparent

### LINKED ISSUE #3029: Make it obvious how maps are represented at TVM level
It's hidden here:
Image

novusnota:
## Issue

Closes #3029.


docs/src/content/docs/book/maps.mdx
```diff
@@ -528,9 +536,7 @@ Here, both maps are formed manually and both contain the same key-value pair. If
 extends fun asCell(self: map<K, V>): Cell?;
 ```
 
-On [TVM][tvm], maps are represented as a [`Cell{:tact}`][cell] type, and it's possible to construct and parse them directly. However, doing so is highly error-prone and quite messy, which is why Tact provides maps as a standalone composite type with many of the helper methods mentioned above.
-
-To cast maps back to the underlying [`Cell{:tact}`][cell] type, use the `.asCell(){:tact}` [method](/book/functions#extension-function). Since maps are initialized to `null{:tact}`, calling `.asCell(){:tact}` on a map with no values assigned will return `null{:tact}` and **not** an empty [`Cell{:tact}`][cell].
+To cast maps back to the [underlying](#low-level-representation) [`Cell{:tact}`][cell] type, use the `.asCell(){:tact}` [method](/book/functions#extension-function). Since maps are initialized to `null{:tact}`, calling `.asCell(){:tact}` on a map with no values assigned will return `null{:tact}` and **not** an empty [`Cell{:tact}`][cell].
```
anton-trunov, line 539: Let's add a test that demonstrates this statement

novusnota: Adjusted existing test suite. Also bumped allocation limits for Node.js & tsc, such that the latter is able to compile testsuite without failures on my machine (and in CI).

anton-trunov (approved):

PR #3119, opened 2025-05-18
fix: check static/instance contexts for method calls

### LINKED ISSUE #2952: Instance methods should be rejected if used as static methods
The following fails at the FunC stage, but should be rejected by the typechecker

```
message Hello {
    name: String
}

contract Test {
    receive() {
        message(MessageParameters{
            to: sender(),
            value: 0,
            mode: SendRemainingValue,
            body: Hello.toCell(), // `toCell` is mistakenly used as a static method here
        });
    }
}
```

Uncovered by @Gosunov

hazyone:
## Issue

Closes #2952.

## Checklist

### DIFF
--- a/src/abi/AbiFunction.ts
+++ b/src/abi/AbiFunction.ts
@@ -6,6 +6,7 @@ import type { SrcInfo } from "@/grammar";
 
 export type AbiFunction = {
     name: string;
+    isStatic?: boolean;
     resolve: (
         ctx: CompilerContext,
         args: readonly TypeRef[],
--- a/src/types/resolveExpression.ts
+++ b/src/types/resolveExpression.ts
@@ -682,7 +682,8 @@ function resolveCall(
             if (StructFunctions.has(idText(exp.method))) {
                 const abi = StructFunctions.get(idText(exp.method))!;
                 const isInstanceCall = exp.self.kind !== "id";
-                const isStaticCallFromType = exp.self.kind === "id" && exp.self.text === src.name;
+                const isStaticCallFromType =
+                    exp.self.kind === "id" && exp.self.text === src.name;
 
                 if (abi.isStatic && isInstanceCall) {
                     throwCompilationError(
--- a/src/types/resolveExpression.ts
+++ b/src/types/resolveExpression.ts
@@ -681,6 +681,21 @@ function resolveCall(
         if (srcT.kind === "struct") {
             if (StructFunctions.has(idText(exp.method))) {
                 const abi = StructFunctions.get(idText(exp.method))!;
+                const isInstanceCall = exp.self.kind !== "id";
+                const isStaticCallFromType = exp.self.kind === "id" && exp.self.text === src.name;
+
+                if (abi.isStatic && isInstanceCall) {
+                    throwCompilationError(
+                        `Cannot call static method ${idTextErr(exp.method)} using an instance.`,
+                        exp.loc,
+                    );
+                }
+                if (!abi.isStatic && isStaticCallFromType) {
+                    throwCompilationError(
+                        `Cannot call non-static method ${idTextErr(exp.method)} from a static context`,
+                        exp.loc,
+                    );
+                }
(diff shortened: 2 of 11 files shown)


hazyone:
Hope I fixed it the way you expected! :)

src/abi/AbiFunction.ts
```diff
@@ -6,6 +6,7 @@ import type { SrcInfo } from "@/grammar";
 
 export type AbiFunction = {
     name: string;
+    isStatic?: boolean;
```
hazyone: Btw I wasn’t sure about this part — should we make isStatic explicit, or leave it optional?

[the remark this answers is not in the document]
hazyone: Done

hazyone: Added explicit isStatic

dev-docs/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- [fix] Added static/instance checks for core static methods
```

src/types/stmts-failed/method-fromcell-call-non-static-context.tact
i582, line 1: Can you reformat all test files with tactfmt?

src/types/resolveExpression.ts
```diff
@@ -681,6 +681,22 @@ function resolveCall(
         if (srcT.kind === "struct") {
             if (StructFunctions.has(idText(exp.method))) {
                 const abi = StructFunctions.get(idText(exp.method))!;
+                const isInstanceCall = exp.self.kind !== "id";
+                const isStaticCallFromType =
+                    exp.self.kind === "id" && exp.self.text === src.name;
+
+                if (abi.isStatic && isInstanceCall) {
+                    throwCompilationError(
+                        `Cannot call static method ${idTextErr(exp.method)} using an instance.`,
```
i582: Maybe `Cannot call static method ${idTextErr(exp.method)} as an instance`?

src/types/resolveExpression.ts
```diff
@@ -681,6 +681,22 @@ function resolveCall(
...
+                        exp.loc,
+                    );
+                }
+                if (!abi.isStatic && isStaticCallFromType) {
+                    throwCompilationError(
+                        `Cannot call non-static method ${idTextErr(exp.method)} from a static context`,
```
i582: Maybe `Cannot call non-static method ${idTextErr(exp.method)} as static?

hazyone: Sounds a bit confusing. 'Static context' is pretty common in coding, I think
We could also go with: `Cannot call non-static method ${idTextErr(exp.method)} without an instance.`

src/types/resolveExpression.ts
hazyone: Not sure, maybe `Cannot call static method ${idTextErr(exp.method)} on an instance.`? 
But i can do any you want :)

src/types/stmts-failed/method-fromcell-call-non-static-context.tact
hazyone: Done, but only for my files. Most of the test files aren’t properly formatted, btw. I ran tactfmt on the whole folder and saw that it would trigger a rebuild of most snapshots.

src/types/resolveExpression.ts
anton-trunov: @hazyone  Let's go for your last option:

> Cannot call static method ${idTextErr(exp.method)} on an instance.

src/types/resolveExpression.ts
anton-trunov: Let's go with `Cannot call instance method ${idTextErr(exp.method)} as a static one`.

anton-trunov (changes requested):
LGTM, modulo two nits in error messages

PR #3172, opened 2025-05-20
fix: mismatched serialization of maps in overriden const/fun

### LINKED ISSUE #3155: Compiler silently accepts mismatched serialization annotations in trait overrides
The compiler incorrectly allows contract constants to override trait constants with mismatched serialization annotations (`as`), causing incompatible overrides to pass type checking.

**Minimal Example:**

```tact
trait T {
  abstract const M: map<Int as uint8, Int>;
}

contract C with T {
  override const M: map<Int as uint16, Int> = emptyMap(); // incompatible key serialization
}
```

**Current Behavior (incorrect):**
Compilation succeeds without errors:

```
Syntax and type checking succeeded.
```

**Expected Behavior:**
Compiler should reject overrides where serialization annotations (`as`) differ, ensuring type consistency and preventing subtle runtime serialization bugs.

---

### LINKED ISSUE #3236: Fix strict equality in `as`-annotations for `map` types
The typechecker incorrectly rejects the following two contracts:

```tact
contract Test {
    receive() {  }

    get fun mi(): map<Int, Int> {
        let res: map<Int as int257, Int as int257> = null;
        return res;
    }
}
```

```tact
contract Test {
    receive() {  }

    get fun mc(): map<Int, Int as coins> {
        let res: map<Int, Int as varuint16> = null;
        return res;
    }
}
```

> Type mismatch: "map<Int as int257, Int as int257>" is not assignable to "map<Int, Int>" 

> Type mismatch: "map<Int, Int as varuint16>" is not assignable to "map<Int, Int as coins>"

hazyone:
## Issue

Closes #3155. Closes #3236.

## Checklist


src/types/resolveDescriptors.ts
```diff
@@ -2699,3 +2715,49 @@ function checkRecursiveTypes(ctx: CompilerContext): void {
...
+    loc: SrcInfo,
+): void {
+    if (expected.kind !== "map" || actual.kind !== "map") return;
+
+    const aliasMap: Record<string, string> = {
+        coins: "varuint16",
+    };
+
+    const serializeKey = (type: TypeRefMap) =>
+        type.key === "Int" && type.keyAs == null ? "int257" : type.keyAs;
```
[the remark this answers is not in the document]
hazyone: Ok, I agree about strict equals(fixed), but TypeRef.keyAs is string | null, null !== undefined 🤔

dev-docs/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- [fix]: Fixed mismatched serialization of maps in const/fun
```

src/types/test-failed/override-const-map-keyas-empty.tact
i582, line 1: Can you reformat all test files with tactfmt?

hazyone:
Some cache problems on CI?
```
Test Suites: 169 passed, 169 total
Tests:       2 skipped, 2060 passed, 2062 total
Snapshots:   943 passed, 943 total
Time:        63.154 s, estimated 74 s
Ran all test suites.
✨  Done in 64.54s.

```

src/types/resolveDescriptors.ts
```diff
@@ -2699,3 +2715,48 @@ function checkRecursiveTypes(ctx: CompilerContext): void {
         return [];
     }
 }
+
+function checkMapSerializationCompatibility(
```
i582: Can we reuse `isAssignable()` here?

hazyone: We can, but I'm not sure it would be the best option. It uses strict equals on key/valueAs, so coins<->varuint16, Int<->as int257 conversions would be impossible.

anton-trunov: It's a bug and we should fix it and re-use `isAssignable`. Nice catch, btw. I opened an issue here: #3236.

hazyone: ok, moved to isAssignable.
also I added some checks to isAssignable, so I hope it now can close #3236

Check please :)

anton-trunov (approved):

Issue #3158, opened 2025-05-19
Compiler accepts invalid `bounced<T>` returns without shuffle, internal error with shuffle


Gusarich:
#3132 related

The compiler incorrectly accepts functions returning `bounced<T>` types unless a return shuffle (e.g., `asm(-> 0)`) is added. When the shuffle is added, it triggers an internal compiler error instead of a proper diagnostic.

**Minimal Example (Incorrectly Accepted):**

```tact
message(0x1) MyMsg {}

asm fun f(): bounced<MyMsg> {}  // Incorrectly accepted by compiler
```

**Minimal Example (Triggers Internal Compiler Error):**

```tact
message(0x1) MyMsg {}

asm(-> 0) fun f(): bounced<MyMsg> {}  // Causes internal compiler error
```

**Current Behavior:**

* Without shuffle: Compiler incorrectly accepts invalid return type.
* With shuffle: Compiler crashes with an internal error:

```
[INTERNAL COMPILER ERROR]: A <bounced> type cannot be returned from a function
```

**Related Issues:**
This issue relates to existing internal compiler error reports involving bounced return types (see issue #3132). Here, the focus is on the inconsistent behavior where adding a shuffle suddenly triggers the error.

**Expected Behavior:**
The compiler should consistently reject `bounced<T>` return types, providing clear diagnostics in both scenarios.

---

LLM Fuzzing discovery (see #3123)

Issue #3132, opened 2025-05-18
Internal compiler error when `asm` function returns `bounced<T>` type


Gusarich:
Defining an `asm` function that returns a `bounced<T>` type results in an internal compiler error.

**Minimal Example:**

```tact
struct S { x: Int; }

asm(->0) fun foo(): bounced<S> {
    ""
}
```

**Observed Compiler Error:**

```
[INTERNAL COMPILER ERROR]: A <bounced> type cannot be returned from a function
Please report at https://github.com/tact-lang/tact/issues
```

**Cause:**
The compiler currently throws an internal error when encountering `bounced<T>` return types in `asm` functions (`resolveDescriptors.ts`).

**Expected Behavior:**
Compiler should explicitly reject `bounced<T>` return types in `asm` functions with a proper user-facing error message instead of an internal crash.

---

LLM Fuzzing discovery (see #3123)

PR #3197, opened 2025-05-20
fix: bounced/trait returns from asm(shuffle/non-shuffle)

### LINKED ISSUE #3130: Internal compiler error when assembly function returns a trait
The compiler crashes with an internal error when an assembly (`asm`) function declares a return type of a trait.

**Minimal Example:**

```tact
trait T {}

asm (-> 0) fun foo(): T { NOP }
```

**Observed Compiler Error:**

```
[INTERNAL COMPILER ERROR]: A trait cannot be returned from a function
```

**Cause:**
The compiler currently doesn't handle traits as return types for assembly functions (`asm`), resulting in an uncaught internal error.

**Expected Behavior:**
The compiler should emit a clear user-facing diagnostic indicating that traits are not allowed as return types, rather than crashing with an internal error.

---

hazyone:
- Added checks for bounced/trait returns in non-shuffle asm functions

## Issue

Closes #3130, #3158, #3132.

## Checklist

### DIFF
--- a/dev-docs/CHANGELOG.md
+++ b/dev-docs/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [fix] Disallow self-inheritance for contracts and traits: PR [#3094](https://github.com/tact-lang/tact/pull/3094)
 - [fix] Added fixed-bytes support to bounced message size calculations: PR [#3129](https://github.com/tact-lang/tact/pull/3129)
+- [fix] Fixed compiler errors for bounced/trait return types
+- [fix] Added checks for bounced/trait returns in non-shuffle asm functions
 
 ### Docs
 
--- a/src/types/__snapshots__/resolveDescriptors.spec.ts.snap
+++ b/src/types/__snapshots__/resolveDescriptors.spec.ts.snap
@@ -1,5 +1,14 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`resolveDescriptors should fail descriptors for asm-fun-bounced-return 1`] = `
+"<unknown>:3:1: Function "foo" returns a <bounced> type, which cannot be returned from an "asm" function.
+  2 | 
+> 3 | asm (-> 0) fun foo(): bounced<Cell> { NOP }
+      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  4 | 
+"
+`;
+
 exports[`resolveDescriptors should fail descriptors for asm-fun-shuffle-arg-duplicate 1`] = `
 "<unknown>:3:1: asm argument rearrangement cannot have duplicates
   2 | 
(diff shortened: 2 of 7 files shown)


src/types/resolveDescriptors.ts
```diff
@@ -1089,9 +1105,9 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
                                 case "primitive_type_decl":
                                     retTupleSize = 1;
                                     break;
-                                case "trait":
+                                default:
```
verytactical: We try to list out all cases for a `kind`, so that in case a new kind is added, we get a TypeScript error telling another case wasn't handled.

If we put `default` here, we might miss handling the case here, merge it as is, and get internal compiler error.

src/types/__snapshots__/resolveDescriptors.spec.ts.snap
```diff
@@ -1,5 +1,14 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`resolveDescriptors should fail descriptors for asm-fun-bounced-return 1`] = `
+"<unknown>:3:1: Function "foo" returns a <bounced> type, which cannot be returned from an "asm" function.
+  2 | 
+> 3 | asm fun foo(): bounced<Cell> { NOP }
```
i582, line 6: Let's not actually forbid "bounced" here, it might make sense in the future

dev-docs/CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [fix] Disallow self-inheritance for contracts and traits: PR [#3094](https://github.com/tact-lang/tact/pull/3094)
 - [fix] Added fixed-bytes support to bounced message size calculations: PR [#3129](https://github.com/tact-lang/tact/pull/3129)
+- [fix] Fixed compiler errors for bounced/trait return types
```

src/types/resolveDescriptors.ts
```diff
@@ -1063,6 +1063,22 @@ export function resolveDescriptors(ctx: CompilerContext, Ast: FactoryAst) {
                     );
                 }
             }
+            if (returns.kind !== "void") {
+                if (returns.kind === "ref") {
+                    const typeInfo = types.get(returns.name);
+                    if (typeInfo?.kind === "trait") {
+                        throwCompilationError(
+                            `Function ${idTextErr(a.name)} returns a trait, but returning traits from "asm" functions is not supported (yet).`,
```
i582: Why `yet`? :D

hazyone: > Cause:
> The compiler currently doesn't handle traits as return types for assembly functions (asm), resulting in an uncaught internal error.

I've decided, that "currently" means tact-team has plans to do it in future :D

Ok, check the new one, please)

src/types/__snapshots__/resolveDescriptors.spec.ts.snap
hazyone: Check the new one, please :)

src/types/resolveDescriptors.ts
hazyone: my bad, ty

dev-docs/CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [fix] Balanced quotation in error messages for out-of-project-root imports: PR [#3242](https://github.com/tact-lang/tact/pull/3242)
 - [fix] Disallow self-inheritance for contracts and traits: PR [#3094](https://github.com/tact-lang/tact/pull/3094)
 - [fix] Added fixed-bytes support to bounced message size calculations: PR [#3129](https://github.com/tact-lang/tact/pull/3129)
+- [fix] Fixed compiler errors for trait return types: PR [#3197](https://github.com/tact-lang/tact/pull/3197)
+- [fix] Added checks for trait returns in non-shuffle asm functions: PR [#3197](https://github.com/tact-lang/tact/pull/3197)
+- Compiler now generates more efficient code for serialization: PR [#3213](https://github.com/tact-lang/tact/pull/3213)
```
```suggestion
```

anton-trunov (approved):
Thanks. 

Decided to merge as is and allow returning `bounced` types from asm functions in a follow-up PR.

PR #3207, opened 2025-05-21
chore(docs): show `self` or the extended type for referencing internal and extension functions, respectively

### LINKED ISSUE #3198: Consider adding `self` as `.reply()` is confusing
https://docs.tact-lang.org/cookbook/single-communication/#how-to-send-a-message-with-the-remaining-value

> The latter example is identical to using the [.reply() function](https://docs.tact-lang.org/cookbook/single-communication/#how-to-make-a-basic-reply).

novusnota:
## Issue

Closes #3198.


anton-trunov (changes requested):
All these changes look like those methods are static, but they are not. We should just remove the dots.

novusnota:
> All these changes look like those methods are static, but they are not. We should just remove the dots.

Not really. Cell, Builder and Slice before the dot and the name of the extension function are ubiquitous and used across the docs.

Sure, static methods would look identical to regular methods (extension functions), but at least it's clear that those are methods and not global functions. And we don't have many static methods for that to be a problem anyways.

docs/src/content/docs/book/cells.mdx
```diff
@@ -425,7 +425,7 @@ let areSlicesNotEqual = aSlice != bSlice; // false
 
 Note that direct comparison via the `=={:tact}` or `!={:tact}` operators implicitly uses [SHA-256](https://en.wikipedia.org/wiki/SHA-2#Hash_standard) hashes of the [standard `Cell{:tact}` representation](#cells-representation) under the hood.
 
-Explicit comparisons using `.hash(){:tact}` are also available:
+Explicit comparisons using `Cell.hash(){:tact}` or `Slice.hash(){:tact}` are also available:
```
anton-trunov:
```suggestion
Explicit comparisons using the `hash(){:tact}` extension function that works on cells and slices are also available:
```

docs/src/content/docs/book/debug.mdx
```diff
@@ -716,7 +716,7 @@ it('emits', async () => {
 });
 ```
 
-To parse the second emitted message, we could manually use several `.loadSomething(){:typescript}` functions, but that's overly brittle — if the fields of the `Ballroom{:tact}` [struct][struct] ever change, you'd need to start all over. That could significantly backfire when you have numerous tests written in that manner.
+To parse the second emitted message, we could manually use several `Slice.loadSomething(){:typescript}` functions, but that's overly brittle — if the fields of the `Ballroom{:tact}` [struct][struct] ever change, you'd need to start all over. That could significantly backfire when you have numerous tests written in that manner.
```
anton-trunov: What is `loadSomething`? I cannot find its definition anywhere in the docs. You should use prose to describe a family of methods

docs/src/content/docs/book/exit-codes.mdx
```diff
@@ -288,7 +288,7 @@ From the [Cells, Builders and Slices page](/book/cells#cells) of the Book:
 
 To construct a [`Cell{:tact}`][cell], a [`Builder{:tact}`][builder] is used. If you try to store more than 1023 bits of data or more than 4 references to other cells, an error with exit code $8$ is thrown: `Cell overflow`.
 
-This error can be triggered by [manual construction](/book/cells#cnp-manually) of the cells via [relevant `.storeSomething()` methods](/ref/core-cells#builder) or when [using structs and Messages and their convenience methods](/book/cells#cnp-structs).
+This error can be triggered by [manual construction](/book/cells#cnp-manually) of the cells via [relevant `Builder.storeSomething(){:tact}` methods](/ref/core-cells#builder) or when [using structs and Messages and their convenience methods](/book/cells#cnp-structs).
```
anton-trunov: `storeSomething` -- same here

docs/src/content/docs/book/operators.mdx
```diff
@@ -478,7 +478,7 @@ nullable != anotherNullable; // true
 
 :::note
 
-  The binary equality `=={:tact}` and inequality `!={:tact}` operators implicitly compare [maps](/book/maps) by the hashes of their respective [cells][cell] via the [`.hash(){:tact}`](/ref/core-cells#cellhash) function. While this is acceptable in the majority of cases—since most map serializers behave identically to the serializer from TON Blockchain sources—it is still possible to obtain false-negative results if a map is serialized manually or if the serialization logic is modified in certain libraries.
+  The binary equality `=={:tact}` and inequality `!={:tact}` operators implicitly compare [maps](/book/maps) by the hashes of their respective [cells][cell] via the [`Cell.hash(){:tact}`](/ref/core-cells#cellhash) function. While this is acceptable in the majority of cases—since most map serializers behave identically to the serializer from TON Blockchain sources—it is still possible to obtain false-negative results if a map is serialized manually or if the serialization logic is modified in certain libraries.
```
anton-trunov:
```suggestion
  The binary equality `=={:tact}` and inequality `!={:tact}` operators implicitly compare [maps](/book/maps) by the hashes of their respective [cells][cell] via the [`hash(){:tact}`](/ref/core-cells#cellhash) extension function. While this is acceptable in the majority of cases—since most map serializers behave identically to the serializer from TON Blockchain sources—it is still possible to obtain false-negative results if a map is serialized manually or if the serialization logic is modified in certain libraries.
```

docs/src/content/docs/book/statements.mdx
```diff
@@ -500,7 +500,7 @@ do {
 
 The `foreach{:tact}` loop operates on key-value pairs (entries) of the [`map<K, V>{:tact}`](/book/maps) type in sequential order: from the smallest keys of the map to the biggest ones.
 
-This loop executes a block of code for each entry in the given map, capturing the key and value on each iteration. This is handy when you don't know in advance how many items there are in the map or don't want to explicitly look for each of the entries using the [`.get(){:tact}`](/book/maps#get) [method](/book/functions#extension-function) of maps.
+This loop executes a block of code for each entry in the given map, capturing the key and value on each iteration. This is handy when you don't know in advance how many items there are in the map or don't want to explicitly look for each of the entries using the [`map.get(){:tact}`](/book/maps#get) [method](/book/functions#extension-function) of maps.
```
anton-trunov:
```suggestion
This loop executes a block of code for each entry in the given map, capturing the key and value on each iteration. This is handy when you don't know in advance how many items there are in the map or don't want to explicitly look for each of the entries using the [`get(){:tact}`](/book/maps#get) [method](/book/functions#extension-function) of maps.
```

novusnota (commented):
UPD: Whoops, didn't mean to comment

anton-trunov (approved):

PR #3216, opened 2025-05-22
feat(codegen): simplify `if-throw` pattern with `throw_if/throw_unless` call

### LINKED ISSUE #3215: Optimize `if (cond) throw` to `throwIf` and `if (!cond) throw` to `throwUnless`
https://github.com/tact-lang/tact/blob/e6e7294a31840f617f1b2cae9796e4a14ab82c24/src/benchmarks/wallet-v5/tact/wallet-v5.tact#L132-L136

i582:
## Issue

Closes #3215.


src/generator/writers/writeFunction.ts
```diff
@@ -614,6 +636,20 @@ function writeCondition(
     returns: TypeRef | null | string,
     ctx: WriterContext,
 ) {
+    const threwCode = extractThrowErrorCode(f.trueStatements);
+    const isAloneIf =
+        f.falseStatements === undefined || f.falseStatements.length === 0;
+
+    if (!elseif && isAloneIf && threwCode !== undefined) {
+        // if (cond) { throw(X) } => throw_if(X, cond)
+        // if (!cond) { throw(X) } => throw_unless(X, cond)
```
anton-trunov, line 674: This is semantically non-equivalent change in the presence of mutation. For instance,

```tact
extends mutates fun inc(self: Int): Int { self += 1; return self }
contract Test {
    receive() {  }
    // returns 43
    get fun ifThrow(): Int {
        let x: Int = 42;
        if (x.inc() == 0) {
            throw(x.inc());
        }
        return x;
    }
    // returns 44
    get fun throwIf(): Int {
        let x: Int = 42;
        throwIf(x.inc(), x.inc() == 0);
        return x;
    }
}
```

anton-trunov (changes requested):
You need to analyze if the expressions have side-effects for this optimization to be correct.

- the `throwIf` version unconditionally does the side-effects of both expressions;
- the `throwIf` version changes the order of evaluation.

So, the exit code expression should be side-effect free and the two expression should be commutative.

src/generator/writers/writeFunction.ts
i582: Indeed, fixed

src/generator/writers/writeFunction.ts
```diff
@@ -615,6 +615,47 @@ const rewriteWithIfNot = (
...
+    ctx: CompilerContext,
+): Ast.Expression | undefined => {
+    const [stmt] = stmts;
+    if (
+        stmt?.kind === "statement_expression" &&
+        stmt.expression.kind === "static_call" &&
+        stmt.expression.function.text === "throw"
+    ) {
+        const arg = stmt.expression.args.at(0);
+        if (arg?.kind === "static_call" || arg?.kind === "method_call") {
```
anton-trunov, line 629: Static calls cannot change state: everything is passed/returned by value

anton-trunov: Ah, you mean calls can change the contract state, by updating `self.foo`. My bad

anton-trunov (approved):

PR #3219, opened 2025-05-22
fix:  compiler crashes when a contract name exceeds filesystem limits

### LINKED ISSUE #3138: Compiler crashes when contract name exceeds filesystem limits
The compiler crashes with an unhandled `ENAMETOOLONG` error when compiling contracts with very long names, due to generated filenames exceeding filesystem limitations.

**Minimal Example:**

```tact
contract AAAAAAAAAAAAAAA... (extremely long name) {}
```

**Observed Error:**

```
Error: ENAMETOOLONG: name too long, open '/tmp/longname.tact_AAAAAAAAAA....fc'
```

**Cause:**
The compiler directly uses the contract name to generate filenames without checking the length, causing unhandled filesystem errors.

**Expected Behavior:**
The compiler should enforce a sensible maximum contract name length or gracefully handle filename generation to avoid crashing due to filesystem constraints.

---

hazyone:
## Issue

Closes #3138.

## Checklist


src/vfs/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+    ext: string,
+    maxLen = 255,
+): string => {
+    const full = name + ext;
+
+    if (full.length <= maxLen) {
+        return full;
+    }
+
+    const hash = sha256_sync(Buffer.from(name)).toString("hex").slice(0, 8);
```
verytactical, line 24: 2^32 hashes are still possible to mine, possibly creating a name clash

I think it would rather be better to just err whenever `len(projectName) + len(contractName) + 1 >= 255`, or even `len(projectName) > 100 || len(contractName) > 100`.

These are not the kind of situations that might happen in non-malicious code, and emitting an error would be completely fine.

PR #3244, opened 2025-05-25
fix(codegen): detect out-of-range exit codes in fallback receivers

### LINKED ISSUE #3117: Large numeric exit codes in fallback receiver lose precision, breaking FunC output
Compiler incorrectly handles large numeric literals in fallback receivers consisting of a single `throw(CODE)` statement (`statically-known-single-throw` case). The compiler converts the `BigInt` to a JavaScript `Number`, losing precision and causing scientific notation to appear in generated FunC code:

```tact
contract BigTest {
    receive(_: Slice) {
        throw(57896044618658097711785492504343953926634992332820282019728792003956564819968);
    }
}
```

Incorrect generated asm function in FunC:

```func
(slice, int) ~load_opcode_internal(slice s) asm( -> 1 0) "32 LDUQ 5.78960446186581e+76 PUSHINT SWAP THROWANYIFNOT";
```

This breaks compilation when the function is invoked and leads to invalid opcode errors.

The compiler must handle large numeric exit codes precisely, preserving the `BigInt` representation to ensure valid FunC assembly generation.

---

anton-trunov:
otherwise it leads to errors at the FunC compilation stage.

It's a workaround we need to do because of the lack of infrastructure. The proper fix would be performed at the constant evaluation stage.

## Issue

Closes #3117.

## Checklist


src/test/compilation-failed/throw-errors.spec.ts
```diff
@@ -0,0 +1,9 @@
+import { itShouldNotCompile } from "@/test/compilation-failed/util";
+
+describe("throw-errors", () => {
```
i582, line 3: let's add a test with a negative number

i582 (dismissed):

i582 (approved):

PR #3272, opened 2025-05-27
fix: don't require explicit re-declaration for abstract methods and fields in traits

### LINKED ISSUE #2630: Traits Cannot Inherit Abstract Functions And Constants Without Re-Declaring Them
When a trait inherits from another trait that declares an abstract function or constant, the compiler requires the child trait to re-declare it — even if it intends to leave it unimplemented and defer it to the contract.

This behavior is **not covered anywhere in the documentation**.

### Reproducible Example

```tact
trait TraitA {
    abstract fun compute(): Int;
    virtual fun getValue(): Int {
        return 10;
    }
}

trait TraitB with TraitA {
    override fun getValue(): Int {
        return 20;
    }
}

contract C with TraitB {
    override fun compute(): Int {
        return self.getValue();
    }

    receive() {
        // nothing
    }
}
```

The compiler produces the following error:

```
Trait 'TraitA' requires function 'compute'
```

even though `compute()` is implemented in the final contract, and not used directly within `TraitB`.

### What’s Missing

The documentation says that **abstract functions and constants must be implemented in inheriting contracts**, but does **not** clarify whether intermediate traits must re-declare them, or whether they may remain abstract until the contract level.

This lack of clarity can confuse users and leads to the false assumption that trait-to-trait inheritance of abstract functions and constants is allowed.

### Suggested Fix

- 📚 **Documentation:** Add a clear explanation of whether traits must explicitly re-declare abstract functions and constants from parent traits, or whether contracts are allowed to implement them directly.
- 🛠️ **Compiler (Optional):** If the current behavior is too strict, consider allowing traits to pass abstract requirements through to contracts, as in typical OOP trait systems.

This behavior is currently undocumented and blocks an intuitive use of abstract functions and constants across traits. It should be clarified or adjusted.

---

i582:
## Issue

Closes #2630.

## Checklist

### DIFF
--- a/dev-docs/CHANGELOG.md
+++ b/dev-docs/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Language features
+
+- [fix] Compiler now doesn't require explicit re-declaration for abstract methods and fields in traits: PR [#3272](https://github.com/tact-lang/tact/pull/3272)
+
+### Release contributors
+
+- [Petr Makhnev](https://github.com/i582)
+
 ## [1.6.11] - 2025-05-26
 
 ### Language features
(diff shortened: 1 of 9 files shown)


anton-trunov (changes requested):
Let's do a couple of things more
- improve the error messages for the cases when abstract functions / constants are not implemented in the contract that inherits those;
- add an end-to-end test for the case with `self.getValue()` being called from the overriden `compute` function

i582:
> Let's do a couple of things more
> 
> * improve the error messages for the cases when abstract functions / constants are not implemented in the contract that inherits those;
> * add an end-to-end test for the case with `self.getValue()` being called from the overriden `compute` function

Done

anton-trunov (approved):

PR #3327, opened 2025-06-01
feat: optimize `__tact_randomized` and `__tact_child_contract_codes`

### LINKED ISSUE #3326: Optimize away unused optional globals (`__tact_child_contract_codes`, `__tact_randomized`)
Currently, optional global variables `__tact_child_contract_codes` and `__tact_randomized` are always included in the generated FunC output, even when they're never used by the contract. These globals can be safely omitted when unused, leading to a minor gas-saving improvement per execution.

**Task:**

* Detect whether `__tact_child_contract_codes` and `__tact_randomized` globals are actually used.
* Omit these globals entirely from FunC output when not referenced.

**Impact:**
This optimization results in a slight improvement in gas usage (saving a few gas points per execution) and reduces contract byte size minimally.

Gusarich:
closes #3326

### DIFF
--- a/dev-docs/CHANGELOG.md
+++ b/dev-docs/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Code generation
 
-- Optimized optional global variables initialization: PR [#3326](https://github.com/tact-lang/tact/pull/3326)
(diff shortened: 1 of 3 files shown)


anton-trunov (changes requested):
Let's add some codegen snapshot test in `src/test/codegen-check/codegen.test.ts` to show the two globals are not written

PR #3334, opened 2025-06-01
fix(docs): adjust signatures of `replace()` and `replaceGet()` map methods

### LINKED ISSUE #3325: Documentation incorrectly omits `null` possibility in `map.replace` method signature
The documentation for the `.replace()` method on `map<K, V>` currently specifies its third argument as strictly non-null (`val: V`). However, the method explicitly supports passing `null` to delete an entry from the map.

**Current documented signature:**

```
extends mutates fun replace(self: map<K, V>, key: K, val: V): Bool;
```

**Actual supported usage:**
Passing `null` as `val` deletes the entry from the map, which compiles and works correctly.

**Issue:**
The documented signature should reflect this allowed use-case explicitly, updating it to:

```
extends mutates fun replace(self: map<K, V>, key: K, val: V?): Bool;
```

**Impact:**
This clarification prevents confusion among developers and ensures the documented method signatures accurately represent actual compiler-supported behavior.

---

novusnota:
## Issue

Closes #3325.

### DIFF
--- a/docs/src/content/docs/book/maps.mdx
+++ b/docs/src/content/docs/book/maps.mdx
@@ -289,7 +289,7 @@ if (gotButUnsure != null) {
 
 ```tact
 // K and V correspond to the key and value types of the given map
-extends mutates fun replace(self: map<K, V>, key: K, val: V): Bool;
+extends mutates fun replace(self: map<K, V>, key: K, val: V?): Bool;
 ```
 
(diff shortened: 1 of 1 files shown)


anton-trunov (approved):
Thanks!

anton-trunov:
We also should fix this in `stubs.tact`

PR #3357, opened 2025-06-04
feat: deep selector hack

### LINKED ISSUE #3336: Optimize function selector by avoiding unnecessary inlining of internal/external receivers
Currently, when contracts have both internal and external receivers, our low-level function selector generates two separate `IFNOTJMP` instructions: one for internal and one for external messages. However, at runtime, only one of these branches can ever be executed, since the incoming message must be either internal or external.

The inefficient compilation occurs because the Fift-assembler decides whether to use `JMPREF` or inline the code (`PUSHCONT` + `JMP`) based on the size of the code block. If the receiver handlers are small enough, the assembler chooses to inline both branches, unnecessarily enlarging the current code cell and causing extra implicit `JMPREF`s.

Since we know that only one receiver (internal or external) can ever execute, we can optimize this scenario by implementing custom logic in our function selector. This logic would explicitly determine whether to inline or reference each receiver's code block, avoiding unnecessary duplication and implicit jumps.

pyAndr3w:
Closes #3336


pyAndr3w:
## Current Selector Structure

The current selector consists of:
- `maybe set c3`
- `if-jmp to internal`
- `if-jmp to external`
- `maybe jmp to c3` (or throw 130)

This logic is implemented in the selector-hack get-method.

---

## Proposed Optimization

I propose moving the selector into `recv_internal` (or `recv_external`, if exists) and making `recv_internal` (`recv_external`) the root cell of the contract.

This maneuver will make the bytecode layout of contracts slightly more compact and help avoid unnecessary "jumps" between code cells.

---

## Proof of Concept Results

At this stage, proof of concept work for this optimization has been completed. The results are available and they are good.

### Current Issues

There is a problem with the absence of the `maybe set c3` condition. There is also a problem that tonapi will not be able to properly index contracts with `recv_external` due to the different location of the function dictionary.

---

## Solution

The solution to the problem with the missing `maybe set c3` condition is known.

We can obtain a dictionary with all functions at the moment when compilation of `recv_internal` (`recv_external`) begins. It will be necessary to manually exclude from the dictionary functions that use the inline flag or are not used in the contract. 

For this purpose, `Asm.fif` has a separate hashmap with function usage flags. However, the `recv_internal` function is not yet compiled and we don't know which functions are used in it (function flags may change). 

To solve this, embedding of `fake_*` functions has been implemented so that all flags are set at the `Asm.fif` level and we can safely obtain the necessary dictionary that includes only used functions.

pyAndr3w:
👉👈 It is also necessary to fix the draft so that it passes absolutely all tests.

Kaladin13:
Can you please add benchmark table result output here in the conversation? I mean to compare the impact of a new hack

pyAndr3w:
@Kaladin13 

image
image
image
image
image
image
image

pyAndr3w:
image

One test is failing due to the lack of a restriction in the Tact compiler on the minimum `method_id`.
In the `deep selector hack`, getters with `method_id` < 65536 end up in the hashmap with non-inline procedures.

pyAndr3w:
base jetton e2e tests passed

https://testnet.tonviewer.com/kQBEg_BB-yM5Hyik9O9xuVvVDGEqhKTDwbTVf3rYQrIk_HSN?section=method

PR #3364, opened 2025-06-05
feat(codegen): don't load contract state unless getter reads or writes it

### LINKED ISSUE #3363: Don't load contract state unless getter reads or writes it
```tact
contract Foo {
   get fun bar(): Int { return 0 }
}
```

i582:
Closes #3363.


Gusarich (commented):
```tact
contract Test {
    get fun zero(): Int {      // does not read state
        return 0;
    }

    get fun bugTrigger() {
        self.zero();        // invokes zero() via the non-mutating wrapper
    }
}
```

->

```
FunC compilation error: /Users/daniil/Coding/tact/test.tact_Test.fc:63:28: error: cannot apply function $Test$_fun_zero : () -> ((), int) to arguments of type tuple: cannot unify type tuple with ()
      $self~$Test$_fun_zero();
```

Gusarich:
This optimization can incorrectly skip generating `contract_load()` for getters that indirectly use `initOf` or `codeOf` via static functions. Currently, `computeGettersEffects()` only considers effects from the arguments of `static_call` and does not analyze the body of called static functions. As a result, it might misclassify getters as storage-independent when they're actually dependent on storage initialization.

**Example that breaks with current implementation:**

```tact
contract Child {
    fun justForCodesDict(): StateInit {
        return initOf Test();
    }
}

fun helper(): StateInit {
    // requires __tact_child_contract_codes initialized by $contract_load
    return initOf Child();
}

contract Test {
    x: Int = 0;

    receive() {
        self.x += 1;
    }

    get fun broken(): Cell {
        // computeGettersEffects sees no storage access here
        return helper().code;
    }
}

```

You should enhance the analysis to recursively examine the bodies of static functions invoked by getters, ensuring proper detection of storage dependencies and preventing runtime errors caused by omitted `contract_load()`.

PR #3378, opened 2025-06-06
feat(docs): receivers accept all incoming funds and the amount of received native, Toncoin funds can be checked with `context().value`

### LINKED ISSUE #1086: Document that receivers accept all incoming funds by default

### LINKED ISSUE #1089: How do I check the amount of native coins incoming message carries with it?
By calling `context().value`

novusnota:
## Issue

Closes #1089.
Closes #1086.


docs/src/content/docs/book/receive.mdx
```diff
@@ -54,6 +54,28 @@ Note that the receiver `receive(msg: Slice){:tact}` acts as a fallback that catc
 
 If there is no receiver to process a message type and the fallback receiver `receive(msg: Slice){:tact}` is not declared, the transaction will fail with exit code [130](/book/exit-codes/#130).
 
+Receivers accept all incoming funds by default. That is, without explicitly sending a message that will refund spare Toncoin with the [`cashback(){:tact}`](/ref/core-send#cashback) function, all the incoming message value will be kept by the contract.
```
Kaladin13, line 59:
```suggestion
Receivers accept all incoming funds by default. That is, without explicitly sending a message that will refund spare Toncoin with the [`cashback(){:tact}`](/ref/core-send#cashback) function, all the incoming message value will be kept by the contract. Note, however, that the exact value by which the balance will be increased, will be less that `context().value`, since compute fee for contract execution will be deducted from this value
```

Kaladin13 (changes requested):
Let's add comment that the incoming value is not `exactly` the context.value. I will add a few examples of this semantic to the cookbook later, we can add cross-link here for more details

docs/src/content/docs/book/receive.mdx
```diff
@@ -54,6 +54,28 @@ Note that the receiver `receive(msg: Slice){:tact}` acts as a fallback that catc
...
+
+To check the amount of Toncoin (native coins) the incoming message carries with it, you can use the [`context(){:tact}`](/ref/core-contextstate#context) function and access the `value` field of its resulting [struct](/book/structs-and-messages#structs).
+
+```tact /context\(\).value/
+contract WalletV4(
+    // ...
+) {
+    // ...
+    receive(msg: PluginRequestFunds) {
+        require(myBalance() - context().value >= msg.amount);
```
Shvandre: require takes 2 arguments :)

Shvandre: I'd better write something like
```
receive(msg: PluginRequest) {
    require(context().value >= getGasFee(myGasFee), "Plugin request need enough TONS for gas")
   ...
}
```
But it is fully up to you

Shvandre (dismissed):
good

docs/src/content/docs/book/receive.mdx
novusnota: Hmm, maybe you've meant the storage fee here?

UPD: Nope, got your point. I've added the note suggested by you and slighly redacted by me to the end of the next paragraph, as it fit there better imo :)

docs/src/content/docs/book/receive.mdx
novusnota: Good suggestion, although in that particular example I'm thinking it's better to use a more straightforward example with less abstractions. Plus, it's a direct `require()` rewrite of the following line from our benchmarks:

https://github.com/tact-lang/tact/blob/02a3b663218c049a6163c97e6a8004eba46760d4/src/benchmarks/wallet-v4/tact/wallet-v4.tact#L87

Kaladin13 (approved):

PR #3399, opened 2025-06-11
feat(docs): remark the futility of on-chain state variables retrieval

### LINKED ISSUE #1153: Docs: how do I read another contract's state variable?
Contracts cannot call each other's getters, so we need to show how to read other contract's variables, provided we have control over both contracts's source code.

novusnota:
## Issue

Closes #1153.


docs/src/content/docs/book/contracts.mdx
```diff
@@ -242,6 +250,14 @@ contract Example {
 }
 ```
 
+:::note
```
anton-trunov, line 245: why this duplicate?

novusnota: To have it for params & variables. We can leave the former out.

PR #3472, opened 2025-08-15
fix(docs): clarify data layout of bounced messages

### LINKED ISSUE #3469: Clarify the data layout in bounced messages
According to [Bounced message section](https://docs.tact-lang.org/book/bounced/) of tact docs 

> Only the first message fields that fit into the 224 bits will be available to use in bounced message receivers.

This phrase is a bit confusing. It seems that if the second field of the message body fits into 224 bits, and the first does not, then the second field will be availabe in bounced message receivers. Although in fact, if the bits of the first field do not fit into this limit, then no fields will be available.

I think the following will be less confusing:

> Each field, sequentially and completelly fitting into the remaining 224 bits, will be available in bounced message receivers. If only a part of the bits of a field fits within this limit, it will not be available.

It is also desirable to add a phrase somewhere that these 224 bits are just a cut of bits from the cell of the original message. That is, even if the first field does not fit into this limit and will not be available in bounced message receivers, some of its bits are still in this cell.

novusnota:
## Issue

Closes #3469.

### DIFF
--- a/docs/src/content/docs/book/bounced.mdx
+++ b/docs/src/content/docs/book/bounced.mdx
@@ -29,7 +29,7 @@ Since message fields are always laid out sequentially and no automatic rearrange
 
 All subsequent fields that exceed the limit will not be available in bounced message receivers. This also includes fields that fit only partially.
 
-If the first field does not wholly fit the 224-bit limit, none of the bounced message fields would be available. Therefore, prefer arranging the important fields that fit into the limit first.
+If the first field does not wholly conform to the 224-bit limit, none of the bounced message fields would be available. Therefore, to ensure access, prefer arranging the important fields first.
 
 ```tact
 contract Bounce() {
--- a/docs/src/content/docs/book/bounced.mdx
+++ b/docs/src/content/docs/book/bounced.mdx
@@ -48,7 +48,7 @@ message TwoFields {
 
 :::caution
 
-  To prevent potential issues, Tact only allows using the fields that wholly fit the limit. However, the blockchain views that limit as a truncation boundary and keeps the raw bits that fit the total 256-bit message limit in the original message body cell.
+  To prevent potential issues, Tact only allows using the fields that wholly fit the limit. However, the blockchain views that limit as a truncation boundary and keeps the raw bits that fit the [total 256-bit body limit](#caveats) in the original bounced message body cell.
 
   You can access that data by removing the concrete receiver and using the [fallback bounced message receiver](#fallback) instead. It will provide you the original, raw body of the received message as a [`Slice{:tact}`][slice].
 
--- a/docs/src/content/docs/book/bounced.mdx
+++ b/docs/src/content/docs/book/bounced.mdx
@@ -46,6 +46,32 @@ message TwoFields {
 }
 ```
 
+:::caution
+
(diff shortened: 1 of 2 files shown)


novusnota:
CI has some issues with the connection to npm.dev-internal.org. The latter may experience some troubles, and/or its maintenance might have been abandoned.

cc @Mobyman @anton-trunov

anton-trunov:
merge conflicts

Karkarmath (dismissed):

novusnota:
> merge conflicts

Resolved!

docs/src/content/docs/book/bounced.mdx
```diff
@@ -42,6 +46,32 @@ message TwoFields {
 }
 ```
 
+:::caution
+
+  To prevent potential issues, Tact only allows using the fields that wholly fit the limit. However, the blockchain views that limit as a truncation boundary and keeps the raw bits that fit the total 256-bit message limit in the original message body cell.
```
anton-trunov: It's confusing: need to clarify 256 vs 224 bits

docs/src/content/docs/book/bounced.mdx
```diff
@@ -25,7 +25,11 @@ contract Bounce() {
 message BB {}
 ```
 
-Only the first message fields that fit into the 224 bits will be available to use in bounced message receivers. Sometimes you need to rearrange the order of the fields in your message so the relevant ones would fit into the limit.
+Since message fields are always laid out sequentially and no automatic rearrangements are made, the partial representation made by the `bounced<M>{:tact}` constructor would only contain the first fields that fit entirely within the 224-bit limit.
+
+All subsequent fields that exceed the limit will not be available in bounced message receivers. This also includes fields that fit only partially.
+
+If the first field does not wholly fit the 224-bit limit, none of the bounced message fields would be available. Therefore, prefer arranging the important fields that fit into the limit first.
```
novusnota:
```suggestion
If the first field does not wholly conform to the 224-bit limit, none of the bounced message fields would be available. Therefore, to ensure access, prefer arranging the important fields first.
```

Due to excessive use of the "fit" elsewhere, I thought that it would be better to replace its use here with a synonym.

docs/src/content/docs/book/bounced.mdx
novusnota: image

Earlier there's an explanation for the "256 vs 224 bits" discrepancy. We can link the phrase "total 256-bit message limit" to the start of the "Caveats".

novusnota, line 51:
```suggestion
  To prevent potential issues, Tact only allows using the fields that wholly fit the limit. However, the blockchain views that limit as a truncation boundary and keeps the raw bits that fit the [total 256-bit message limit](#caveats) in the original message body cell.
```

novusnota:
```suggestion
  To prevent potential issues, Tact only allows using the fields that wholly fit the limit. However, the blockchain views that limit as a truncation boundary and keeps the raw bits that fit the [total 256-bit body limit](#caveats) in the original bounced message body cell.
```

novusnota: I've applied the changes above :)

PR #3492, opened 2025-10-20
fix(docs) : updated invalid link for Address type

fruitful1l:
Now given link in docs/src/content/docs/book/types.mdx leads to https://docs.ton.org/v3/documentation/smart-contracts/addresses/address as I suppose it should. The link before was invalid.


anton-trunov (approved):
Thanks a lot!

PR #3499, opened 2025-11-27
fix(docs): clarify that `try...catch` cannot catch an exit code -14

### LINKED ISSUE #3498: Example for exit code 13 (-14) cannot catch the exit code, while it says it can
```tact
try {
    repeat (pow(2, 31) - 1) { }
} catch (exitCode) {
    // ← we won't get here, as execution will
    //   end in the try {} block due to lack of gas
}

### PR DESCRIPTION
**Pull request by novusnota**
## Issue

Closes #3498.

### DIFF
--- a/docs/src/content/docs/book/exit-codes.mdx
+++ b/docs/src/content/docs/book/exit-codes.mdx
@@ -402,9 +402,11 @@ This is done to prevent the resulting code ($-14$) from being produced artificia
 
 ```tact
 try {
+    // The exit code -14 can never be caught in try...catch,
+    // and it is always thrown for the whole contract
     repeat (pow(2, 31) - 1) {}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by anton-trunov — APPROVED** [review]

File: docs/src/content/docs/book/exit-codes.mdx [hunk 1]
```diff
@@ -402,9 +402,11 @@ This is done to prevent the resulting code ($-14$) from being produced artificia
 
 ```tact
 try {
+    // The exit code -14 can never be caught in try...catch,
+    // and it is always thrown for the whole contract
     repeat (pow(2, 31) - 1) {}
-} catch (exitCode) {
```
ucef94, line 406: (+)

ucef94 (commented):
+

PR #3506, opened 2026-02-03
fix(docs): replace 'it's own' with 'its own' in `receive.mdx`

zent7x:
## Summary
- Fixed grammatical error in docs/src/content/docs/book/receive.mdx
- Changed "it's own" to "its own" (possessive form, not contraction)

## Test plan
- No tests needed, documentation change only


novusnota (approved):
Good catch, thanks!

PR #3507, opened 2026-02-03
fix(docs): remove duplicate 'the' in `base.tact`

zent7x:
## Summary
- Fixed typo in src/stdlib/stdlib/std/internal/base.tact
- Removed duplicate "the" in doc comment: "use the the `send` function" -> "use the `send` function"

## Test plan
- No tests needed, comment/documentation change only


novusnota (approved):
Changes in `stdlib` require building a new `stdlib.ts` bundle, so I went ahead and ran `yarn build:fast` to do so :)

PR #3508, opened 2026-02-03
fix(docs): replace 'cannot defined be' with 'cannot be defined as' in `functions.mdx`

zent7x:
## Summary
- Fixed word order typo in docs/src/content/docs/book/functions.mdx
- Changed "cannot defined be virtual or abstract" to "cannot be defined as virtual or abstract"

## Test plan
- No tests needed, documentation change only


novusnota (approved):
Good catch, thanks!

